PR #21, opened 2023-05-15
ci: use xdist to parallelize test suite

jorgenengelsen:

This pull request is needed because the test suite is slow. Parallelization improves runtime of tests by 2x locally on my machine

- Add xdist as dev-dependency
- Run CI tests with pytest-xdist


TeeeJay (approved):
:)

PR #26, opened 2023-05-16
refactor: fix more typing

jorgenengelsen:

On our road to mypy ´strict´ mode

- Turn on a few more mypy rules, fixes some errors

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
@@ -35,7 +35,7 @@ class Rates:
             The corresponding stream day rates
         """
         regularity = np.asarray(regularity, dtype=np.float64)  # type: ignore[assignment]
-        return np.divide(calendar_day_rates, regularity, out=np.zeros_like(calendar_day_rates), where=regularity != 0.0)
+        return np.divide(calendar_day_rates, regularity, out=np.zeros_like(calendar_day_rates), where=regularity != 0.0)  # type: ignore[comparison-overlap]
 
     @staticmethod
     def to_calendar_day(stream_day_rates: np.ndarray, regularity: List[float]) -> np.ndarray:
@@ -130,12 +130,12 @@ class TimeSeries(GenericModel, Generic[TimeSeriesValue], ABC):
         use_enum_values = True
 
     @validator("values", each_item=True, pre=True)
-    def convert_none_to_nan(cls, v, field: ModelField):
+    def convert_none_to_nan(cls, v: float, field: ModelField) -> TimeSeriesValue:
         if field.outer_type_ is float and v is None:
             return math.nan
         return v
 
-    def __len__(self):
+    def __len__(self) -> int:
         return len(self.values)
 
     @abstractmethod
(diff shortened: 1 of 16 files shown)


src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -214,7 +214,9 @@ def __setitem__(self, indices: Union[slice, int, List[int]], values: Union[TimeS
                     f"Could not update timeseries, Combination of indices of type '{type(indices)}' and values of type '{type(values)}' is not supported"
                 )
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: object) -> bool:
```
TeeeJay, line 217: `object` or `Any`? In general we have used any, what is the reason to use object here?

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -281,7 +283,7 @@ def resample(self, freq: Frequency) -> Self:
             unit=self.unit,
         )
 
-    def __mul__(self, other) -> Self:
+    def __mul__(self, other: object) -> Self:
```
TeeeJay, line 286: and here etc?

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -365,7 +367,7 @@ def resample(self, freq: Frequency) -> Self:
             unit=self.unit,
         )
 
-    def __truediv__(self, other) -> TimeSeriesRate:
+    def __truediv__(self, other: object) -> TimeSeriesRate:
```
TeeeJay, line 370: here

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -754,7 +756,9 @@ def __getitem__(self, indices: Union[slice, int, List[int], np.ndarray]) -> Time
             f"Unsupported indexing operation. Got '{type(indices)}', expected indices as a slice, single index or a list of indices"
         )
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: object) -> bool:
```
TeeeJay, line 759: here

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/legacy_consumer/consumer_function/results.py
```diff
@@ -46,7 +46,7 @@ class Config:
         arbitrary_types_allowed = True
 
     @abstractmethod
-    def extend(self, other) -> ConsumerFunctionResultBase:
+    def extend(self, other: object) -> ConsumerFunctionResultBase:
```
TeeeJay, line 49: here

TeeeJay (approved):
nice cleanup. did you do it manually or use a tool for it?

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
jorgenengelsen: I believe this is due to "liskov substitution"?  Since the class inherits from object, it should also support object in all functions. Mypy does not like 'any', but not sure it is any better. 

Googled around and found some support for using 'object' but I am open for discussion

https://stackoverflow.com/questions/37557411/why-does-defining-the-argument-types-for-eq-throw-a-mypy-type-error

jorgenengelsen:
> nice cleanup. did you do it manually or use a tool for it?

Not had any luck with the automatic tools. Best to do it manually.

Turn on one rule at the time and try to resolve the errors. Still 300-400 errors left, so plenty to choose from 😄

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
jorgenengelsen: https://mypy.readthedocs.io/en/latest/dynamic_typing.html#any-vs-object

Mypy reccomends 'object' over 'any'. Will keep as-is for now. May revert later

PR #30, opened 2023-05-16
Add consumer system v2 sub results

mortendaehli:
Adding detailed results to the new Consumer System V2

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/core/consumers/compressor/component.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/core/consumers/compressor/component.py
@@ -133,7 +133,7 @@ class Compressor(BaseConsumerWithoutOperationalSettings):
             sub_components=[],
             models=[
                 core_results.CompressorModelResult(
-                    name="N/A",
+                    name="N/A",  # No context available to populate model name
                     timesteps=evaluated_timesteps,
                     is_valid=TimeSeriesBoolean(
                         timesteps=evaluated_timesteps,
--- a/src/ecalc/libraries/libecalc/common/libecalc/core/consumers/compressor/component.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/core/consumers/compressor/component.py
@@ -4,10 +4,15 @@ from datetime import datetime
 from typing import Dict, List, Optional
 
 import numpy as np
+from libecalc import dto
 from libecalc.common.exceptions import ProgrammingError
 from libecalc.common.temporal_model import TemporalModel
 from libecalc.common.units import Unit
-from libecalc.common.utils.rates import TimeSeriesBoolean, TimeSeriesRate
+from libecalc.common.utils.rates import (
+    TimeSeriesBoolean,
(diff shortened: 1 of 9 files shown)


src/ecalc/libraries/libecalc/common/libecalc/core/consumers/compressor/component.py
```diff
@@ -94,10 +98,53 @@ def evaluate(
                 values=aggregated_result.is_valid, timesteps=evaluated_timesteps, unit=Unit.NONE
             ),
             id=self.id,
+            recirculation_loss=TimeSeriesRate(
+                values=aggregated_result.recirculation_loss,
```
jorgenengelsen, line 102: What about regularity here? Should it not be included in all timeseries?

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/consumer_system.py
```diff
@@ -211,9 +216,40 @@ def evaluate(
...
+
+            sub_component_results.extend(
+                [
+                    consumer.evaluate(adjusted_operational_setting)
+                    for consumer, adjusted_operational_setting in zip(sorted_consumers, adjusted_operational_settings)
+                ]
+            )
+
+        # Change to 1-index operational_settings_used
+        for idx, value in enumerate(operational_settings_used.values):
```
jorgenengelsen, line 251: idx?

jorgenengelsen (approved):

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/compressor/component.py
mortendaehli: Updated :)

Not sure exactly where we want to isolate regularity. I think it is Okay to pass it down to the consumer, but we want to investigate if we can remove it completely from libecalc.core

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/consumer_system.py
mortendaehli: Renamed to index. I think idx is a common short for index.

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/pump/component.py
```diff
@@ -81,6 +88,23 @@ def evaluate(
...
+            ),
+            inlet_pressure_bar=TimeSeriesFloat(
+                values=aggregated_result.suction_pressure,
+                timesteps=evaluated_timesteps,
+                unit=Unit.BARA,
+            ),
+            outlet_pressure_bar=TimeSeriesFloat(
+                values=aggregated_result.discharge_pressure,
+                timesteps=evaluated_timesteps,
+                unit=Unit.BARA,
```
jorgenengelsen, line 107: regularity may be needed here aswell then?

jorgenengelsen (changes requested):

jorgenengelsen (approved):

PR #39, opened 2023-05-26
fix: cast float to numpy array in function call

olelod:
The compressor code expects numpy arrays, here a single float was passed.


TeeeJay (approved):
:+1:  can you also add missing/fix wrong type hints, if that was the case there?

PR #44, opened 2023-05-30
ci: fix issue with api refernce docs generation

markusrf:

The static html files generated by docusaurus can be slightly different in each build, which causes the link to the API reference docs page to sometimes not work.

The link to the API reference page has been changed from a relative link to a full URL which should be more stable


markusrf:
This works locally, but can not test properly until the docs are rebuilt and deployed which happens on merge

TeeeJay (approved):

PR #46, opened 2023-05-31
refactor: numpy ndarray typing

olelod:


TeeeJay (approved):
nice job. let's see where we end up wrt the granularity of type hints. Was there a mypy flag that could be checked/crossed out after this job? Or to remove numpy from an ignore list?

olelod:
> nice job. let's see where we end up wrt the granularity of type hints. Was there a mypy flag that could be checked/crossed out after this job? Or to remove numpy from an ignore list?

There are a lot less e.g. [type-arg] complaints, but they are not fully gone.

PR #47, opened 2023-05-31
feat: Improve simple example ipynb plotting

patnr:
Plot all components on the same axes/panel/figure, rather than creating new figure for each.

Also plot co2 emissions.

Match colors.

# Before:

# After:


TeeeJay:
Hi Patrick,

Thank you very much for your contribution to eCalc™, it is highly appreciated.

We will review your PR as soon as we have time and capacity to follow up.

Thank you for understanding.

Regards,
eCalc™ Team

PR #49, opened 2023-06-01
feat: output pump head to json-file

frodehk:

To save time in calibration and help diagnose pump issues: When calibrating or seeing why a pump is invalid, the most important thing to look at is the head. 

- [x] Extend eCalc-result object with pump operational head values
- [x] Write operational head to output json-file
- [x] Additional parameters written to json-file: pump inlet liquid rate, pump inlet pressure and pump outlet pressure
- [x] Operational head and additional parameters are also written to csv-file

https://github.com/equinor/ecalc-engine/issues/4400


src/ecalc/libraries/libecalc/common/libecalc/core/result/results.py
```diff
@@ -51,6 +51,7 @@ class PumpResult(GenericComponentResult):
     inlet_liquid_rate_m3_per_d: TimeSeriesRate
     inlet_pressure_bar: TimeSeriesFloat
     outlet_pressure_bar: TimeSeriesFloat
+    operational_head: Optional[List[float]]
```
TeeeJay, line 54: bruk timeseries som over.

TeeeJay: helst ikke optional, hvis mulig. er den optional?

src/ecalc/libraries/libecalc/common/libecalc/core/models/results/pump.py
```diff
@@ -10,3 +10,4 @@ class PumpModelResult(EnergyFunctionResult):
     suction_pressure: List[Optional[float]]
     discharge_pressure: List[Optional[float]]
     fluid_density: List[Optional[float]]
+    operational_head: Optional[List[float]]
```
TeeeJay, line 13: er denne nødvendig? kan være den er det, bare spør... skader ikke, men litt uheldig om det er dobbelt opp uten effekt

src/ecalc/libraries/libecalc/common/libecalc/core/result/results.py
frodehk: No need to be optional, will change and use timeseries as well.

src/ecalc/libraries/libecalc/common/libecalc/core/models/results/pump.py
TeeeJay: never mind, ser det nå ja

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
```diff
@@ -251,6 +251,10 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
                         .dict()
                         if consumer_result.component_result.power is not None
                         else None,
+                        "operational_head": consumer_result.component_result.operational_head
```
TeeeJay, line 254: denne skal ikke være her afaik

src/ecalc/libraries/libecalc/common/libecalc/core/result/results.py
TeeeJay: ser du har laga timeseries over rett, men her har du angitt at det er liste

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/legacy_consumer/component.py
```diff
@@ -183,12 +187,38 @@ def evaluate(
...
+            consumer_result = PumpResult(
                 id=self._consumer_dto.id,
                 timesteps=variables_map.time_vector,
                 is_valid=is_valid,
                 energy_usage=energy_usage_time_series,
                 power=power_time_series,
+                inlet_liquid_rate_m3_per_d=inlet_rate_time_series,
+                inlet_pressure_bar=inlet_pressure_time_series,
+                outlet_pressure_bar=outlet_pressure_time_series,
+                operational_head=list(consumer_function_result.energy_function_result.operational_head),
```
TeeeJay, line 221: timeseries...

src/ecalc/libraries/libecalc/common/libecalc/core/result/results.py
frodehk: Ja, den var liste - men oppdateres nå

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/legacy_consumer/component.py
```diff
@@ -183,12 +187,38 @@ def evaluate(
 
         elif self._consumer_dto.component_type == ComponentType.PUMP:
             # Using generic consumer result as pump has no specific results currently
-            consumer_result = GenericComponentResult(
+
+            inlet_rate_time_series = TimeSeriesRate(
```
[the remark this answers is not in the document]
olelod: No, I don't remember, sorry.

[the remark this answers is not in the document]
TeeeJay: @frodehk All these timeseries needs to be reindexed in order to make sure that we get the correct resulting time vector that we want, e.g. if the time vector for a pump consists of mid-year changes, but the user requests yearly resolution. It works well as long as all the changes are on the standard input time vector, but gets wrong and possibly a risk that we map incorrect value to incorrect timestep in the final resulting time vector. At least, that is what I seem to remember and understand from this.

Also, we should probably consider whether this functionality should be a part of the timeseries instead of remembering to have to deal with the vectors correctly for each and one of them...

Please correct me if I am wrong...

src/ecalc/libraries/libecalc/common/libecalc/dto/result/results.py
```diff
@@ -96,6 +96,10 @@ class GenericConsumerResult(EquipmentResultBase):
 
 class PumpResult(EquipmentResultBase):
     componentType: Literal[ComponentType.PUMP]
+    inlet_liquid_rate_m3_per_d: TimeSeriesRate
```
TeeeJay, line 99:
```suggestion
    inlet_liquid_rate_m3_per_day: TimeSeriesRate
```

please spell fully out. we have space and time and auto-complete ;) ! :)

TeeeJay (changes requested):
Very nice job. :+1: 

We need to be sure to handle reindexing correctly though. It seems that we do not have any tests with temporal models and in particular of pumps mid-year, which we should have. Then we will probably see the need for reindexing and possibly other gotchas. A quick test to this would be to take a test we already have and split it in 2-3 parts of temporal models, where one of them changes mid-year, but they are all the same model. the total result should then be correct, but model results will be a bit different in structure.

Also, when you merge this, use squash to get 1 commit and use `feat:` as you ahve already indicated with a user-friendly explanation of the consequence of the commit, as you basically have in the initial commit.

frodehk:
I will have a look into the re-indexing and try to add a test for the temporal models as you describe. There are a few tests on temporal models in test_ltp.py, but they are based on dtos and not a full yaml-setup (https://github.com/equinor/ecalc/blob/e5b896b9f46a7e13c6d806237c4d4bef44833b77/src/ecalc/libraries/libecalc/common/tests/output/results/test_ltp.py#L96).

frodehk:
@TeeeJay: 
- All pump timeseries are now reindexed
- The functionality is included in TimeSeries
- A test is added to check temporal models and reindexing: Split pump energy usage model in three temporal models, changing mid-year and end of year. The frequency is set to yearly.

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -230,6 +232,27 @@ def __setitem__(
...
 
+    def __reindex_time_vector__(
+        self,
+        new_time_vector: Iterable[datetime],
+        fillna: Union[float, str] = 0.0,
+    ) -> np.ndarray:
+        """Based on a consumer time function result (EnergyFunctionResult), the corresponding time vector and
+        the consumer time vector, we calculate the actual consumer (consumption) rate.
+        """
+        new_values: DefaultDict[datetime, Union[float, str]] = defaultdict(float)
```
TeeeJay, line 243: Why can there a a string-value here?

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -230,6 +232,27 @@ def __setitem__(
                     f"Could not update timeseries, Combination of indices of type '{type(indices)}' and values of type '{type(values)}' is not supported"
                 )
 
+    def __reindex_time_vector__(
+        self,
+        new_time_vector: Iterable[datetime],
+        fillna: Union[float, str] = 0.0,
```
TeeeJay, line 238: here as well...string?

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -343,6 +366,10 @@ def resample(self, freq: Frequency) -> Self:
             unit=self.unit,
         )
 
+    def reindex(self, new_time_vector):
```
TeeeJay, line 369: typehints missing

TeeeJay: and docstring..or is that provided in abstract class?

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
frodehk: This reindex function is a pure copy of the reindex_time_vector in legacy_consumer/component.py (also used to reindex compressors) - have not checked in details, so not sure why string is allowed here.

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -230,6 +232,27 @@ def __setitem__(
                     f"Could not update timeseries, Combination of indices of type '{type(indices)}' and values of type '{type(values)}' is not supported"
                 )
 
+    def __reindex_time_vector__(
```
TeeeJay, line 235: looks good. But why make it a dunder method? This usually indicates some kind of magic invocation using an operator etc.

We should also further refactor compressors to use this method instead of the static method and the bit cumbersome way that they handle reindexing. If there is a unit test for reindexing etc there, then reuse that for this, otherwise we must create unit tests for this reindex method. That can be done in a follow-up PR though with focus on making the compressor and pump reindexing (and others) work the same way).

TeeeJay (approved):
nice, a few minor comments.

did you manage to verify that without the reindexing it broke, or was the test data not good enough for that? Did you figure out when you introduced the new test data with temporal models?

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
TeeeJay: ok. Difficult to know if it was a reason in the past, or if it is still relevant. I think it might have been at the time when units were provided for each value, but not sure. I do not see any problems with that, but would be nice to know why. We should also ahve test for that wrt. regression, since if we change the functionality, it may easily break the "string-part" of reindexing imo, because one forgets to think about that part.

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
frodehk: It should not be a dunder method, will fix:-)

frodehk:
> nice, a few minor comments.
> 
> did you manage to verify that without the reindexing it broke, or was the test data not good enough for that? Did you figure out when you introduced the new test data with temporal models?

I checked, it did not break without reindexing. We may need to adjust the case to stress-test?

TeeeJay:
> I checked, it did not break without reindexing. We may need to adjust the case to stress-test? One thing to check: Input dates-frequencies on timeseries rates etc for the actual case. Will do.

That is exactly what I had in mind now. That should hopefully verify well :)

PR #55, opened 2023-06-01
ci: set default ownership for source

TeeeJay:
github-actions and a few other meta directories did not get automatic review request due to missing ownership

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,8 @@
 # Each line is a file pattern followed by one or more owners.
 
+# Default/fallback ownership
+* @equinor/ecalc-core-library-team
+
 # eCalc Lib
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,5 +1,8 @@
 # Each line is a file pattern followed by one or more owners.
 
+# Default/fallback ownership
+* @equinor/ecalc-core-library-team
+
 # eCalc Lib
 /src/ecalc/ @equinor/ecalc-core-library-team
```
jorgenengelsen, line 7:
```suggestion
```

guess this is not needed anymore then

jorgenengelsen (approved):

.github/CODEOWNERS
TeeeJay: I want to keep it. We may introduce a different team for fallback instead actually...

PR #59, opened 2023-06-05
chore: do not resample model results, only component and sub-component results

olelod:


TeeeJay:
@olelod any new thoughts on this? Do you remember why this was brought up?

olelod:
> @olelod any new thoughts on this? Do you remember why this was brought up?

No, not really. And it does not seems to be a problem of any sort, so I am just closing it.

PR #63, opened 2023-06-06
fix: extend lists within lists in compressor train results with multiple streams properly

olelod:
A VariableSpeedCompressorTrainCommonShaftMultipleStreamsAndPressures have some results (or more specifically return of rate input) that are lists within lists. The extend method does not handle these results properly. Currently the extend method will add more lists within a list, but it should extend each of the lists inside a list instead.


TeeeJay (approved):
Nice. Squash commit to use the `fix:` conventional commit handle, and create a user friendly explanation in one line that it was a bugfix to temporal models in multi streams. Then we can use that line directly in changelog. Less work :)

Should we consider patching it on v8.2.3? (release/v8.2 branch?)

TeeeJay:
Ah yes, also update the changelog manually (using the same text as in the commit?) until we are sure that we are good and consistent and can use the git commit changelog directly :)

PR #65, opened 2023-06-07
chore: simplify dependencies for use with komodo

TeeeJay:
typer optionally depends on shellingham, rich and
colorama. Rich is already an explicit dependency in ecalc, while colorama is a dependency through click and
already in komodo.

By only depending on typer and not typer[all] we remove the need for shellingham, as that dependency should not be needed in komodo. It is used to install auto-completion for ecalc, and it automatically detects the shell. If the user still wants auto-completion, they will need to specify shell explicitly with the argument.

We should aim to keep the number of dependencies to a minimum, in general, but in particular with Komodo, in order to make it easier to maintain.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2692,18 +2692,6 @@ docs = ["matplotlib", "numpydoc (>=1.1.0,<1.2.0)", "sphinx", "sphinx-book-theme"
 test = ["pytest", "pytest-cov"]
 
 [[package]]
-name = "shellingham"
-version = "1.4.0"
-description = "Tool to Detect Surrounding Shell"
-category = "main"
-optional = false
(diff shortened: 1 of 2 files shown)


jorgenengelsen (approved):
In the future I think we could consider it as a extra to libecalc:

libecalc[cli] 
libecalc[all]

Some user might not want jupyter, typer, or matplotlib either

TeeeJay:
> In the future I think we could consider it as a extra to libecalc:
> 
> libecalc[cli] libecalc[all]
> 
> Some user might not want jupyter, typer, or matplotlib either

I agree. We were initially thinking that for this version, but tried initially to make it less complicated to understand.

PR #66, opened 2023-06-07
fix: do not validate operational conditions when rate is zero

olelod:

User experienced extrapolation of results when CONDITION(S) imply that no calculations should be done.

When CONDITION(S) imply that no calculations should be performed for a model, the rate(s) are set to zero for that time step. The time step should still be valid - we do not want extrapolation of results. This means that results for time steps with zero rate(s) should always be considered as valid.

This PR changes the validation of the operational conditions (rate(s), suction/discharge pressures) such that time steps with zero rate(s) are always considered valid, and no failure status will be returned for that time step (even if pressures are zero or negative). Any invalid pressures will still be set to 1, because the may be used in a call to neqsim.

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/base.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/base.py
@@ -54,6 +54,73 @@ class CompressorModel(BaseModel):
         """
         raise NotImplementedError
 
+    def validate_operational_conditions(
+        self,
+        rate: NDArray[np.float64],
+        suction_pressure: NDArray[np.float64],
+        discharge_pressure: NDArray[np.float64],
+        intermediate_pressure: Optional[NDArray[np.float64]] = None,
+    ) -> Tuple[
+        NDArray[np.float64],
+        NDArray[np.float64],
+        NDArray[np.float64],
+        NDArray[np.float64],
+        List[CompressorTrainCommonShaftFailureStatus],
+    ]:
+        indices_to_validate = self._find_indices_to_validate(rate=rate)
+        validated_failure_status = [None] * len(suction_pressure)
+        validated_rate = rate.copy()
+        validated_suction_pressure = suction_pressure.copy()
+        validated_discharge_pressure = discharge_pressure.copy()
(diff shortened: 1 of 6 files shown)


src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/base.py
```diff
@@ -98,6 +100,10 @@ def _validate_operational_conditions(
             None,
         ]
 
+        # calculations are performed only if rate is positive.
+        # do not check validity of operational conditions if rate is zero
+        check_validity_for_time_step = np.any(rate > 0, axis=0) if np.ndim(rate) == 2 else rate > 0
```
TeeeJay, line 105: denne vil returnere hva? enkel bool enner ndarray? ndarray med indekser  for hvilke som evaluerer til henholdsvis true og false?

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/base.py
```diff
@@ -113,7 +119,7 @@ def _validate_operational_conditions(
                 np.where(np.any(rate < 0, axis=0), 0, rate) if np.ndim(rate) == 2 else np.where(rate < 0, 0, rate)
             )  # if the rate for one stream is negative, set the rates for all streams to zero for that time step
         if intermediate_pressure is not None:
-            if not np.all(intermediate_pressure > 0):
+            if not np.all(intermediate_pressure[check_validity_for_time_step] > 0):
```
TeeeJay, line 122: How is `check_validity_for_time_step` used here? Not sure I understand what is happening here

TeeeJay (approved):
Seems to do the job! :+1:  But the logic is a bit to intricate to me. Is it possible to extract it into a separate function and add a unit test to more clearly show how it works? A bit too much numpy magic for my understanding as well, even though it seems to be correct.

In the light of our current bugs, I think it is very important that we ofte wrap numpy functionality and test edge cases to show to ourself and ours that we understand how numpy works instead of hoping for the best on a few happy path cases :)

Please let me know what you think! :)

jorgenengelsen (approved):

PR #68, opened 2023-06-08
fix: resample emissions correctly to create valid json

TeeeJay:
The json schema before and after resampling differed. The resampling method to handle dicts was incorrect. There are no other data that is using dicts so this bug only occured after changing the structure to use dicts for emissions.


TeeeJay:
will add tests

src/ecalc/libraries/libecalc/common/libecalc/dto/result/tabular_time_series.py
```diff
@@ -51,6 +51,15 @@ def to_dataframe(
...
 
     def resample(self, freq: Frequency) -> Self:
+        """
+        Immutable - returns a copy of itself
+
+        Resample the given time series to the new Frequency given. Only data
+        that is defined as a timeseries will be resampled.
+
+        :param freq:
+        :return: return a copy of itself with all data resampled to given frequency
```
jorgenengelsen, line 61: Did we decide on using google doc-string format?

Pycharm does it automatically for me, I can show you if you would like

src/ecalc/cli/tests/test_app.py
```diff
@@ -526,6 +526,25 @@ def test_csv_resampled(self, simple_run, monkeypatch, snapshot):
         output_text = result.stdout
         snapshot.assert_match(output_text, snapshot_name="results_resampled.csv")
 
+    @pytest.mark.snapshot
+    def test_json_resampled(self, simple_run, monkeypatch, snapshot):
```
jorgenengelsen, line 530: Is a snapshot test needed here? Could it be tested without adding more json snapshots that need to be updated?

jorgenengelsen (approved):

src/ecalc/libraries/libecalc/common/libecalc/dto/result/tabular_time_series.py
TeeeJay: Didnt make any changes since we discussed this, so forgot to change setup in pycharm.

Does interrogate and/or our scripts handle both or make sure we are consistent somehow?

src/ecalc/cli/tests/test_app.py
TeeeJay: as mentioned, I think so. But I can and should specify why in this test instead of just adding it

PR #70, opened 2023-06-12
chore: update NeqSim version to 2.5.1

jorgenengelsen:

This pull request is needed to keep NeqSim updated. Potentially speed up performance

Updates NeqSim jar to 2.5.1, and updates tests


src/ecalc/libraries/libecalc/common/tests/core/models/compressor_modelling/test_simplified_compressor_train.py
```diff
@@ -963,8 +965,8 @@ def test_calculate_enthalpy_change_head_iteration_and_outlet_stream(dry_fluid):
...
     ]
 
-    np.testing.assert_allclose(expected_inlet_z, [s.z for s in inlet_streams])
-    np.testing.assert_allclose(expected_outlet_z, [s.z for s in outlet_streams])
-    np.testing.assert_allclose(expected_inlet_kappa, [s.kappa for s in inlet_streams])
-    np.testing.assert_allclose(expected_outlet_kappa, [s.kappa for s in outlet_streams])
-    np.testing.assert_allclose(expected_enthalpy_change, enthalpy_change_joule_per_kg)
+    np.testing.assert_allclose(expected_inlet_z, [s.z for s in inlet_streams], rtol=1e-5)
+    np.testing.assert_allclose(expected_outlet_z, [s.z for s in outlet_streams], rtol=1e-4)
+    np.testing.assert_allclose(expected_inlet_kappa, [s.kappa for s in inlet_streams], rtol=1e-3)
```
jorgenengelsen, line 970: Had to go down to 1e-3 in relative tolerance for kappa. Not sure why this change is that big

src/ecalc/libraries/neqsim/tests/integration_tests/test_not_threadsafe.py
```diff
@@ -14,16 +15,17 @@
...
     assert Fluid.isHasWater() is not has_water
     Fluid.setHasWater(has_water)
 
 
+@pytest.mark.xfail(reason="No longer a static variable in NeqSim, could remove test if not new case found")
 def test_run_neqsim_in_2_threads():
     neqsim = java_gateway.jvm.neqsim
     # Initially, has water is false
-    Fluid = neqsim.thermo.Fluid
+    Fluid = neqsim.thermo.Fluid()
```
jorgenengelsen, line 28: isHasFluid is no longer a static parameter, and can't be accessed directly

src/ecalc/libraries/libecalc/common/tests/core/models/compressor_modelling/test_simplified_compressor_train.py
TeeeJay: any chance of asking Even if he knows? possibly a bugfix etc? It said in documentation how old our Neqsim was...so I guess a few bugfixes and improvements since then...

src/ecalc/libraries/neqsim/neqsim_ecalc_wrapper/lib/neqsim_version_info.md
```diff
@@ -1,4 +1,4 @@
-Current neqsim.jar from latest release of NeqSim (v2.3.0 - https://github.com/equinor/neqsim/releases/tag/v2.3.0)
+Current neqsim.jar from latest release of NeqSim (v2.5.1 - https://github.com/equinor/neqsim/releases/tag/v2.3.0)
```
TeeeJay, line 1:
```suggestion
Current neqsim.jar from latest release of NeqSim (v2.5.1 - https://github.com/equinor/neqsim/releases/tag/v2.5.1)
```
?

src/ecalc/libraries/neqsim/tests/integration_tests/test_not_threadsafe.py
TeeeJay: can be replaced with another static variable or static set/get method, if there is any. it was used as a simple example demonstrating that 2 threads would lead to race conditions, to avoid anyone in future to remove the lock in the code, which would in 99% of the cases silently deliver incorrect results and only sometimes fail...

TeeeJay:
So, as far as I can see the changes are not that big considering Neqsim in isolation, but the total changes may be relevant to consider.

Users must be given heads up. Difficult to know how big changes this will lead to for all models, even though it is more correct. Users may have to recalibrate again?

Maybe we can test this first on a few relevant models to see how it changes in practice?

@olelod what do you think? any ideas? Or is this completely fine merge? Just saying that we upgraded neqsim...

src/ecalc/libraries/libecalc/common/tests/core/models/compressor_modelling/test_simplified_compressor_train.py
jorgenengelsen: tbf, it is still <1% change, so I think it is quite negligible.

From NeqSimAPI, most changes come from updated database entries. Could try to first update to 2.4.13 (before database change), to see if results are more similar there, but not sure it is worth the work

olelod:
> So, as far as I can see the changes are not that big considering Neqsim in isolation, but the total changes may be relevant to consider.
> 
> Users must be given heads up. Difficult to know how big changes this will lead to for all models, even though it is more correct. Users may have to recalibrate again?
> 
> Maybe we can test this first on a few relevant models to see how it changes in practice?
> 
> @olelod what do you think? any ideas? Or is this completely fine to merge? Just saying that we upgraded neqsim, and we must trust it.

Probably a good idea to check the actual impact on a real case. I agree.

PR #75, opened 2023-06-13
chore: numpy ndarray typing

olelod:
Patch of numpy typing onto release branch


jorgenengelsen (changes requested):
Is this needed? Typing only helps developers

olelod:
> Is this needed? Typing only helps developers

Another patch was made after this was included (https://github.com/equinor/ecalc/pull/74) but I can change to np.ndarray for that patch if it is a problem.

PR #77, opened 2023-06-13
chore: improve error message when bad yaml file name

frodehk:

Improve error message when Yaml model file name is wrong, e.g. includes illegal special characters. 

- [x] Add check for illegal characters in the pyyaml_yaml_model.py (load). The check is now done when loading the file, instead of after reading and verifying all data.
- [x] Add test to verify failure when loading Yaml file with illegal name, and to verify improved error message.

https://github.com/equinor/ecalc-engine/issues/4498


src/ecalc/libraries/libecalc/common/libecalc/input/yaml/yaml_models/pyyaml_yaml_model.py
```diff
@@ -125,6 +126,12 @@ def __init__(
                 )
 
         def load(self, yaml_file: ResourceStream):
+            if ComponentNameStr.regex.search(Path(yaml_file.name).stem) is None:
```
jorgenengelsen, line 129: Some regex magic I don't understand here? What regex pattern is used?

frodehk: https://github.com/equinor/ecalc/blob/830c75e27a29549157658c606e618da381c24e81/src/ecalc/libraries/libecalc/common/libecalc/dto/utils/validators.py#L9

TeeeJay: Imo you should write which characters that are allowed using the regexp/string we have already

src/ecalc/cli/tests/test_app.py
```diff
@@ -640,3 +646,36 @@ def test_full_simplified_json(self, simple_run, monkeypatch, snapshot):
 
         output_text = result.stdout
         snapshot.assert_match(output_text, snapshot_name="results.json")
+
+
+class TestYamlFile:
+    def test_yaml_file_error(self, simple_wrong_name_yaml_path, tmp_path):
```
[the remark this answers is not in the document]
frodehk: Yes, I was planning to do that

src/ecalc/libraries/libecalc/common/libecalc/input/yaml/yaml_models/pyyaml_yaml_model.py
frodehk: ComponentNameStr is the regex we have already, it is imported from validators. Or do I misunderstand your point?

TeeeJay: The error message here will easily go out of sync with the regex. What if we include $ as acceptable in component name? We will forget to update the error message. Imo there is not really any good way to present this to users (there are libraries that create human friendly descriptions of reg exp tho), but is this an option or sth along those lines:

```
COMPONENT_NAME_ALLOWED_CHARS = 'A-ZÆØÅa-zæøå\d_/\- '
ComponentNameStr = constr(regex=r"^[" + COMPONENT_NAME_ALLOWED_CHARS + "]*$")
...
error_message = f"The eCalc Model filename contains illegal characters. Allowed characters are: {COMPONENT_NAME_ALLOWED_CHARS}"
```

frodehk: Or should we just have a separate regex for the filename, independent of what is allowed for components? I don´t see why they should be identical?

TeeeJay: It is because we do not have another identificator for the whole eCalc Model (file) other than the filename. In order to reference the total results, we use the filename for that. This is represented as a componet in the results. Alternatives are e.g.:

* could just convert the filename to something that only has accepted characters
* introduce an attribute in the yaml file where use must specify "name of eCalc model"

frodehk: > 'A-ZÆØÅa-zæøå\d_/\- '

I think many users will get confused by the 'A-ZÆØÅa-zæøå\d_/\- '. Isn´t it better to write that the `filename contains illegal special characters` - as long as the filename is included in the error message, the user knows where to look?

frodehk:
@TeeeJay and @jorgenengelsen: Can you have a look now? The error message is changed, removed example with ´$´, ´.´ etc. and included the allowed characters. In addition, the test is simplified (no separate yaml-file, just testing the name).

src/ecalc/cli/tests/test_app.py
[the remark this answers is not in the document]
frodehk: Done

src/ecalc/libraries/libecalc/common/libecalc/input/yaml/yaml_models/pyyaml_yaml_model.py
TeeeJay: I agree. No good solution. A bit reluctant to introduce a new big library for "such a small thing as well".

TeeeJay: https://www.autoregex.xyz/

frodehk: I added the allowed characters, as suggested above. Maybe it is ok, as long as we also communicate that the model file name is wrong - and include the filename in the error message?

TeeeJay (approved):
yes, sorry. forgot to approve :)

PR #78, opened 2023-06-13
fix: make apply_condition work for 2D numpy arrays also

olelod:
`apply_condition` does not work for multiple streams. If condition(s) are not met, rates for all streams at a time step should be set to 0. This PR fixes that.


jorgenengelsen (approved):
Det enkle er ofte det beste? 😄

PR #83, opened 2023-06-20
Refactor/variable speed compressors

jorgenengelsen:

This pull request is needed to update and improve documentation of variable speed compressors

### DIFF
--- a/src/ecalc/cli/cli/commands/run.py
+++ b/src/ecalc/cli/cli/commands/run.py
@@ -108,11 +108,10 @@ def run(
 
     model = YamlModel(path=model_file, output_frequency=output_frequency)
 
-    if flow_diagram or ltp_export:
-        if model.start is None or model.end is None:
-            logger.warning(
-                "When using Flow Diagram or Long Term Prognosis export, START and END should be specified in YAML to make sure you get the intended period as output. See documentation for more information."
-            )
+    if (flow_diagram or ltp_export) and (model.start is None or model.end is None):
+        logger.warning(
+            "When using Flow Diagram or Long Term Prognosis export, START and END should be specified in YAML to make sure you get the intended period as output. See documentation for more information."
+        )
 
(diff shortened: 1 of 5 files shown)


src/ecalc/cli/cli/commands/run.py
```diff
@@ -108,11 +108,10 @@ def run(
 
     model = YamlModel(path=model_file, output_frequency=output_frequency)
 
-    if flow_diagram or ltp_export:
-        if model.start is None or model.end is None:
-            logger.warning(
-                "When using Flow Diagram or Long Term Prognosis export, START and END should be specified in YAML to make sure you get the intended period as output. See documentation for more information."
-            )
+    if (flow_diagram or ltp_export) and (model.start is None or model.end is None):
```
TeeeJay, line 111: mypy or linter that you use that suggest this? I guess I agree though, a bit easier to read :)

TeeeJay (approved):
saw a bit late that it was 2 commits and it made sense to check those separately wrt 2 different logical changes in this PR, so would be nice to get a hint next time :).

great improvements though :)

jorgenengelsen:
> saw a bit late that it was 2 commits and it made sense to check those separately wrt 2 different logical changes in this PR, so would be nice to get a hint next time :).
> 
> great improvements though :)

Sorry about that, I see now that I could be more clear.

Tried to understand variable-speed compressor code, Used Ruff rule set "SIM", which automatically suggests "simplifications" to the code.

TeeeJay:
> > saw a bit late that it was 2 commits and it made sense to check those separately wrt 2 different logical changes in this PR, so would be nice to get a hint next time :).
> > great improvements though :)
> 
> Sorry about that, I see now that I could be more clear.
> 
> Tried to understand variable-speed compressor code, Used Ruff rule set "SIM", which automatically suggests "simplifications" to the code.

thanks :)

PR #84, opened 2023-06-20
chore: enforce unique fuel type names, and unique emission names within one fuel type

frodehk:

- Names for `FUEL_TYPE` must be unique across installations in an eCalc model file
- Names for `EMISSIONS` must be unique within one `FUEL_TYPE`

Duplicate names may lead to debug problems e.g. for vector lengths etc. It may also override previous value without the user noticing (e.g. when Ctrl-C Ctrl-V syntax)

- [x] Add extra verification for duplicate fuel type names in Asset-class in components.py (library)
- [x] Add extra verifications in create_references.py for duplicate fuel type names in Yaml-file
- [x] Add extra verifications in create_references.py for duplicate emission names in Yaml-file
- [x] Add test for duplicate fuel type names using dto and library (no yaml)
- [x] Add test for duplicate fuel type names using yaml-file
- [x] Add test for duplicate emission names within one fuel, using yaml-file

https://github.com/equinor/ecalc-engine/issues/3401

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
@@ -456,8 +456,8 @@ class Asset(Component):
     def validate_unique_names(cls, values):
         """Ensure unique component names within installation."""
         names = [values["name"]]
-        fuel_types = []
-        fuel_names = []
+        fuel_types = [dto.FuelType]
+        fuel_names = [str]
         for installation in values["installations"]:
             names.append(installation.name)
             fuel_consumers = installation.fuel_consumers
--- a/src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
@@ -457,6 +457,7 @@ class Asset(Component):
         """Ensure unique component names within installation."""
         names = [values["name"]]
         fuel_types = []
+        fuel_names = []
         for installation in values["installations"]:
             names.append(installation.name)
             fuel_consumers = installation.fuel_consumers
@@ -473,14 +474,22 @@ class Asset(Component):
                         # Need to verify that it is a different fuel
                         if fuel_type is not None and fuel_type not in fuel_types:
                             fuel_types.append(fuel_type)
-                            names.append(fuel_type.name)
+                            fuel_names.append(fuel_type.name)
 
         duplicated_names = get_duplicates(names)
+        duplicated_fuel_names = get_duplicates(fuel_names)
+
         if len(duplicated_names) > 0:
             raise ValueError(
-                "Component- and fuel type names must be unique. Components include asset/ecalc-model, installations,"
-                " generator sets, electricity consumers, fuel consumers, direct emitters and fuel types."
+                "Component names must be unique. Components include asset/ecalc-model, installations,"
+                " generator sets, electricity consumers, fuel consumers and direct emitters."
                 f" Duplicated names are: {', '.join(duplicated_names)}"
+            )
+
+        if len(duplicated_fuel_names) > 0:
+            raise ValueError(
+                "Fuel type names must be unique across installations."
+                f" Duplicated names are: {', '.join(duplicated_fuel_names)}"
             )
         return values
 
--- a/src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
@@ -456,6 +456,7 @@ class Asset(Component):
     def validate_unique_names(cls, values):
         """Ensure unique component names within installation."""
         names = [values["name"]]
+        fuel_types = []
         for installation in values["installations"]:
             names.append(installation.name)
             fuel_consumers = installation.fuel_consumers
@@ -467,12 +468,18 @@ class Asset(Component):
                 if isinstance(fuel_consumer, GeneratorSet):
                     for electricity_consumer in fuel_consumer.consumers:
                         names.append(electricity_consumer.name)
+                if fuel_consumer.fuel is not None:
+                    for fuel_type in fuel_consumer.fuel.values():
+                        # Need to verify that it is a different fuel
+                        if fuel_type is not None and fuel_type not in fuel_types:
+                            fuel_types.append(fuel_type)
+                            names.append(fuel_type.name)
 
         duplicated_names = get_duplicates(names)
         if len(duplicated_names) > 0:
             raise ValueError(
-                "Component names must be unique. Components include asset/ecalc-model, installations,"
-                " generator sets, electricity consumers, fuel consumers and direct emitters."
+                "Component- and fuel type names must be unique. Components include asset/ecalc-model, installations,"
+                " generator sets, electricity consumers, fuel consumers, direct emitters and fuel types."
                 f" Duplicated names are: {', '.join(duplicated_names)}"
             )
         return values
--- a/src/ecalc/libraries/libecalc/common/libecalc/examples/simple/model_duplicate_names.yaml
+++ b/src/ecalc/libraries/libecalc/common/libecalc/examples/simple/model_duplicate_names.yaml
@@ -1,0 +1,111 @@
+TIME_SERIES:
+  - NAME: SIM
+    FILE: production_data.csv
+    TYPE: DEFAULT
+FACILITY_INPUTS:
+  - NAME: genset
+    FILE: genset.csv
+    TYPE: ELECTRICITY2FUEL
+  - NAME: compressor_sampled
+    FILE: compressor_sampled.csv
+    TYPE: COMPRESSOR_TABULAR
+  - NAME: compressor_with_turbine_sampled
+    FILE: compressor_sampled_with_turbine.csv
+    TYPE: COMPRESSOR_TABULAR
+  - NAME: pump_sampled
+    FILE: pump_sampled.csv #!include, commented !include is now supported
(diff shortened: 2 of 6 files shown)


src/ecalc/libraries/libecalc/common/tests/dto/test_fuel_consumer.py
```diff
@@ -21,23 +103,83 @@ def test_missing_fuel(self):
...
+        """
+        TEST SCOPE: Check that validation of duplicate fuel type names do not reject
+        when same fuel is used across installations.
+
+        Even though duplicate fuel type names are not allowed across installations,
+        it should be possible to re-use the same fuel. This test verifies that this still
+        works.
+        """
+
+        fuel_consumer1 = get_fuel_consumer(
```
TeeeJay, line 156: I think that this example here is somehow unproblematic. This can be a problem that the user names a fuel with the same name twice, but they are both accessible. The main problem here I guess is that it can be confusing wrt reporting and aggregating "emissions by fuel_name", then that would cause a problem.

src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
TeeeJay, line 1: You did find another possible bug to fuels that have duplicate names, but this does not solve the problem when there are > 1 fueltype in the yaml file with the same name/key I think. In that case, only the latter one will proceed to be processed, while the first one will be silently ignored.

The use case may be that the user copy paste the same fueltype in the yaml, and forgets to change the name of the fueltype. It has happened in the past.

src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
```diff
@@ -467,14 +469,28 @@ def validate_unique_names(cls, values):
...
         duplicated_names = get_duplicates(names)
+        duplicated_fuel_names = get_duplicates(fuel_names)
+
         if len(duplicated_names) > 0:
             raise ValueError(
                 "Component names must be unique. Components include asset/ecalc-model, installations,"
                 " generator sets, electricity consumers, fuel consumers and direct emitters."
                 f" Duplicated names are: {', '.join(duplicated_names)}"
             )
+
```
TeeeJay, line 488: I do think that this check is needed though, but we need an additional check/validation at yaml level.

src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
```diff
@@ -467,14 +469,28 @@ def validate_unique_names(cls, values):
                 if isinstance(fuel_consumer, GeneratorSet):
                     for electricity_consumer in fuel_consumer.consumers:
                         names.append(electricity_consumer.name)
+                if fuel_consumer.fuel is not None:
+                    for fuel_type in fuel_consumer.fuel.values():
+                        # Need to verify that it is a different fuel
+                        if fuel_type is not None and fuel_type not in fuel_types:
```
TeeeJay, line 475: can fuel_type be None?

TeeeJay: why only add fuel_name if fuel_type not already added?

src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
```diff
@@ -467,14 +469,28 @@ def validate_unique_names(cls, values):
...
+                            fuel_types.append(fuel_type)
+                            fuel_names.append(fuel_type.name)
 
         duplicated_names = get_duplicates(names)
+        duplicated_fuel_names = get_duplicates(fuel_names)
```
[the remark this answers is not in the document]
frodehk: This check is actually doing that, checking if it is a different fuel. The fuel_types list includes the whole fuel, not only the name. So, this will catch situations where e.g. 2 fuels are called fuel 1 for co2 but some other values are different:

https://github.com/equinor/ecalc/blob/c41166fc458636aa35748a9438fa2db4e5bdca81/src/ecalc/libraries/libecalc/common/libecalc/dto/components.py#L475

As far as I understand, the only thing we don´t catch is the situation where the user has duplicated the fuel and forgets to change parameters.

src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
frodehk: A test failed due to fuel_type None, don´t remember what caused it.

Adding fuel_name only if it is a new fuel - it should be possible to re-use the same fuel for several consumers/installations. So, the duplicate name check is only done if it is a different/new fuel (e.g. some fuel parameters are different)

src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
frodehk: Ok, you mean if there are more identical fuels in the yaml-file? I agree, that is not captured. But if there are > 1 fuel types in the yaml with same name, but different parameters, it will be captured.

src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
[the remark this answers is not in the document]
frodehk: I have added a check for duplicate fuels in yaml. This should capture the copy-paste issue where the user forgets to change name etc.:

https://github.com/equinor/ecalc/blob/5fceec63355229a9179ec740e12184f87e83a986/src/ecalc/libraries/libecalc/common/libecalc/input/mappers/create_references.py#LL31C1-L31C1

Also added a test with yaml file containing duplicate fuels:
https://github.com/equinor/ecalc/blob/5fceec63355229a9179ec740e12184f87e83a986/src/ecalc/cli/tests/test_app.py#L683

src/ecalc/libraries/libecalc/common/libecalc/examples/simple/model_duplicate_names.yaml
```diff
@@ -0,0 +1,111 @@
...
+    TYPE: TABULAR
+  - NAME: pump_chart
+    FILE: pump_chart.csv
+    TYPE: PUMP_CHART_SINGLE_SPEED
+    UNITS:
+      HEAD: M
+      RATE: AM3_PER_HOUR
+      EFFICIENCY: PERCENTAGE
+
+FUEL_TYPES:
```
TeeeJay, line 26: is it only fuel_gas that is tested here? can the yaml test be made smaller to not include a lot other data that is not needed for the test?

frodehk: Yes only fuel_gas is tested. Have simplified the yaml-file now.

Also added verification for duplicate emissions within one fuel type, for the yaml.

src/ecalc/libraries/libecalc/common/libecalc/dto/components.py
```diff
@@ -456,6 +456,8 @@ def get_installation(self, installation_id: str) -> Installation:
     def validate_unique_names(cls, values):
         """Ensure unique component names within installation."""
         names = [values["name"]]
+        fuel_types = []
```
TeeeJay, line 459: should specify what type it is with type hints here :)

src/ecalc/libraries/libecalc/common/libecalc/input/mappers/create_references.py
```diff
@@ -26,6 +27,31 @@ def create_references(configuration: PyYamlYamlModel, resources: Resources) -> R
...
+        raise ValueError(
+            "Fuel type names must be unique across installations."
+            f" Duplicated names are: {', '.join(duplicated_fuel_names)}"
+        )
+
+    fuel_types_emissions = [list(fuel_data[EcalcYamlKeywords.emissions]) for fuel_data in configuration.fuel_types]
+
+    # Check each fuel for duplicated emissions
+    duplicated_emissions = []
+    for emissions in fuel_types_emissions:
```
TeeeJay, line 45: instead of grabbing all failures here, I would perhaps want to just break on the first duplicate we find, to make the logic easier to grasp. Should not happen that often after all...

src/ecalc/libraries/libecalc/common/tests/dto/test_fuel_consumer.py
```diff
@@ -21,23 +103,83 @@ def test_missing_fuel(self):
...
+                    installation2,
+                ],
             )
-        assert "Missing fuel for fuel consumer 'test'" in str(exc_info.value)
+
+        assert "Duplicated names are: fuel1" in str(exc_info.value)
+
+    def test_same_fuel(self):
+        """
+        TEST SCOPE: Check that validation of duplicate fuel type names do not reject
```
TeeeJay, line 148: well thought of :)

src/ecalc/libraries/libecalc/common/tests/dto/test_fuel_consumer.py
TeeeJay, line 1: missing a test on duplicate emissions . i can see that the yaml file is there. forgot to include perhaps?

TeeeJay (approved):
nice, just add the missing test :)

src/ecalc/libraries/libecalc/common/tests/dto/test_fuel_consumer.py
frodehk: This one you mean?
https://github.com/equinor/ecalc/blob/e54b3c5437a05f3817c8e943a75218f0a0d642f9/src/ecalc/cli/tests/test_app.py#L713

TeeeJay: ah ok, didnt see it :)

PR #85, opened 2023-06-21
fix: parse scientific notation numbers in expression

markusrf:
The expression lexer was missing parsing for scientific notation (e.g. `1.23e-05` or `3.4e7`) for numbers part of an expressions with multiple tokens.

Ex. expression `2 {*} 3.4e-05` would be parsed as tokens `[2, {*}, 3.4, e-05]`, which would give issues as there are no variables named e-05.

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/expression/expression_evaluator.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/expression/expression_evaluator.py
@@ -325,6 +325,7 @@ def lexer(expression: Union[str, int, float]) -> List[Token]:
         (r"while", TokenTag.operator),
         (r"do", TokenTag.operator),
         (r"end", TokenTag.operator),
+        (r"[0-9](\.[0-9]+)?e[-+]?[0-9]+", TokenTag.numeric),  # Scientific notation, e.g. 1.23e-05, 3.4e7, 1e+1
         (r"[0-9.]+", TokenTag.numeric),
         (r"[A-Za-z][A-Za-z0-9._;:+*/-]*", TokenTag.reference),
         (r"\$var\.[A-Za-z][A-Za-z0-9_]*", TokenTag.reference),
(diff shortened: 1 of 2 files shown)


src/ecalc/libraries/libecalc/common/libecalc/expression/expression_evaluator.py
```diff
@@ -325,6 +325,7 @@ def lexer(expression: Union[str, int, float]) -> List[Token]:
         (r"while", TokenTag.operator),
         (r"do", TokenTag.operator),
         (r"end", TokenTag.operator),
+        (r"[0-9]\.[0-9]+e-?[0-9]+", TokenTag.numeric),  # Scientific notation, e.g. 1.23e-05 or 3.4e7
```
TeeeJay, line 328: what about 1e-8? It would. Can this happen to big numbers as well? In that case 3e29 and 3e+29?

src/ecalc/libraries/libecalc/common/libecalc/expression/expression_evaluator.py
```diff
@@ -356,7 +357,7 @@ def check_tokens(tokens):
     if str(first_token) in list(OPERATORS.keys()):
         raise ValueError(f"Expression ({var}) can not start with an operator")
     if str(last_token) in list(OPERATORS.keys()):
-        raise ValueError(f"Expression ({var}) can not start with an operator")
+        raise ValueError(f"Expression ({var}) can not end with an operator")
```
TeeeJay, line 360: :D

src/ecalc/libraries/libecalc/common/tests/expression/test_expression_evaluator.py
```diff
@@ -93,6 +93,12 @@ def test_lexer():
     token_values = [token.value for token in tokens]
     assert token_values == compare
 
+    expression = "2.4e6 {/} 100 {*} 1.35e-05"
```
TeeeJay, line 96: I would make a separate test. I see that we have fewer tests here than anticipated. Make sure you cover all edge cases. A good idea is to add parametric tests (my preference) to test both the case with scientific numbers alone, in an expression with other scientific numbers and combined with other tokentypes :)

markusrf: Have updated with more parametric tests now

TeeeJay (approved):
nice :)

PR #87, opened 2023-06-21
chore: add installation filter to flare nmvoc

frodehk:

All flare variables reported to ltp have installation filter, except for nmvoc. It should be consistent.

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/output/results/configs/configs.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/output/results/configs/configs.py
@@ -347,6 +347,7 @@ class LTPConfig(ResultConfig):
                     title="NMVOC From Flare",
                     unit=Unit.TONS,
                     query=EmissionQuery(
+                        installation_category="FIXED",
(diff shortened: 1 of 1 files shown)


TeeeJay (approved):
this was the only one that was left that was not consistent? or it was the one that was asked about in the other issue? are we overall consistent now?

frodehk:
From what I understand we are consistent now, based on earlier discussions with @halsyl.

halsyl:
What about storage/loading, @frodehk? Shouldn't they also be fixed? Ref. picture in https://github.com/equinor/ecalc-engine/issues/4643

frodehk:
You are right @halsyl. I have looked again at your table, and loading/storage are fixed:
https://docs.eerd.equinor.com/LTP/Centuries_LTP_file/

frodehk:
I have created a new PR to fix this. One thing, should we ensure that _all_ columns have installation filters? So, if the user forget to include installation category, no variables are reported?

https://github.com/equinor/ecalc/pull/91

PR #91, opened 2023-06-23
chore: add installation filter to remaining ltp-columns

frodehk:

Installation filter may not be relevant for all variables in the ltp-report, but to ensure consistency installation filter should be included. This will reduce confusion if e.g. the user has forgotten to include installation category (it is optional) in the yaml-file - and only the columns without installation filter are included in ltp-report.

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/output/results/configs/configs.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/output/results/configs/configs.py
@@ -441,6 +441,7 @@ class LTPConfig(ResultConfig):
                     title="NMVOC From Storage",
                     unit=Unit.TONS,
                     query=EmissionQuery(
+                        installation_category="FIXED",
                         consumer_categories=["STORAGE"],
                         emission_type="nmvoc",
                     ),
@@ -450,6 +451,7 @@ class LTPConfig(ResultConfig):
                     title="NMVOC From Loading",
                     unit=Unit.TONS,
                     query=EmissionQuery(
+                        installation_category="FIXED",
                         consumer_categories=["LOADING"],
                         emission_type="nmvoc",
                     ),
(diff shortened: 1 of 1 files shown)


frodehk:
@halsyl, currently, these columns do not have installation filter:

turbineFuelGasConsumption
turbineFuelGasCo2Mass
co2VentingMass
turbineFuelGasNoxMass
turbineFuelGasNmvocMass
turbineFuelGasCh4Mass
gasTurbineGeneratorConsumption
gasTurbineCompressorConsumption
offshoreWindConsumption
fromShoreConsumption
steamTurbineGeneratorConsumption

Should we include it for some of them, or is it ok as it is now?

halsyl:
Initially, I would say that we could put the "FIXED" filter on all of these, as they're currently not relevant for mobile units. But then I remembered that there's a rig that may get power from shore when drilling at Martin Linge, so maybe that column should be left out. Not sure how it's going to be modeled...

frodehk:
Ok @halsyl, then all columns are filtered on installation category, except for power from shore

src/ecalc/libraries/libecalc/common/libecalc/output/results/configs/configs.py
```diff
@@ -597,6 +609,7 @@ def aggregator(frequency: Frequency) -> Aggregator:
                     title="Total Power Consumed By Gas-Turbine Driven Compressors",
                     unit=Unit.GIGA_WATT_HOURS,
                     query=FuelConsumerPowerConsumptionQuery(
+                        # installation_category="FIXED",
```
TeeeJay, line 612: please add a comment why this is commented out

frodehk: Ah, forgot to remove - part of debugging in code development process. Thanks.

frodehk: It is fixed now

TeeeJay (approved):
very difficult to see in the diff what has been changed. approved if it is only to add check on installation category. other changes should be in a separate PR to make it clear that there are > 1 change, if any

PR #94, opened 2023-06-26
chore: count parantheses in list of tokens only among the elements th…

olelod:
…at are strings

From release notes of numpy 1.25:
```
== and != warnings have been finalized. The == and != operators on arrays now always:

    raise errors that occur during comparisons such as when the arrays have incompatible shapes (np.array([1, 2]) == np.array([1, 2, 3])).

    return an array of all True or all False when values are fundamentally not comparable (e.g. have different dtypes). An example is np.array(["a"]) == np.array([1]).

    This mimics the Python behavior of returning False and True when comparing incompatible types like "a" == 1 and "a" != 1. For a long time these gave DeprecationWarning or FutureWarning.

```

This will make the current use of the list.count() method to count the number of parentheses when elements in the list are numpy arrays. This pull request make sure that we count parentheses only among elements that can contain then, namely strings.

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/expression/expression_evaluator.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/expression/expression_evaluator.py
@@ -46,7 +46,7 @@ def eval_tokens(tokens: List[Token], array_length: int) -> NDArray[np.float64]:
     check_tokens(token_values)
 
     evaluated_values = np.nan_to_num(
-        x=eval_parenteses(
+        x=eval_parentheses(
             tokens=token_values,
         )  # type: ignore[arg-type]
     )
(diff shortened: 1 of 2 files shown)


jorgenengelsen:
Did we not just downgrade to numpy 1.24? Not sure how it affects this PR.. #93

TeeeJay:
> Did we not just downgrade to numpy 1.24? Not sure how it affects this PR.. #93

Good that you are paying attention ;)

In numpy v1.24.x it was a FutureWarning for this, so we should have fixed it earlier. In Komodo, we cannot go higher than v1.24.2 at the moment since another package relies on that (since there is a bug in v1.24.3 that will only be fixed in upcoming v1.24.4 or a v1.25.x)...

For Web we dont  have the same restriction...

src/ecalc/libraries/libecalc/common/libecalc/expression/expression_evaluator.py
```diff
@@ -100,11 +99,21 @@ def eval_parenteses(
                 ) from e
 
             tokens = tokens[:substart] + [tokens_evaluated] + tokens[subend + 1 :]
+            number_of_left_parentheses, number_of_right_parentheses = count_parentheses(tokens=tokens)
 
     return eval_logicals(tokens)
 
 
+def count_parentheses(tokens: List[Union[float, int, bool, NDArray[np.float64], str]]) -> Tuple[int, int]:
```
TeeeJay, line 107: nice to extract it into a function. Makes it much easier to test it in isolation as well. *hint* :)

TeeeJay (approved):
nice, looks good. see comment

PR #96, opened 2023-06-26
chore: update changelog for v8.2 with interstage pressure control fix

olelod:


TeeeJay (approved):
more important to add this to the migration guide? I guess it is getting a bit confusing for us and users wrt changelog vs migration guide. Should we merge them somehow? We do generate an automatic changelog as well, that can/should be used instead of this, as long as we manage to keep that changelog readable for users...

olelod:
> more important to add this to the migration guide? I guess it is getting a bit confusing for us and users wrt changelog vs migration guide. Should we merge them somehow? We do generate an automatic changelog as well, that can/should be used instead of this, as long as we manage to keep that changelog readable for users...

But nothing has really changed for the user, except for that the results are now correct... The yaml-file is not affected by this.

HeddaSvendsen:
I think it is good enough to have it in the changelog for now, the affected models have already been checked. We can discuss how to do it in the future. 

If it still only apply for models with multiple streams, I would include that in the text.

halsyl:
> I think it is good enough to have it in the changelog for now, the affected models have already been checked. We can discuss how to do it in the future.
> 
> If it still only apply for models with multiple streams, I would include that in the text.

Support @HeddaSvendsen's comments

PR #97, opened 2023-06-26
chore: improve documentation on defining compressor charts using CURV…

olelod:
…E and CURVES

Documentation is a little bit confusing. Trying to tidy up by making separate pages for CURVE and CURVES (used in single speed compressor charts and variable speed compressor charts, respectively)


HeddaSvendsen (approved):
Looks good :)

PR #101, opened 2023-06-27
chore: update dependencies to be aligned with external requirements

TeeeJay:
up and downgrades that is needed to stay compatible with external dependencies.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -320,20 +320,6 @@ test = ["pytest"]
 typing = ["mypy (>=0.990)"]
 
 [[package]]
-name = "commonmark"
-version = "0.9.1"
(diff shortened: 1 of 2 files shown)


jorgenengelsen (approved):
Looks good to me, but there was some breaking changes in pandas 2.0? Are the changes done backwardscompatible?

(tried running with fixed "oldest version possible"?)

TeeeJay:
> Looks good to me, but there was some breaking changes in pandas 2.0? Are the changes done backwardscompatible?
> 
> (tried running with fixed "oldest version possible"?)


I remember that as well. I was just going to push it and see if it broke. Cant remember which changes it was. May have been changes that was introduced earlier for future compatibility, but required in 2.x. I will try to find if the build doesnt break.

PR #104, opened 2023-06-29
fix: make sure that suction pressure is less than or equal to dischar…

olelod:
…ge pressure for compressor train

Suction pressure higher than discharge pressure gave negative enthalpy change over a compressor train and hence negative power. 

Validates that suction pressure is less than or equal to the discharge pressure for a compressor train. If the suction pressure is higher than the discharge pressure a failure status INVALID_SUCTION_PRESSURE_INPUT is given. In reality it could be either the suction pressure being too high or the discharge pressure being too low, but that will be up to the user to figure out...it can be hard to automatically detect which one it is.

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/base.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/base.py
@@ -188,28 +188,29 @@ class CompressorModel(BaseModel):
                     f" The affected time steps will not be calculated, and rate is set to zero."
                 )
                 rate = np.where(intermediate_pressure <= 0, 0, rate)
-            intermediate_pressure = np.where(
-                np.logical_and(np.min(rate, axis=0) <= 0, intermediate_pressure <= 0), 1, intermediate_pressure
-            )
+                intermediate_pressure = np.where(intermediate_pressure <= 0, 1, intermediate_pressure)
         if not np.all(suction_pressure > 0):
             logger.warning(
                 f"Inlet pressure needs to be a positive value. Given values: {suction_pressure.tolist()}."
                 f" The affected time steps will not be calculated, and rate is set to zero."
             )
             rate = np.where(suction_pressure <= 0, 0, rate)
-        suction_pressure = np.where(
-            np.logical_and(np.min(rate, axis=0) <= 0, suction_pressure <= 0), 1, suction_pressure
-        )
+            suction_pressure = np.where(suction_pressure <= 0, 1, suction_pressure)
         if not np.all(discharge_pressure > 0):
             logger.warning(
                 f"Outlet pressure needs to be a positive value. Given values: {discharge_pressure.tolist()}"
                 f" The affected time steps will not be calculated, and rate is set to zero."
             )
             rate = np.where(discharge_pressure <= 0, 0, rate)
-        discharge_pressure = np.where(
-            np.logical_and(np.min(rate, axis=0) <= 0, discharge_pressure <= 0), 1, discharge_pressure
-        )
-
+            discharge_pressure = np.where(discharge_pressure <= 0, 1, discharge_pressure)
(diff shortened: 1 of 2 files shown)


src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/base.py
```diff
@@ -209,7 +209,19 @@ def _validate_operational_conditions(
...
+                f" pressure: {suction_pressure.tolist()}. Given values for outlet pressure:"
+                f" {discharge_pressure.tolist()}. The affected time steps will not be calculated,"
+                f" and rate is set to zero."
+            )
+            rate = np.where(discharge_pressure < suction_pressure, 0, rate)
+        suction_pressure = np.where(
+            np.logical_and(np.min(rate, axis=0) <= 0, suction_pressure > discharge_pressure),
+            discharge_pressure,
+            suction_pressure,
+        )
```
TeeeJay, line 224: fairly complicated to read this numpy juggling. can it be extracted in a separate function and given some tests?

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/base.py
TeeeJay, line 1: nice catch :)

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/base.py
```diff
@@ -209,7 +209,19 @@ def _validate_operational_conditions(
         discharge_pressure = np.where(
             np.logical_and(np.min(rate, axis=0) <= 0, discharge_pressure <= 0), 1, discharge_pressure
         )
-
+        if not np.all(discharge_pressure >= suction_pressure):
```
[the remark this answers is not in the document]
olelod: Not straight forward to know if it is the suction or discharge pressure which is wrong. The validation will return INVALID_SUCTION_PRESSURE_INPUT.

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/base.py
olelod: Since we are now only validating input when rate != 0 the numpy juggling can be simplified a bit, it seems.

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/base.py
```diff
@@ -188,28 +188,30 @@ def _validate_operational_conditions(
...
+            discharge_pressure = np.where(discharge_pressure <= 0, 1, discharge_pressure)
+        if not np.all(discharge_pressure >= suction_pressure):
+            logger.warning(
+                f"Inlet pressure needs to be a less than or equal to outlet pressure. Given values for inlet"
+                f" pressure: {suction_pressure.tolist()}. Given values for outlet pressure:"
+                f" {discharge_pressure.tolist()}. The affected time steps will not be calculated,"
+                f" and rate is set to zero."
+            )
+            rate = np.where(discharge_pressure < suction_pressure, 0, rate)
+            suction_pressure = np.where(suction_pressure > discharge_pressure, discharge_pressure, suction_pressure)
```
TeeeJay, line 214: Why do we change suction pressure to discharge pressure here when it is not calculated (ie suction > discharge)? Will it not be confusing?

TeeeJay (approved):
Good now, just want to know/clarify why we change the pressure when returning to user (I guess?). Should possibly notify user about this change as well?

olelod:
> Good now, just want to know/clarify why we change the pressure when returning to user (I guess?). Should possibly notify user about this change as well?

Good catch. On second thought I guess we don't need to change them. I was coloured by the similar checks against negative pressure - there we have to change them due to initialisation of fluids in NeqSim.

PR #106, opened 2023-07-04
ci: Lock pydantic version in CI and update hooks

jorgenengelsen:

This pull request is needed because CI fails on main

- Lock version of pydantic used in mypy pre-commit hook
- Update pre-commit hooks
- Resolve new errors in new version of hooks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ default_language_version:
   node: 16.15.1
 repos:
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: 'v0.0.261'
+    rev: 'v0.0.276'
     hooks:
       - id: ruff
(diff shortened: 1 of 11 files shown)


.pre-commit-config.yaml
```diff
@@ -19,15 +19,15 @@ repos:
...
+    rev: v1.4.1
     hooks:
       - id: mypy
         files: ^(src/ecalc/libraries/)
         exclude: tests|conftest.py
         args: ['--config-file=./pyproject.toml']
         additional_dependencies:
           - types-PyYAML
-          - pydantic
+          - pydantic==1.10.9
```
TeeeJay, line 30: it is only possible to specify a fixed version and not a range?

what about the other deps, set version for those as well?

TeeeJay (approved):

PR #111, opened 2023-07-11
fix: yaml dumper, indentless parameter hard-coded

jorgenengelsen:

Found a dead-parameter, could not find it being used anywhere, so not sure if it was actually a bug. Still confusing with unused parameters

Use indentless parameter instead of hard-coding False


jorgenengelsen:
Wrong

PR #125, opened 2023-07-14
Fix: bug in fluid stream mixing

jorgenengelsen:

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Dict, List, Union
+from typing import Dict, List, Optional, Union
 
 import numpy as np
 from libecalc import dto
@@ -58,27 +58,39 @@ class FluidStream:
         fluid_model: dto.FluidModel,
         pressure_bara: float = UnitConstants.STANDARD_PRESSURE_BARA,
         temperature_kelvin: float = UnitConstants.STANDARD_TEMPERATURE_KELVIN,
+        existing_fluid: Optional[NeqsimFluid] = None,
     ):
         self.fluid_model = fluid_model
+        self.initial_temperature_kelvin = temperature_kelvin
+        self.initial_pressure_bare = pressure_bara
 
         if not temperature_kelvin > 0:
             raise ValueError("FluidStream temperature needs to be above 0.")
         if not pressure_bara > 0:
             raise ValueError("FluidStream pressure needs to be above 0.")
+        if existing_fluid is None:
+            _neqsim_fluid_stream = NeqsimFluid.create_thermo_system(
+                composition=self._map_fluid_composition_to_neqsim(fluid_composition=self.fluid_model.composition),
+                temperature_kelvin=temperature_kelvin,
+                pressure_bara=pressure_bara,
+                eos_model=_map_eos_model_to_neqsim[self.fluid_model.eos_model],
+            )
+        else:
+            _neqsim_fluid_stream = existing_fluid
 
-        self._neqsim_fluid_stream = NeqsimFluid.create_thermo_system(
-            composition=self._map_fluid_composition_to_neqsim(fluid_composition=self.fluid_model.composition),
-            temperature_kelvin=temperature_kelvin,
-            pressure_bara=pressure_bara,
-            eos_model=_map_eos_model_to_neqsim[self.fluid_model.eos_model],
-        )
+        self._pressure_bara = _neqsim_fluid_stream.pressure_bara
+        self._temperature_kelvin = _neqsim_fluid_stream.temperature_kelvin
+        self._kappa = _neqsim_fluid_stream.kappa
+        self._density = _neqsim_fluid_stream.density
+        self._z = _neqsim_fluid_stream.z
+        self._enthalpy_joule_per_kg = _neqsim_fluid_stream.enthalpy_joule_per_kg
 
         if (
             pressure_bara == UnitConstants.STANDARD_PRESSURE_BARA
             and temperature_kelvin == UnitConstants.STANDARD_TEMPERATURE_KELVIN
         ):
-            self.standard_conditions_density = self._neqsim_fluid_stream.density
-            self.molar_mass_kg_per_mol = self._neqsim_fluid_stream.molar_mass
+            self.standard_conditions_density = _neqsim_fluid_stream.density
+            self.molar_mass_kg_per_mol = _neqsim_fluid_stream.molar_mass
         else:
             _neqsim_fluid_at_standard_conditions = NeqsimFluid.create_thermo_system(
                 composition=self._map_fluid_composition_to_neqsim(fluid_composition=self.fluid_model.composition),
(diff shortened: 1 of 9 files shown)


src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
```diff
@@ -62,7 +62,6 @@ def __init__(
             )
 
             self.standard_conditions_density = _neqsim_fluid_at_standard_conditions.density
-            self.molar_mass_kg_per_mol = _neqsim_fluid_at_standard_conditions.molar_mass
```
TeeeJay, line 65: not needed?

jorgenengelsen: Could not find it to be used anywhere. So I think so.

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
```diff
@@ -218,12 +217,12 @@ def mix_in_stream(
 
         other_fluid = NeqsimFluid.create_thermo_system(
             composition=other_fluid_stream.fluid_model.composition,
-            temperature_kelvin=other_fluid_stream.temperature_kelvin,
-            pressure_bara=other_fluid_stream.pressure_bara,
+            temperature_kelvin=temperature_kelvin,
```
TeeeJay, line 220: so, now we are basically saying that the temp and pressure for both streams is the same, even though it has been specified differently for the 2nd stream...how can this be right? A comment in code why this is correct?

Got this verified by @Jfoul , if he knows?

TeeeJay: never mind, saw wrong :p

TeeeJay: but, should there be a check that the streams has same perssure and temperature? even if that is not used any more?

src/ecalc/libraries/libecalc/common/tests/core/models/compressor_modelling/test_fluid.py
```diff
@@ -132,3 +132,34 @@ def test_set_new_pressure_and_enthalpy_or_temperature(fluid_streams: List[List[F
...
+
+def test_fluid_mixing(dry_fluid, rich_fluid):
+    """Test mixing two fluids together, check that the order does not change"""
+
+    dry_fluid_stream = FluidStream(fluid_model=dry_fluid)
+    rich_fluid_stream = FluidStream(fluid_model=rich_fluid)
+
+    mix_rich_into_dry = dry_fluid_stream.mix_in_stream(
+        other_fluid_stream=rich_fluid_stream,
+        self_mass_rate=1,
```
TeeeJay, line 145: should use something else than 1 for mass rate for streams? too simple?

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
```diff
@@ -218,12 +217,12 @@ def mix_in_stream(
...
+            pressure_bara=pressure_bara,
             eos_model=other_fluid_stream.fluid_model.eos_model,
         )
 
-        new_fluid_composition, _neqsim_fluid_stream = new_fluid.mix_streams(
+        mixed_fluid_composition, mixed_neqsim_fluid_stream = new_fluid.mix_streams(
```
TeeeJay, line 225: good renaming :)

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
```diff
@@ -233,11 +232,9 @@ def mix_in_stream(
...
         )
 
-        mixed_fluid = FluidStream(
-            existing_fluid=_neqsim_fluid_stream,
-            fluid_model=self.fluid_model,
+        return FluidStream(
+            existing_fluid=mixed_neqsim_fluid_stream,
+            temperature_kelvin=temperature_kelvin,
+            pressure_bara=pressure_bara,
+            fluid_model=dto.FluidModel(composition=mixed_fluid_composition, eos_model=self.fluid_model.eos_model),
```
TeeeJay, line 239: should EOS model be a part of the fluid stream object? should we consider moving it out and only specify when we do something with the streams? in theory self and other can have different EOS models here...

TeeeJay:
We must emphasize in changelog (docs) that this change can significantly change some values in results, and perhaps also give a clue on "what might change"

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
jorgenengelsen: Good question..

@Jfoul and I discussed the possibility of deviating FluidModels, I think we concluded that it won't cause an issue? You only care about the composition of the two fluids. We only set a new fluid_model for the mixed fluid stream. As it calculates its properties at initialization

src/ecalc/libraries/libecalc/common/tests/core/models/compressor_modelling/test_fluid.py
jorgenengelsen: Mass rate is only used to define the mixing of the components. So it is basicly just a fraction. " 1 part of self mixed with 1 part of new fluid", I guess we could change to 2:1 ratio if you think it could catch some faulty logic

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
jorgenengelsen: NeqSim will complain about it. But we might not want to rely on NeqSims Error messages?

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
Jfoul: In theory we should not mix two fluids with different EOS's, but it wouldn't cause a direct error since you use the EOS from the original stream (?). Maybe we should include a check though to make sure that the users doesn't try mix two different EOSs (might result in a breaking change for some users though?).

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
Jfoul: For the mixing of the streams it is quite important that they have the same pressure (else in reality you would have flow with the difference in pressure). It is not true that they need to have the same temperature though (this will cause no issue), so the more accurately the combined temperature should be used (an estimation is a weighted average of temperature with mass rate of the streams - not entirely accurate though, but more so than having the same temperature)

TeeeJay: I would want us to do this check yes, but might be out of scope of this issue, so ok to leave for another PR. Perhaps there are other checks we should do as well?

With regards to the temperature, that is currently not being used. Also, the args are misleading wrt. that pressure and temp must be the same, which is currently not tested.

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
TeeeJay: Should leave for the "consistency and check/validation PR" as well then.

TeeeJay: I guess we would not want to e.g. set a EOS for ALL fluid models for a given eCalc model. We might ahve some with GERG and some with SRK within the same eCalc model..right?

src/ecalc/libraries/libecalc/common/tests/core/models/compressor_modelling/test_fluid.py
TeeeJay: ok, just checking. should in general have 1 check per test, so that should probably be covered in another test, or might already have been done...?

TeeeJay (approved):
good to merge I guess, after extracting 1-2 issues wrt checking/validation

TeeeJay:
> good to merge I guess, after extracting 1-2 issues wrt checking/validation

ah, and updating changelog with changes this results in! Important for users and for QA

TeeeJay (approved):
lgtm

PR #128, opened 2023-08-01
Chore/add fluid mix checks

jorgenengelsen:

This pull request is needed because of discussions in #125 

Checks that pressure and eos model is equal of models being mixed


Jfoul (approved):
EOS and pressure check looks good :)

PR #131, opened 2023-08-01
chore!: energy model type not allowed to change over time

frodehk:
BREAKING CHANGE: Not possible to change energy model `TYPE` over time within one consumer.

eCalc web is not yet ready to handle results with multiple energy model types, hence it should not be allowed to specify in the yaml model file.

- [x] Add yaml model file check for multiple energy model types within one consumer. Raise value error if found.
- [x] Modify some yaml model test files to remove multiple energy model types within one consumer.
- [x] Update snapshots
- [x] Add test

https://github.com/equinor/ecalc-engine/issues/4635


src/ecalc/libraries/libecalc/common/libecalc/input/mappers/create_references.py
```diff
@@ -56,12 +56,58 @@ def create_references(configuration: PyYamlYamlModel, resources: Resources) -> R
...
+
+    check_multiple_energy_models(consumers_installations)
+
     return References(
         models=models,
         fuel_types=fuel_types,
     )
 
 
+def check_multiple_energy_models(consumers_installations: list):
```
jorgenengelsen, line 82:
```suggestion
def check_multiple_energy_models(consumers_installations: List[..?]):
```
correct typing, "list" is only supported in 3.10+

src/ecalc/libraries/libecalc/common/libecalc/input/mappers/create_references.py
```diff
@@ -56,12 +56,58 @@ def create_references(configuration: PyYamlYamlModel, resources: Resources) -> R
...
+    """
+    Check for different energy model types within one consumer.
+
+    Args:
+        consumers_installations (list): List of consumers per installation
+
+    Returns:
+        Value error if different energy model types found within one consumer
```
jorgenengelsen, line 90: Is it more correct to say it "returns None"
```suggestion
     Raises:
        Value error if different energy model types found within one consumer
```

frodehk: I agree, returns None is more reasonable. Will fix.

frodehk:
> Looks good to me.
> 
> Some comments:
> 
> * Is it needed to document this change? At least in the changelog?
> * Breaking changes should be marked with an exclamation point, like fix!, or feat!

Agree, should documentation be put in a different PR?

jorgenengelsen (approved):

jorgenengelsen:
> > Looks good to me.
> > Some comments:
> > 
> > * Is it needed to document this change? At least in the changelog?
> > * Breaking changes should be marked with an exclamation point, like fix!, or feat!
> 
> Agree, should documentation be put in a different PR?

Both are fine by me, I have no preference

TeeeJay (approved):
nice. remember clear changelog, but should hopefully not cause problems, as it "per documentation" shouldnt have been allowed already :)

PR #134, opened 2023-08-02
refactor: only do tp-flash once

jorgenengelsen:
Wierd thing:When doing this all tests pass, but entire test suite takes 30s longer, 33% increase in time. 

From 1:37 to 2:07 on my MacBook

Any input on why this would slow down performance?


jorgenengelsen:
Causes problem with caching. Will not pursue.

PR #135, opened 2023-08-07
chore: fix typing of fluid composition

jorgenengelsen:

confloat is not a type

Use field with float constraints


TeeeJay (approved):
AFAIK this has been done many places. Thought it was a common approach in pydantic to do it like this, but not much of a shortcut... Is this related to lint-rules ?

PR #136, opened 2023-08-08
chore: add power adjustment constant also for compressor trains with…

olelod:
… interstage pressure


src/ecalc/libraries/libecalc/common/tests/integration/snapshots/test_all_energy_usage_models/test_all_results/all_energy_usage_models_v3.json
```diff
@@ -1022,9 +1022,9 @@
                     "typ": "STREAM_DAY",
                     "unit": "MW",
                     "values": [
-                        355.37085,
-                        346.048583,
-                        358.164004,
+                        356.423481,
```
olelod, line 1025: Change is > 1 MW because there is a 0.05 power loss factor on top

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/variable_speed_compressor_train_common_shaft_multiple_streams_and_pressures.py
```diff
@@ -769,15 +769,18 @@ def evaluate_rate_ps_pint_pd(
                 train_results.append(compressor_train_result)
 
         power_mw = np.array([time_step.power_megawatt for time_step in train_results])
+        power_mw_adjusted = np.where(
+            power_mw > 0, power_mw + self.data_transfer_object.energy_usage_adjustment_constant, power_mw
```
jorgenengelsen, line 773:
```suggestion
            power_mw > 0, power_mw + self.energy_usage_adjustment_constant, power_mw
```
In my opinion I am not fan of the DTO layer not being contained in the initialisation of objects.

Suggest creating 'self.data_transfer_object.energy_usage_adjustment_constant' in the init method?

src/ecalc/libraries/libecalc/common/tests/integration/snapshots/test_all_energy_usage_models/test_all_results/all_energy_usage_models_v3.json
jorgenengelsen: Was this pull request not about power adjustment constants? Or power loss factor? In my head thats two different things.

Guess power adjustment constant would here be 1MW?

jorgenengelsen (commented):
Looks ok. What about power usage adjustment factor? Is that not supported for compressors?

Would also like a test of this outside "all_energy_usage_models"

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/variable_speed_compressor_train_common_shaft_multiple_streams_and_pressures.py
TeeeJay: I agree, but this has been violated all over the place already ..

src/ecalc/libraries/libecalc/common/tests/integration/snapshots/test_all_energy_usage_models/test_all_results/all_energy_usage_models_v3.json
TeeeJay: yes, but there is an additional power loss factor (0.05) in addition to the power adjustment constant (1 mw), which means 1 / 0.95 = 1.05... I guess

src/ecalc/libraries/libecalc/common/tests/integration/snapshots/test_all_energy_usage_models/test_all_results/all_energy_usage_models_v3.json
```diff
@@ -1044,9 +1044,9 @@
                     "typ": "STREAM_DAY",
                     "unit": "MW",
                     "values": [
-                        644.62915,
```
TeeeJay, line 1047: here the values decrease...do you know why? was that expected?

TeeeJay: as far as I can see they have all 1 MW in PAC

TeeeJay (approved):

src/ecalc/libraries/libecalc/common/tests/integration/snapshots/test_all_energy_usage_models/test_all_results/all_energy_usage_models_v3.json
olelod: The power adjustment constant is applied to the compressor train result (or other energy function results). The power loss factor is applied to the compressor train result (or other energy function results) before making it a consumer function result. So if the MODEL has a POWER_ADJUSTMENT_CONSTANT and the CONSUMER have a POWERLOSSFACTOR, they will accumulate in the consumer results which is what we look at in the json file here.

src/ecalc/libraries/libecalc/common/tests/integration/snapshots/test_all_energy_usage_models/test_all_results/all_energy_usage_models_v3.json
olelod: This is the power capacity margin, which should decrease when the power increase.

PR #137, opened 2023-08-09
feat: add compressor input pressures to results

frodehk:

Add compressor inlet- and outlet pressures from input data to eCalc-result object, as _requested_ inlet- and outlet pressures. When implemented in web, the input pressures can be plotted together with calculated pressures. Make it easier to understand that in case of pressure control mechanisms are active, the requested (input) inlet and outlet pressures may differ from what is actually calculated by eCalc.

- [x] Extend eCalc-result object with compressor inlet- and outlet pressures from input data
- [x] Write the input pressures to core- and dto results objects
- [x] Write the input pressures to output json-file
- [x] Write the input pressures to output csv-file
- [x] Update tests
- [x] Update snapshots

https://github.com/equinor/ecalc-engine/issues/4431


src/ecalc/libraries/libecalc/common/libecalc/core/consumers/legacy_consumer/consumer_function/results.py
```diff
@@ -41,6 +41,8 @@ class ConsumerFunctionResultBase(BaseModel):
 
     # New! to support fuel to power rate...for e.g. compressors emulating turbine
     power: Optional[NDArray[np.float64]]
+    requested_inlet_pressure: Optional[NDArray[np.float64]]
```
TeeeJay, line 44: this should not be here, but on the compressorresult, at least for now, as it should only be for compressors...right?

src/ecalc/libraries/libecalc/common/libecalc/core/result/results.py
```diff
@@ -79,6 +81,9 @@ def component_type(self):
 
 
 class CompressorModelResult(ConsumerModelResultBase, CompressorTrainResult):
+    requested_inlet_pressure: Optional[List[Optional[float]]]
```
TeeeJay, line 84: should this be here? used?

TeeeJay (commented):
Looks good.

Verify that it makes sense to put the `requested_in/outlet_pressure` where it is put. It may be, but should in general be put at `compressor`-level and reused for all different implementations of compressor.

I didnt realize what this issue was initially I see, since it introduces to combine input and output data, which we in general should have a general solution for. It also "touches" some of the things discussed in another issue to introduce a more structured object for all the different states for pressures, but it will be easy to change that later, even though it will change the output structure to make it easier to work with.

@olelod can you give your opinion?

src/ecalc/libraries/libecalc/common/libecalc/core/result/results.py
```diff
@@ -45,6 +45,8 @@ class CompressorResult(GenericComponentResult):
     recirculation_loss: TimeSeriesRate
     rate_exceeds_maximum: TimeSeriesBoolean
     outlet_pressure_before_choking: TimeSeriesFloat
+    requested_inlet_pressure: Optional[TimeSeriesFloat]
```
olelod, line 48: Do they need to be Optional? I think suction_pressure and discharge_pressure are required in a compressor energy_usage_model.

src/ecalc/libraries/libecalc/common/libecalc/core/result/results.py
frodehk: It is removed now

frodehk:
> Looks good.
> 
> Verify that it makes sense to put the `requested_in/outlet_pressure` where it is put. It may be, but should in general be put at `compressor`-level and reused for all different implementations of compressor.
> 
> I didnt realize what this issue was initially I see, since it introduces to combine input and output data, which we in general should have a general solution for. It also "touches" some of the things discussed in another issue to introduce a more structured object for all the different states for pressures, but it will be easy to change that later, even though it will change the output structure to make it easier to work with.
> 
> @olelod can you give your opinion?

@TeeeJay: ended up with a bit back en forth here. To sum up:
The requested inlet- and outlet pressures are now moved to `CompressorTrainResult`, which means it will only be in the models-object under consumer_results (and not in component_result). It will not dump the pressures to the csv (only json). 

If the changes have become difficult to review I can start with a clean PR, and have more structure if you want?

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/legacy_consumer/consumer_function/results.py
frodehk: Fixed

src/ecalc/cli/tests/snapshots/test_app/test_json_true/test_v3.json
```diff
@@ -787,6 +787,34 @@
         2000000.0,
         3000000.0
       ],
+      "requested_inlet_pressure": [
```
TeeeJay, line 790: Usually we specify in and outlet pressure, but with e.g. ASV it might be different from input. So, why are there so many empty/null values here? It's probably something I dont understand/see ... :)

src/ecalc/libraries/libecalc/common/libecalc/core/models/results/compressor.py
```diff
@@ -127,6 +127,8 @@ class CompressorTrainResult(EnergyFunctionResult):
     stage_results: List[CompressorStageResult]
     failure_status: List[Optional[CompressorTrainCommonShaftFailureStatus]]
     turbine_result: Optional[TurbineResult] = None
+    requested_inlet_pressure: List[Optional[float]]
```
TeeeJay, line 130: Isnt this always set? but as math.nan as a fallback. Can the values ever be `None`?

src/ecalc/libraries/libecalc/common/tests/core/consumers/system/conftest.py
```diff
@@ -203,6 +203,8 @@ def compressor_model_result() -> CompressorTrainResult:
         ],
         rate_sm3_day=[np.nan, np.nan, np.nan],
         failure_status=[None, None, None],
+        requested_inlet_pressure=[np.nan, np.nan, np.nan],
```
TeeeJay, line 206: Why set to NaN? Wouldnt it be better with real numbers?

frodehk: As it is only a test, not using the requested in- and outlet pressures to anything - I figured it was ok just to add NaN. No problem to add real numbers though.

TeeeJay:
> > Looks good.
> > Verify that it makes sense to put the `requested_in/outlet_pressure` where it is put. It may be, but should in general be put at `compressor`-level and reused for all different implementations of compressor.
> > I didnt realize what this issue was initially I see, since it introduces to combine input and output data, which we in general should have a general solution for. It also "touches" some of the things discussed in another issue to introduce a more structured object for all the different states for pressures, but it will be easy to change that later, even though it will change the output structure to make it easier to work with.
> > @olelod can you give your opinion?
> 
> @TeeeJay: ended up with a bit back and forth here. To sum up: The requested inlet- and outlet pressures are now moved to `CompressorTrainResult`, which means it will only be in the models-object under consumer_results (and not in component_result). It will not dump the pressures to the csv (only json).
> 
> If the changes have become difficult to review I can start with a clean PR, and have more structure if you want?

Looking good, but the results needs to be added to `dto.CompressorResult` afaik, and mapped to it, which I cannot see is happening ...

src/ecalc/libraries/libecalc/common/libecalc/core/models/results/compressor.py
frodehk: I need to double check this

frodehk: I though so as well. However, when running the TestShowResultsCommand (using show), cached results from an eCalc run are loaded. In this process, empty/null values are converted to None. So, I keep it for now.

TeeeJay: ah, good point

src/ecalc/cli/tests/snapshots/test_app/test_json_true/test_v3.json
TeeeJay: ok

frodehk:
> > > Looks good.
> > > Verify that it makes sense to put the `requested_in/outlet_pressure` where it is put. It may be, but should in general be put at `compressor`-level and reused for all different implementations of compressor.
> > > I didnt realize what this issue was initially I see, since it introduces to combine input and output data, which we in general should have a general solution for. It also "touches" some of the things discussed in another issue to introduce a more structured object for all the different states for pressures, but it will be easy to change that later, even though it will change the output structure to make it easier to work with.
> > > @olelod can you give your opinion?
> > 
> > 
> > @TeeeJay: ended up with a bit back and forth here. To sum up: The requested inlet- and outlet pressures are now moved to `CompressorTrainResult`, which means it will only be in the models-object under consumer_results (and not in component_result). It will not dump the pressures to the csv (only json).
> > If the changes have become difficult to review I can start with a clean PR, and have more structure if you want?
> 
> Looking good, but the results needs to be added to `dto.CompressorResult` afaik, and mapped to it, which I cannot see is happening ...

Are you sure we need it in dto.CompressorResult? You mean here: https://github.com/equinor/ecalc/blob/c8a486114ec713358798a5dba2a5500dfbbef21d/src/ecalc/libraries/libecalc/common/libecalc/dto/result/results.py#L105C11-L105C11?

The pressures are now in the eCalc result object, under `models` (CompressorModelResult, see figure), but not under component_result (`CompressorRresult`). Originally I put it both places - but is there a need for that? I see that `outlet_pressure_before_choking` is both places.

consumer_results navn

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/sampled/compressor_model_sampled.py
```diff
@@ -198,6 +199,13 @@ def evaluate_rate_ps_pd(
         ps_to_evaluate = suction_pressure[indices_to_evaluate] if suction_pressure is not None else []
         pd_to_evaluate = discharge_pressure[indices_to_evaluate] if discharge_pressure is not None else []
 
+        requested_inlet_pressure = (
+            list(ps_to_evaluate) if list(ps_to_evaluate) else [math.nan] * len(list(indices_to_evaluate))
```
TeeeJay, line 203: not sure if I completely understand this check, supposed to be: 
```suggestion
            list(ps_to_evaluate) if ps_to_evaluate else [math.nan] * len(list(indices_to_evaluate))
```

?

frodehk: ps_to_evaluate can return an empty list, and the test is checking for that - and returns tha fallback full list of NaN instead of an empty list.

frodehk:
@TeeeJay, @jsolaas: The pressures are now added to dto.CompressorResults, so I think they should be ready for web now.

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/sampled/compressor_model_sampled.py
TeeeJay: an empty list, and None both evaluates to False (I guess not None can occur though), but it is better to just test more explicitly to understand what one expects? e.g. `if  ps_to_evaluate` or `if len(ps_to_evaluate) == 0` ...but in that case it is already a list so no need to add `list([])` ?

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/legacy_consumer/component.py
```diff
@@ -265,10 +265,27 @@ def evaluate(
                         new_time_vector=variables_map.time_vector,
                     )
                 )
+
```
TeeeJay, line 268: looks ok, but we are a bit inconsistent in the way we create the TimeSeries here, should be done in the same way unless there is a reason not to (see above)

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/sampled/compressor_model_sampled.py
```diff
@@ -198,6 +199,13 @@ def evaluate_rate_ps_pd(
...
+        )
+        requested_outlet_pressure = (
+            list(pd_to_evaluate) if list(pd_to_evaluate) else [math.nan] * len(list(indices_to_evaluate))
```
TeeeJay, line 206: Are we always consistent in when we add a vector with n `NaN`s and when we just set `null/None`? It is easier to deal with n values, but we might want to reconsider this as it causes a lot of excess "no data" ... just sth to think about

TeeeJay (commented):
looks ok, but maybe we are adding more than needed e.g. in core. check with jostein :)

PR #145, opened 2023-08-23
feat: add conditions per consumer in system v2

jsolaas:


jsolaas:
condition not part of MVP, not sure how this should be implemented

PR #148, opened 2023-08-29
fix: system v2 evaluation

jsolaas:
This commit fixes the system v2. The previous implementation did not work correctly when the crossover changed between operational settings. The test has been updated to reflect this bug.

The system v2 component has also been updated to create consumer results for each consumer. This result should reflect the operational settings used, i.e. system v2 should pick the correct results per timestep for each consumer based on the operational conditions used.


src/ecalc/libraries/libecalc/common/libecalc/core/consumers/compressor/component.py
```diff
@@ -90,20 +90,25 @@ def evaluate(
...
                 unit=aggregated_result.power_unit,
                 regularity=regularity,
-            ),
-            energy_usage=TimeSeriesRate(
-                values=aggregated_result.energy_usage,
-                timesteps=evaluated_timesteps,
-                unit=aggregated_result.energy_usage_unit,
-                regularity=regularity,
-            ),
+            ).fill_nan(0.0),
```
jsolaas, line 110: Fill nan because this is done in legacy. Should we handle this some other way?

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/consumer_system.py
```diff
@@ -138,87 +139,107 @@ def evaluate(
...
-
-        models = list(itertools.chain(*[consumer_result.models for consumer_result in consumer_system_result]))
-
         return EcalcModelResult(
-            component_result=consumer_result,
-            sub_components=[],  # Keeping this backward compatible with V1 for now.
-            models=models,
+            component_result=consumer_system_result,
+            sub_components=consumer_results,
+            models=[],
```
jsolaas, line 209: We should figure out how to solve models for system v2. Currently the core results does not have any parent attribute, so we lose track of which model result belongs to which consumer. 

We also need to merge/collect model results based on operational settings used if we are going to provide model results.

src/ecalc/libraries/libecalc/fixtures/libecalc/fixtures/cases/consumer_system_v2/consumer_system_v2_dto.py
```diff
@@ -173,23 +173,23 @@ def consumer_system_v2_dto() -> DTOCase:
...
-                ),
+                    crossover=[0, 0, 0],
+                ),  # Invalid operational setting, should not be valid for any timesteps
                 dto.components.CompressorSystemOperationalSetting(
-                    rates=[Expression.setup_from_expression(x) for x in [500000, 2500000, 3000000]],
+                    rates=[Expression.setup_from_expression(x) for x in [1000000, 5000000, 5000000]],
                     inlet_pressures=[Expression.setup_from_expression("50")] * 3,
                     outlet_pressures=[Expression.setup_from_expression("125")] * 3,
-                    crossover=[0, 1, 1],
+                    crossover=[0, 1, 1],  # Crossover makes this valid. 1 mill from consumer 2 and 3 sent to 1.
```
jsolaas, line 185: Before this PR this test would have failed because the system did not evaluate with this crossover, instead the crossover in the first operational setting was always used.

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/consumer_system.py
```diff
@@ -138,134 +138,109 @@ def evaluate(
         operational_settings_used = TimeSeriesInt(
             timesteps=variables_map.time_vector, values=[0] * len(variables_map.time_vector), unit=Unit.NONE
         )
-        operational_settings_results: Dict[int, List[ConsumerModelResult]] = defaultdict(list)
+        operational_settings_results: Dict[int, Dict[str, EcalcModelResult]] = defaultdict(
```
jsolaas, line 141: We still need to expand on this, as this will fail when temporal operational settings are used.

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/compressor/component.py
TeeeJay: Think we should stick to v1 for comparison reasone initially, yes. Then make an overall decision across "everything" on how we should handle NaN, instead of deciding on a per case basis...

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/compressor/component.py
```diff
@@ -90,20 +90,25 @@ def evaluate(
...
 
+        energy_usage = TimeSeriesRate(
+            values=aggregated_result.energy_usage,
+            timesteps=evaluated_timesteps,
+            unit=aggregated_result.energy_usage_unit,
+            regularity=regularity,
+        )
+
+        if energy_usage.unit == Unit.STANDARD_CUBIC_METER_PER_DAY:
+            energy_usage = energy_usage.to_calendar_day()  # legacy consumer converts to calendar day if fuel consumer
```
TeeeJay, line 101: can you elaborate a bit more on the comment, I dont quite understand it as it is

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/consumer_system.py
TeeeJay: leave as is now, will look at it :)

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/consumer_system.py
```diff
@@ -138,134 +138,115 @@ def evaluate(
...
+                )
+
+                for consumer, adjusted_operational_setting in zip(self._consumers, adjusted_operational_settings):
+                    consumer_result = consumer.evaluate(adjusted_operational_setting)
+                    operational_settings_results[operational_setting_index][consumer.id] = consumer_result
+
+                consumer_results = operational_settings_results[operational_setting_index].values()
+
+                # Check if consumers are valid for this operational setting, should be valid for all consumers
+                is_all_consumers_valid = reduce(
```
TeeeJay, line 163: is it the consumer that is valid or the results? rename to `all_consumer_results_valid`? prefix `is_` is reserved for functions?

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/consumer_system.py
TeeeJay: you will follow up this in separate PR?

src/ecalc/libraries/libecalc/common/libecalc/core/result/results.py
```diff
@@ -29,6 +32,56 @@ class CommonResultBase(EcalcResultBaseModel):
 class GenericComponentResult(CommonResultBase):
     id: str
 
+    @property
```
TeeeJay, line 35: Rename to `attributes` or sth? Make a separate type/class for this and make it possible to have "somehow" "(static) typing", e.g. check that the attribute actually exists?

src/ecalc/libraries/libecalc/common/libecalc/core/result/results.py
```diff
@@ -29,6 +32,56 @@ class CommonResultBase(EcalcResultBaseModel):
...
+            elif isinstance(value, TimeSeriesRate):
+                columns[key] = value.values
+                if value.regularity is not None:
+                    columns[f"{key}_regularity"] = value.regularity
+            elif isinstance(value, TimeSeries):
+                columns[key] = value.values
+        return columns
+
+    @property
+    def dataframe(self):
```
TeeeJay, line 53:
```suggestion
    def dataframe(self) -> pd.DataFrame:
```

TeeeJay (approved):

src/ecalc/libraries/libecalc/common/libecalc/core/result/results.py
jsolaas: Isn't attributes more generic than columns? We want to indicate that this will give the attributes that are sequences. 

Are you thinking of a separate class that has the columns property? Then multiple inheritance or a subclass of GenericComponentResult? We have something similar to what I'm doing here in `dto.result.tabular_time_series.py`. TabularTimeSeries has a method to_dataframe method. What do you think of that solution?

jsolaas: I think we should be able to have generic functions for merge and get_subset also, so a separate class for tabular_time_series with that logic makes sense to me. We are adding DataFrame functionality to the results, with a bit more control and some additional data. If I remember correctly that was also the case/reasoning when adding TabularTimeSeries to dto.result

jsolaas: Ja, jeg skjønner. Det jeg tenkte å eksponere fra denne klassen var get_subset og merge, resten var hjelpemetoder. Så vi kan jo diskutere om det burde finnes en separat klasse for de to metodene eller om vi skal ha de på dto. Også kan vi se om vi skal flytte noe fra hjelpemetodene i egne klasser så klart.

jsolaas: https://equinor-ecalc.atlassian.net/browse/ECALC-136

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/consumer_system.py
TeeeJay: I have briefly thought about this problem, so yes, lets figure out

PR #152, opened 2023-09-05
feat: add compressor inlet- and outlet pressures to models/train level

frodehk:
Add compressor inlet- and outlet pressures from input data to models/train level (as requested inlet- and outlet pressures). It is initially implemented at component level, this is now removed - as it is most relevant for the users at train-level. When implemented in web, the input pressures can be plotted together with calculated pressures. Make it easier to understand that in case of pressure control mechanisms are active, the requested (input) inlet and outlet pressures may differ from what is actually calculated by eCalc.

- [x] graph_result.py: update and add new function to add pressures at compressor train-level
- [ ] graph_result.py: remove pressures at compressor component level
- [x] Write the input pressures to dto.result.results.CompressorModelResult object
- [x] Update snapshots

https://github.com/equinor/ecalc-engine/issues/4431


src/ecalc/cli/tests/snapshots/test_app/test_json_true/test_v3.json
```diff
@@ -1066,52 +1128,6 @@
       },
       "name": "Gas injection compressor",
       "parent": "e9057e3053eac8e2dca7909380050de9",
-      "power": {
```
TeeeJay, line 1069: intended that this is removed?

src/ecalc/cli/tests/snapshots/test_app/test_json_true/test_v3.json
```diff
@@ -787,6 +787,68 @@
...
+          "2024-12-01 00:00:00",
+          "2026-01-01 00:00:00",
+          "2027-01-01 00:00:00",
+          "2028-01-01 00:00:00",
+          "2029-01-01 00:00:00",
+          "2030-01-01 00:00:00",
+          "2031-01-01 00:00:00"
+        ],
+        "unit": "bara",
+        "values": [
```
TeeeJay, line 837: compressor tabular  I assume this was?

src/ecalc/cli/tests/snapshots/test_app/test_json_true_detailed_output/test_full_json_v3.json
```diff
@@ -1066,52 +1128,6 @@
       },
       "name": "Gas injection compressor",
       "parent": "e9057e3053eac8e2dca7909380050de9",
-      "power": {
```
TeeeJay, line 1069: also intended?

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
```diff
@@ -204,31 +204,50 @@ def _parse_emissions(emissions: Dict[str, EmissionResult]) -> Dict[str, libecalc
...
         :param energy_usage_model: Temporal energy model
         :param pressure_type: Compressor pressure type, inlet- or outlet
+        :param name: name of compressor
         :return: Temporal model with pressures as expressions
         """
         evaluated_temporal_energy_usage_models = {}
         for period, model in TemporalModel(energy_usage_model).items():
-            pressures = model.suction_pressure
+            if isinstance(model, CompressorSystemConsumerFunction):
+                for compressor, operational_setting in zip(model.compressors, model.operational_settings):
```
TeeeJay, line 223: does this handle if the operational settings is/has temporal models as well? Is that possible?

TeeeJay:
We should have consumer systems that return sth else than `null` for requested pressures. Now all snapshot tests only return `null` meaning that we dont have any snapshot tests that has a train that returns data? Sounds weird. If that is the case, we should have at least one happy path case for that, ideally for each "possibility"/edge case. For the cases where it is not relevant, which is probably coverd in snapshot tests, it looks ok (`null` as in no data).

src/ecalc/cli/tests/snapshots/test_app/test_json_true_detailed_output/test_full_json_v3.json
TeeeJay: if intended, and as a result of a bug fix, please make that a separate PR :)

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
frodehk: That was the intention at least. Thought I have checked that it actually does that - but need to double check:-)

frodehk:
There is also a mismatch with the train/model timevector and the individual variables (e.g. rates, pressure etc) per now, i.e. in the way it is implemented in this PR. Think there is a need to slice the variables map as input to the temporal expression, to match correct time intervals.

src/ecalc/cli/tests/snapshots/test_app/test_json_true/test_v3.json
frodehk: No, this is fixed

src/ecalc/cli/tests/snapshots/test_app/test_json_true/test_v3.json
frodehk: Yes, `COMPRESSOR_TABULAR`

src/ecalc/cli/tests/snapshots/test_app/test_json_true_detailed_output/test_full_json_v3.json
frodehk: Not intended, fixed

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
frodehk: And more detailed:
image

frodehk: So as far as I can see, the function handles temporal models both for compressor systems, and for compressors

frodehk:
> We should have consumer systems that return sth else than `null` for requested pressures. Now all snapshot tests only return `null` meaning that we dont have any snapshot tests that has a train that returns data? Sounds weird. If that is the case, we should have at least one happy path case for that, ideally for each "possibility"/edge case. For the cases where it is not relevant, which is probably coverd in snapshot tests, it looks ok (`null` as in no data).

Seems like the csv-tests are the only ones using the advanced yaml (with more advanced settings for compressors and compressor systems). I think we should extend the json-tests with one using the advanced yaml-file.

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
TeeeJay: nice. My mistake, thought we could have temporal models on operational settings only. That is not the case. I trust you :) ;)

TeeeJay:
> > We should have consumer systems that return sth else than `null` for requested pressures. Now all snapshot tests only return `null` meaning that we dont have any snapshot tests that has a train that returns data? Sounds weird. If that is the case, we should have at least one happy path case for that, ideally for each "possibility"/edge case. For the cases where it is not relevant, which is probably coverd in snapshot tests, it looks ok (`null` as in no data).
> 
> Seems like the csv-tests are the only ones using the advanced yaml (with more advanced settings for compressors and compressor systems) - but data for models/trains are only dumped to json. I think we should extend the json-tests with one using the advanced yaml-file.

yes, or add a relevant test (can be really simple) to the existing json-export-tests

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
```diff
@@ -246,108 +266,111 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
...
-                requested_inlet_pressure = TimeSeriesFloat(
-                    timesteps=self.timesteps,
-                    values=TemporalExpression.evaluate(inlet_pressure_eval, self.variables_map),
-                    unit=Unit.BARA,
-                )
+                    requested_inlet_pressure = TimeSeriesFloat(
+                        timesteps=self.timesteps,
+                        values=TemporalExpression.evaluate(inlet_pressure_eval, self.variables_map),
+                        unit=Unit.BARA,
+                    )[start_index : end_index + 1]
```
frodehk, line 291: @TeeeJay: Added this to get data for the correct time interval for the model (and not data for the the whole global time vector). Makes sense?

src/ecalc/cli/tests/test_app.py
```diff
@@ -301,6 +301,28 @@ def test_json_true_detailed_output(self, simple_yaml_path, tmp_path, snapshot):
...
+                detailed_output=True,
+            ),
+            catch_exceptions=False,
+        )
+
+        v3_json_actual_path = tmp_path / f"{run_name_prefix}_v3.json"
+        assert v3_json_actual_path.is_file()
+        with open(v3_json_actual_path) as json_file:
+            json_data = json.loads(json_file.read())
+        snapshot.assert_match(
```
frodehk, line 322: @TeeeJay, not sure why this fails in the pipeline. It succeeds when running locally.

TeeeJay: run `poetry install` locally, in case you have outdated dependencies installed locally

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
TeeeJay: I guess this is more or less correct, but we should be able to get the subset of the TimeSeries directly using the `TimeSeries.subset(start, end)` or `TimeSeries.subset(period)` instead of trixing and mixing with indices all over the place, if possible.

src/ecalc/cli/tests/test_app.py
frodehk: Ran `poetry install `locally, didn´t help

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
frodehk: TimeSeries(period) is added.

src/ecalc/cli/tests/test_app.py
TeeeJay: still fails only in pipeline? locally ok? do you run pytest in docker or not?

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -192,6 +192,10 @@ def forward_fill(self) -> Self:
     def fill_nan(self, fill_value: float) -> Self:
         return self.copy(update={"values": pd.Series(self.values).fillna(fill_value).tolist()})
 
+    def subset(self, period: Period) -> TimeSeries:
```
TeeeJay, line 195: the `for_period` method does this, use that! :)

src/ecalc/cli/tests/test_app.py
TeeeJay: Seems to be a classic difference between python (or numpy) versions. Which python version do you use when testing? the difference is in xth decimal AFAI can see, so no big problem. Ideally we should round off that output to fewer decimals, if possible, but I guess it is a dump, and hence not rounded off in serialization? Maybe we can do that after all? ...

TeeeJay: We use 3.8/3.9 in CICD, in general etc due to internal restrictions. To make sure there are no differences, I recommend to downgrade/use 3.9.17 (or newer) locally.

TeeeJay: this is why one should use docker, in order to make sure that the environment is correctly reproduced, and not dependent on  local configuration

src/ecalc/libraries/libecalc/common/libecalc/dto/result/results.py
```diff
@@ -138,6 +136,8 @@ class PumpModelResult(ConsumerModelResultBase):
 
 class CompressorModelResult(ConsumerModelResultBase, CompressorTrainResult):
     componentType: Literal[ComponentType.COMPRESSOR]
+    requested_inlet_pressure: TimeSeriesFloat
```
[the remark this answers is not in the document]
frodehk: We can have COMPRESSOR_TABULAR (no compressor train), where suction- and discharge pressures are specified:
image

src/ecalc/cli/tests/test_app.py
frodehk: Tested with docker, python 3.8.14, 3.8.16, 3.8.18, 3.9.17 - same. Will try to modify the test, do it without snapshot.

src/ecalc/libraries/libecalc/common/libecalc/dto/result/results.py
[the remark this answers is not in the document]
TeeeJay: Yes, but then input and output (result) will be the same every time right? It is just with compressor trains it is useful?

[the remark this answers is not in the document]
TeeeJay: I agree that this should be everywhere - but we are already leaving it our many places, and the easiest definition on where we and users can expect this to be available is for trains only?

src/ecalc/cli/tests/test_app.py
TeeeJay: Same results in all versions? I can give it a go on my machine...possibly numpy version differs then, but that should be "locked"...

src/ecalc/libraries/libecalc/common/libecalc/dto/result/results.py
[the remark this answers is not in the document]
frodehk: Yes I assume that the input- and output is the same in that case. However, isn´t it better to include everywhere - as far as I understand some of the v2-ideas, it is to define it for all - even at the consumer level. Comment @jsolaas?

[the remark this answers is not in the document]
frodehk: Yes, I think this was the intention. That is why we focused only on the model-level (not consumer). It was quite easy to implement for CompressorModelResult (and not only CompressorTrainResult - I think you are right that the original focus was train-only)

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
```diff
@@ -248,108 +266,110 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
...
-                    energy_usage_model=component.energy_usage_model,
-                    pressure_type=CompressorPressureType.INLET_PRESSURE,
-                )
+                for model in consumer_result.models:
+                    inlet_pressure_eval = GraphResult.extract_input_compressor_pressures_from_models(
+                        energy_usage_model=component.energy_usage_model,
+                        pressure_type=CompressorPressureType.INLET_PRESSURE,
+                        name=model.name,
+                    )
+                    outlet_pressure_eval = GraphResult.extract_input_compressor_pressures_from_models(
```
TeeeJay, line 278: extract input to get output? ...ctrl-c-ctrl-v error...

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
```diff
@@ -405,7 +425,7 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
             id=asset.id,
             name=asset_node_info.name,
             component_level=asset_node_info.component_level,
-            componentType=asset_node_info.component_level,
+            componentType=asset_node_info.component_type,
```
TeeeJay, line 428: bugfix?

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
```diff
@@ -205,32 +205,50 @@ def _parse_emissions(emissions: Dict[str, EmissionResult]) -> Dict[str, libecalc
         }
 
     @staticmethod
-    def extract_requested_pressures_from_model(
-        energy_usage_model: Dict[datetime, EnergyUsageModel],
+    def extract_input_compressor_pressures_from_models(
```
TeeeJay, line 208:
```suggestion
    def get_requested_compressor_pressures(
```

I dont like the naming, easily confused with "inlet_pressure" imo. Changing it to the name above makes it more agnostic of "how it is done", and more focused on "what is it that we get"...what do you think?

src/ecalc/cli/tests/test_app.py
```diff
@@ -301,6 +301,46 @@ def test_json_true_detailed_output(self, simple_yaml_path, tmp_path, snapshot):
...
+
+        compressor_tabular = json_data["models"][0]
+        compressor_train = json_data["models"][1]
+
+        requested_inlet_pressure_tabular = compressor_tabular["requested_inlet_pressure"]["values"]
+        requested_outlet_pressure_tabular = compressor_tabular["requested_outlet_pressure"]["values"]
+        requested_inlet_pressure_train = compressor_train["requested_inlet_pressure"]["values"]
+        requested_outlet_pressure_train = compressor_train["requested_outlet_pressure"]["values"]
+
+        assert requested_inlet_pressure_tabular == [20] * len(compressor_tabular["timesteps"])
```
TeeeJay, line 339: do we have the calculated pressure here as well? does it differ from the requested?

TeeeJay (approved):
all checks are ok, not sure what you meant by erroneous snapshot? did you perhaps just change to the values we got in CICD , but that it is still failing localyl for you?

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
frodehk: I agree

src/ecalc/cli/tests/test_app.py
frodehk: I cannot find it in the json data object, read from disk. Seems like inlet_stream and outlet_stream (where calculated pressures are) are not written to json. Even though it is part of the dto that goes in to the write_json function:
image

frodehk: @TeeeJay any ideas why inlet_stream is not written to json, is it because of the CompressorStreamCondition class - mapping issue?

src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
frodehk: Yes, removed it. Will include in new PR.

frodehk:
> all checks are ok, not sure what you meant by erroneous snapshot? did you perhaps just change to the values we got in CICD , but that it is still failing localyl for you?

All tests are ok now, as the snapshot assert is removed, i.e. not using snapshot. Originally it was using snapshot. The snapshot is generated by copying the output json from the test, to the snapshot directory. Then running the test with --snapshot-update. Then, locally all is fine but it fails in the pipeline (in the 10th decimal I think). But as you see, snapshot is not used now.

src/ecalc/cli/tests/test_app.py
frodehk: Found the calculated pressures in stage results.

PR #154, opened 2023-09-07
Handle dates corectly in yaml

TeeeJay:
See the commits separately.

1:
[fix: handle dates in yaml correctly](https://github.com/equinor/ecalc/pull/154/commits/ea31a24f31fecd35108c62eeb522904e6c3d31dd)

if we dont explicitly handle dates with our own date type, then
PyYAML will change to datetime.date for what we want to have
datetime.datetime. To make sure that all are converted to
datetime.datetime, we set all dates in yaml as our date type.

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/yaml_temporal_model.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/yaml_temporal_model.py
@@ -1,11 +1,6 @@
 from typing import Dict, TypeVar, Union
 
-from pydantic import ConstrainedStr
-
-
-class DatetimeString(ConstrainedStr):
-    regex = "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$"
-
+from libecalc.input.yaml_types.yaml_variable import YamlDefaultDatetime
 
 TModel = TypeVar("TModel")
-YamlTemporalModel = Union[TModel, Dict[DatetimeString, TModel]]
+YamlTemporalModel = Union[TModel, Dict[YamlDefaultDatetime, TModel]]
--- a/src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/yaml_variable.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/yaml_variable.py
@@ -9,6 +9,10 @@ from pydantic.datetime_parse import parse_date, parse_datetime
 
 
 class YamlSingleVariable(YamlBase):
+    """
+    A variable in YAML that can be set to a valid eCalc Expression
+    """
+
     value: Expression
 
     def to_dto(self):
@@ -16,12 +20,25 @@ class YamlSingleVariable(YamlBase):
 
 
 class YamlDefaultDatetime(datetime):
+    """
+    PyYAML is smart and detects datetime.date and datetime.datetime differently in YAML, and parses usually
+    dates to datetime.date. However, in eCalc we required datetime.datetime, and there is a subtle difference
+    in behaviour between those too. Therefore we need to cast to datetime.datetime as early as possible to make
+    sure eCalc behaves correctly.
+    """
+
     @classmethod
     def __get_validators__(cls):
         yield cls.date_to_datetime
 
     @classmethod
     def date_to_datetime(cls, value) -> datetime:
+        """
+        Handles both datetimes and dates supported formats, and converts them to
+        datetime.datetime internally to avoid problems with mixing datetime.date and datetime.datetime
+        :param value: string with a PyYAML supported date or datetime format
+        :return: the corresponding datetime.datetime. datetime.date will have HH:MM:SS set to 00:00:00
+        """
         try:
             date = parse_date(value)
             return convert_date_to_datetime(date)
--- a/src/ecalc/libraries/libecalc/common/tests/input/validation/snapshots/test_validation_json_schemas/test_json_schema_changed/schemas.json
+++ b/src/ecalc/libraries/libecalc/common/tests/input/validation/snapshots/test_validation_json_schemas/test_json_schema_changed/schemas.json
@@ -55,11 +55,6 @@
                                     "additionalProperties": {
                                         "type": "string"
                                     },
-                                    "patternProperties": {
-                                        "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$": {
-                                            "type": "string"
-                                        }
-                                    },
                                     "type": "object"
                                 }
                             ],
@@ -112,14 +107,6 @@
                                             "$ref": "#/definitions/YamlCompressorSystemOperationalSettings"
                                         },
                                         "type": "array"
-                                    },
-                                    "patternProperties": {
-                                        "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$": {
-                                            "items": {
-                                                "$ref": "#/definitions/YamlCompressorSystemOperationalSettings"
-                                            },
-                                            "type": "array"
-                                        }
                                     },
                                     "type": "object"
                                 }
@@ -353,11 +340,6 @@
                                     "additionalProperties": {
                                         "$ref": "https://test.ecalc.equinor.com/api/v1/schema-validation/energy-usage-model.json#properties/ENERGY_USAGE_MODEL"
                                     },
-                                    "patternProperties": {
-                                        "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$": {
-                                            "$ref": "https://test.ecalc.equinor.com/api/v1/schema-validation/energy-usage-model.json#properties/ENERGY_USAGE_MODEL"
-                                        }
-                                    },
                                     "type": "object"
                                 }
                             ],
@@ -399,11 +381,6 @@
                                     "additionalProperties": {
                                         "$ref": "https://test.ecalc.equinor.com/api/v1/schema-validation/energy-usage-model.json#properties/ENERGY_USAGE_MODEL"
                                     },
-                                    "patternProperties": {
-                                        "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$": {
-                                            "$ref": "https://test.ecalc.equinor.com/api/v1/schema-validation/energy-usage-model.json#properties/ENERGY_USAGE_MODEL"
-                                        }
-                                    },
                                     "type": "object"
                                 }
                             ],
@@ -418,11 +395,6 @@
                                 {
                                     "additionalProperties": {
                                         "type": "string"
-                                    },
-                                    "patternProperties": {
-                                        "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$": {
-                                            "type": "string"
-                                        }
                                     },
                                     "type": "object"
                                 }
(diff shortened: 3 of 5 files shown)


TeeeJay:
The real work is not done, but prework in order to work with temporal models for operational settings. This should be possible to merge, as it only fixes, tests and documents previous functionality, done in 2 separate commits to be looked at separately. Can be split in 2 PRs, if needed.

Ideally this PR should be merged before the intended functionality is added.

Must clarify why 2 of the tests added are failing. May be due to misunderstanding of how `_get_crossover_rates` works

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/consumer_system.py
```diff
@@ -279,8 +279,31 @@ def _topologically_sort_consumers_by_crossover(crossover: List[int], consumers:
...
         max_rate: List[float], rates: List[List[float]]
     ) -> Tuple[NDArray[np.float64], List[List[float]]]:
+        """
+        This function is run over a single consumer only, and is normally run in a for loop
+        across all "dependent" consumers in the consumer system, such as here, in a consumer system.
+
+        This function will calculate:
+         1. the additional crossover rate that is required if the total incoming
+        rate exceeds the max rate capacity for the consumer in question.
+         2. the rates within capacity of the given consumer
```
jsolaas, line 289: Nice explanation, could this be under Returns or is that unclear when using a tuple return?

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/variable.py
```diff
@@ -16,6 +16,13 @@ def to_dto(self):
 
 
 class DefaultDatetime(datetime):
+    """
```
jsolaas, line 19: We could move this class also

src/ecalc/libraries/libecalc/common/tests/core/consumers/test_crossover_rates.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # Exceeds capacity, cross over required - bug? leads to expected_rates_within_capacity = [[6, 6], [0, 0]]
+    # max_rate = [4, 4]  # 4 for both timesteps
+    # rates = [[5, 5], [0, 0]]  # 4 in rate
+    # expected_crossover_rate = [1, 1]  # 1 exceeding
+    # expected_rates_within_capacity = [[4, 4], [0, 0]]
+    #
+    # crossover_rate, rates_within_capacity = ConsumerSystem._get_crossover_rates(max_rate, rates)
+    # assert (np.array_equal(crossover_rate, expected_crossover_rate))
+    # assert (np.array_equal(rates_within_capacity, expected_rates_within_capacity))
```
jsolaas, line 74: Seems like a bug to me

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/consumer_system.py
TeeeJay: Forgot about returns, yes, the idea was that it should be there :p

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/variable.py
TeeeJay: Just wanted to start doing things right...not too much at the same time, but I agree :)

src/ecalc/libraries/libecalc/common/tests/core/consumers/test_crossover_rates.py
TeeeJay: Ok, so you agree that both my tests should succeeed?

jsolaas: Yes

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -181,6 +181,16 @@ def for_period(self, period: Period) -> Self:
             unit=self.unit,
         )
 
+    def for_timestep(self, current_timestep: datetime) -> Self:
+        timestep_index = next(
+            timestep_index for timestep_index, timestep in enumerate(self.timesteps) if timestep == current_timestep
+        )
```
jsolaas, line 187: `timestep_index = self.timesteps.index(current_timestep)`?

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -660,6 +670,18 @@ def for_period(self, period: Period) -> Self:
...
+        return self.__class__(
+            timesteps=self.timesteps[timestep_index : timestep_index + 1],
+            values=self.values[timestep_index : timestep_index + 1],
+            regularity=self.regularity[timestep_index : timestep_index + 1],  # type: ignore
```
jsolaas, line 680: Why ignore type?

src/ecalc/libraries/libecalc/common/libecalc/dto/core_specs/compressor/operational_settings.py
```diff
@@ -27,6 +27,18 @@ def get_subset_from_period(self, period: Period) -> Self:
             timesteps=self.timesteps[start_index:end_index],
         )
 
+    def get_subset_for_timestep(self, current_timestep: datetime) -> Self:
+        timestep_index = next(
+            timestep_index for timestep_index, timestep in enumerate(self.timesteps) if current_timestep == timestep
+        )
```
jsolaas, line 33: index method

src/ecalc/libraries/libecalc/common/libecalc/dto/core_specs/compressor/operational_settings.py
```diff
@@ -27,6 +27,18 @@ def get_subset_from_period(self, period: Period) -> Self:
             timesteps=self.timesteps[start_index:end_index],
         )
 
+    def get_subset_for_timestep(self, current_timestep: datetime) -> Self:
```
jsolaas, line 30: for_timestep here also?

src/ecalc/libraries/libecalc/common/libecalc/dto/variables.py
```diff
@@ -49,5 +49,12 @@ def get_subset_from_period(self, period: Period) -> VariablesMap:
         start_index, end_index = period.get_timestep_indices(self.time_vector)
         return self.get_subset(start_index, end_index)
 
+    def get_subset_for_timestep(self, current_timestep: datetime) -> VariablesMap:
```
jsolaas, line 52: Use the same name for this everywhere?

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/consumer_system.py
```diff
@@ -138,34 +150,41 @@ def evaluate(
...
+
+        operational_settings_results: Dict[datetime, Dict[int, Dict[str, EcalcModelResult]]] = defaultdict(
             dict
         )  # map operational settings index and consumer id to consumer result.
 
-        for period, operational_settings in temporal_operational_settings.items():
-            variables_map_for_period = variables_map.get_subset_from_period(period)
-            start_index, end_index = period.get_timestep_indices(variables_map.time_vector)
+        # for period, operational_settings in temporal_operational_settings.items():
+        for timestep_index, timestep in enumerate(variables_map.time_vector):
```
jsolaas, line 159: I guess this is fine? Is this something we want to do for all v2 models? i.e. in compressor and pump also? So all models are evaluated with only a single timestep?

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
TeeeJay: copy paste...idk :p i think it was a bug where i copied from, that fixed it easily with ignoring it..

src/ecalc/libraries/libecalc/common/libecalc/core/consumers/consumer_system.py
TeeeJay: Waoh. I didnt expect you to agree on this, since it is, as you say, changing in general the way we evaluate stuff to timesteps instead of periods.

I didnt add it to the PR, thought about discussing it, but in general, it was (imo) so complicated to get things right with temporal models overlapping temporal models and so on, that at least the initial approach had to be timestep by timestep, to be sure to get it right...

However, I realized after mentioning it the other day, that this will probably be in many case hard to cache since the runtime variables change for each timestep, so it will most definitely be slower than the previous solution. Therefore, it makes more sense to be to bring this "timestep by timestep" even one more layer/level out, to create eCalc models for each timestep (WITHOUT temporal models to deal with, only do that at evaluation/preprocessing step)...but there will be a performance penalty I think. We should definitely discuss it...

I havent thought too much about it wrt. consequences of performance penalty etc, but it will remove a lot of unnecessary complexity (for developers) and potentially add benefits for users, that I think might be worth it...

conftest.py
```diff
@@ -1 +1 @@
-pytest_plugins = ["libecalc.fixtures", "libecalc.examples"]
+pytest_plugins = ["libecalc.fixtures", "libecalc.examples"]  # where to find fixtures
```
jsolaas, line 1: Hva mener du med kommentaren?

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -583,11 +593,13 @@ class TimeSeriesRate(TimeSeries[float]):
...
     typ: Optional[RateType] = RateType.STREAM_DAY
-    regularity: Optional[List[float]]
+    regularity: Optional[List[float]]  # TODO: Optional? No, we need to have it...default to 1...
 
     @validator("regularity", pre=True, always=True)
     def set_regularity(cls, regularity: Optional[List[float]], values: Dict[str, Any]) -> List[float]:
-        if regularity is not None:
+        if (
+            regularity is not None and regularity != []
+        ):  # TODO: empty list..why? the regularity list and values list are off...
```
jsolaas, line 602: Why are you asking me? You added it!

src/ecalc/libraries/libecalc/common/libecalc/common/utils/rates.py
```diff
@@ -823,4 +847,7 @@ def reindex(self, new_time_vector: Iterable[datetime]) -> TimeSeriesRate:
         Ensure to map correct value to correct timestep in the final resulting time vector.
         """
         reindex_values = self.reindex_time_vector(new_time_vector)
-        return TimeSeriesRate(timesteps=new_time_vector, values=reindex_values.tolist(), unit=self.unit)
+        # TODO: Regularity?
+        return TimeSeriesRate(
+            timesteps=new_time_vector, values=reindex_values.tolist(), unit=self.unit, regularity=self.regularity
```
jsolaas, line 852: need to reindex regularity right?

src/ecalc/libraries/libecalc/common/tests/integration/test_consumer_system_v2.py
```diff
@@ -26,10 +30,34 @@ def result(consumer_system_v2_dto) -> EnergyCalculatorResult:
...
+
+@pytest.mark.parametrize(
+    "name, consumer_system_v2",
+    parameterized_v2_parameters,
+)
 @pytest.mark.snapshot
-def test_compressor_system_v2_results(
-    result: EnergyCalculatorResult, rounded_snapshot, consumer_system_v2_dto: DTOCase
-):
+def test_compressor_system_v2_results(name: str, consumer_system_v2: DTOCase, request):
```
jsolaas, line 60: I guess I would prefer to split the test, makes it easier to debug one of them. Or all in one as before, but separate tests and separate cases seems like a good idea

jsolaas (commented):
I would prefer it if we could get done with something and merge then continue discussion on new stuff in a separate PR. I.e. if we want to go for the single-timestep solution then merge it, if we don't want to go for it then it would be nice to be able to merge the changes in the tests.

Just whip out a new branch instead of continuing when you see something else you want to change. Merge the "safe" change, then rebase this branch

jsolaas (approved):

PR #155, opened 2023-09-08
fix: requested pressures not always an attribute

jsolaas:
CompressorSampled does not have a suction pressure and crashed ecalc before this change.

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/core/graph_result.py
@@ -2,7 +2,7 @@ import math
 import operator
 from datetime import datetime
 from functools import reduce
-from typing import Any, Dict, List
+from typing import Dict, List
 
 import libecalc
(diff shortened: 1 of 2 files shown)


TeeeJay (approved):
Is this still relevant for the new changes wrt adding this to train only (and hence systems?) Needs to be fixed anyways, as long as the new implementation is not in place.

I would prefer that if the docstring is only autogenerated, and has no info, it would be better to leave it out for now (use git commit -n) instead of adding a docstring that doesnt add any value, and rather improve the docstring later

jsolaas:
> Is this still relevant for the new changes wrt adding this to train only (and hence systems?) Needs to be fixed anyways, as long as the new implementation is not in place.

I don't know

> 
> I would prefer that if the docstring is only autogenerated, and has no info, it would be better to leave it out for now (use git commit -n) instead of adding a docstring that doesnt add any value, and rather improve the docstring later

Makes sense

PR #157, opened 2023-09-08
refactor: generate asset/ecalc model schema

jsolaas:

Using pydantic classes and schema generation should make it easier for us to keep json-schema updated. We just have to start using the pydantic yaml classes as validation for ecalc as well.

Replaces the ecalc-model.json schema with a pydantic generated one (with some changes as it is WIP)


src/ecalc/libraries/libecalc/common/libecalc/input/validation/generate_json_schema.py
```diff
@@ -33,18 +33,58 @@ def get_template(schema: str, schema_name: str, is_root: bool = False) -> str:
     return template
 
 
+def replace_property_with_legacy_json_schema(schema: dict, property_key: str, property_ref: str) -> dict:
```
TeeeJay, line 36: for which JSON Schema version/revision does this apply to? Is it due to our requirements of an older version of the schema than what is recommended ?

src/ecalc/libraries/libecalc/common/libecalc/input/validation/generate_json_schema.py
```diff
@@ -33,18 +33,58 @@ def get_template(schema: str, schema_name: str, is_root: bool = False) -> str:
...
+        property_ref="$SERVER_NAME/api/v1/schema-validation/fuel-types.json#properties/FUEL_TYPES",
+    )
+    schema = replace_property_with_legacy_json_schema(
+        schema=schema,
+        property_key="INSTALLATIONS",
+        property_ref="$SERVER_NAME/api/v1/schema-validation/installations.json#properties/INSTALLATIONS",
+    )
+
+    variables_schema = schema["schema"]["properties"]["VARIABLES"]
+    # Patterned properties not supported in monaco-yaml, replace with additionalProperties.
```
TeeeJay, line 79: In order to make this portable, we should be able to generate whatever "revision" is needed by end user? For us, it will be the "legacy" version, for other usage (e.g. when loading with PyYAML, Ruamel, or some external usage) it may be relevant to use the current version?

src/ecalc/libraries/libecalc/common/libecalc/input/validation/generate_json_schema.py
```diff
@@ -33,18 +33,58 @@ def get_template(schema: str, schema_name: str, is_root: bool = False) -> str:
...
+    monaco_variables_schema = {key: value for key, value in variables_schema.items() if key != "patternProperties"}
```
TeeeJay, line 80: should not refer to monaco here, but rather use different versions, ie. json schema version/revision, if that is correct (unless e.g. monaco and others have their own implementation)

src/ecalc/libraries/libecalc/common/libecalc/input/validation/json_schemas/ecalc-model.json
```diff
@@ -1,40 +0,0 @@
-{
```
TeeeJay, line 1: We may want to have the static json files (as long as they are aligned with the python code; can e.g. add a test on this?) for simpler use?

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/asset.py
```diff
@@ -0,0 +1,62 @@
...
+
+from libecalc.input.yaml_types import YamlBase
+from libecalc.input.yaml_types.variable import Variables
+from pydantic import Field
+
+# Dummy type used for attributes that aren't yet converted to pydantic yaml models
+DummyType = Optional[Dict]
+
+
+class YamlAsset(YamlBase):
```
TeeeJay, line 11: We already got feedback that `asset` is not a representative name, as this varies a lot. We should however find a name that makes sense. But as long as this does not "bleed" over to users, it is ok for the time being

src/ecalc/libraries/libecalc/common/tests/input/validation/snapshots/test_validation_json_schemas/test_json_schema_changed/schemas.json
```diff
@@ -32,20 +70,32 @@
                 },
                 "TIME_SERIES": {
                     "$ref": "https://test.ecalc.equinor.com/api/v1/schema-validation/time-series.json#properties/TIME_SERIES",
-                    "description": "Defines the inputs for time dependent variables, or \u201creservoir variables\u201d.\n\nhttps://test.ecalc.equinor.com/docs/docs/modelling/keywords/TIME_SERIES",
+                    "description": "Defines the inputs for time dependent variables, or 'reservoir variables'.\n\nhttps://test.ecalc.equinor.com/docs/docs/modelling/keywords/TIME_SERIES",
                     "title": "TIME_SERIES"
                 },
                 "VARIABLES": {
-                    "$ref": "https://test.ecalc.equinor.com/api/v1/schema-validation/definitions.json#definitions/VARIABLES"
+                    "additionalProperties": {
```
TeeeJay, line 77: I assume this was missing in the old json schema, but should have been there?

src/ecalc/libraries/libecalc/common/tests/input/validation/snapshots/test_validation_json_schemas/test_json_schema_changed/schemas.json
```diff
@@ -32,20 +70,32 @@
...
+                    "title": "VARIABLES",
+                    "type": "object"
                 }
             },
             "required": [
-                "INSTALLATIONS",
-                "TIME_SERIES",
-                "FACILITY_INPUTS",
-                "FUEL_TYPES"
+                "INSTALLATIONS"
```
TeeeJay, line 96: I think this is correct, great to see that schema and code is aligned :)

src/ecalc/libraries/libecalc/common/libecalc/input/validation/generate_json_schema.py
jsolaas: Ah, legacy_json_schema refers to our previous implementation (json-schema files), i.e. replace the asset property TIME_SERIES with a `$ref` to the `time-series.json` json-schema. So this is a temporary thing until we can create all the pydantic classes for yaml

src/ecalc/libraries/libecalc/common/libecalc/input/validation/generate_json_schema.py
jsolaas: I guess we could do that. If we were to generate json-schema for validation in pycharm for example, and the json-schema version there is different from monaco-editor.

src/ecalc/libraries/libecalc/common/libecalc/input/validation/generate_json_schema.py
jsolaas: I don't know which revision monaco supports, so monaco was the only thing I had. I guess it is monaco-yaml that decides that, or maybe the language server.

src/ecalc/libraries/libecalc/common/libecalc/input/validation/json_schemas/ecalc-model.json
jsolaas: We can generate json files, if that is what you mean. If you are talking about a test while converting I did that, and removed it when I saw it worked similarly.

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/asset.py
jsolaas: It does bleed, and I need to set a name. EcalcModel and Asset are the two that I know we use right now. I decided to keep using Asset while we still have dto.Asset

jsolaas: It might bleed, not sure if it does currently

src/ecalc/libraries/libecalc/common/tests/input/validation/snapshots/test_validation_json_schemas/test_json_schema_changed/schemas.json
jsolaas: The difference is that it was referenced before, now we are adding it directly or in definitions. So we are going to end up with a single schema "file" instead of several.

src/ecalc/libraries/libecalc/common/tests/input/validation/snapshots/test_validation_json_schemas/test_json_schema_changed/schemas.json
jsolaas: Yes, we might need a conditional required here, not sure how to handle that. There are several combinations that is enough, i.e. Start, end and a direct consumer. 

I think FUEL_TYPES might be required though. Or do you have an example where it isn't?

PR #160, opened 2023-09-11
refactor: generate fuel consumer schema

jsolaas:


src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/asset.py
```diff
@@ -7,12 +7,40 @@
 from pydantic import Field
 
 
+def replace_asset_property_with_legacy_json_schema(schema: dict, property_key: str, property_ref: str) -> None:
```
TeeeJay, line 10: Comment on why we have this?

TeeeJay: In the docstring I mean :)

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/installation.py
```diff
@@ -1,16 +1,39 @@
...
 from libecalc.input.yaml_types.components.category import CategoryField
+from libecalc.input.yaml_types.components.legacy.yaml_fuel_consumer import (
+    YamlFuelConsumer,
+)
 from libecalc.input.yaml_types.placeholder_type import PlaceholderType
 from libecalc.input.yaml_types.temporal_model import TemporalModel
 from pydantic import Field
 
 
+def replace_installation_property_with_legacy_json_schema(schema: dict, property_key: str, property_ref: str) -> None:
```
TeeeJay, line 15: possible to have this "util function" as a helper to yaml for now instead of replicating in every "yaml module"?

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/legacy/yaml_fuel_consumer.py
```diff
@@ -0,0 +1,39 @@
...
+
+from libecalc.dto.base import ConsumerUserDefinedCategoryType
+from libecalc.input.yaml_types import YamlBase
+from libecalc.input.yaml_types.components.category import CategoryField
+from libecalc.input.yaml_types.placeholder_type import PlaceholderType
+from libecalc.input.yaml_types.temporal_model import TemporalModel
+from pydantic import Field
+
+
+class YamlFuelConsumer(YamlBase):
```
TeeeJay, line 11: include Legacy or similar in the name of the class as well? Do we have a "non-legacy fuel consumer"? What is the reason why the legacy yaml type is included here? Plan to remove/change it?

TeeeJay (approved):
dont work so fast ;) :p see comments :)

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/legacy/yaml_fuel_consumer.py
jsolaas: By using legacy I'm trying to make it clear which files are connected across yaml,dto,core etc. If we end up creating several versions for each type (system v2,v3,v4 etc), then we probably need to group those classes in a similar way (by version, not `legacy`)

PR #161, opened 2023-09-11
fix: generate system v2 schema

jsolaas:


.pre-commit-config.yaml
```diff
@@ -24,6 +24,8 @@ repos:
       - id: codespell
         name: Check for typos in code
         additional_dependencies: [tomli]
+        exclude_types:
```
TeeeJay, line 27: what is it complaining on?

TeeeJay: \n or that we have a lot of typos in docstrings? or that it doesnt make sense because there are non-words that we dotn want to fix?

TeeeJay (approved):

.pre-commit-config.yaml
jsolaas: Complained about `nin` not being correct, combination of `\n` and `in` without space

PR #162, opened 2023-09-11
fix: expression type in system v2

jsolaas:

Only str was allowed for expressions in the yaml models for system v2 (yellow mark because of json-schema)

Sets the type of expression to a union of int, float and str

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/base.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/base.py
@@ -1,5 +1,6 @@
 from typing import List, Optional
 
+from libecalc.expression.expression import ExpressionType
 from libecalc.input.yaml_types import YamlBase
 from pydantic import Field
 
@@ -21,7 +22,7 @@ class ConsumerBase(YamlBase):
 
 
 class OperationalConditionBase(YamlBase):
-    condition: Optional[str] = Field(
+    condition: Optional[ExpressionType] = Field(
         None,
         title="Condition",
         description="""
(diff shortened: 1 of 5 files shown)


src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/pump_system.py
```diff
@@ -204,7 +205,8 @@ def to_dto(
                 else:
                     rates = [
                         Expression.multiply(
-                            Expression.setup_from_expression(self.rate), Expression.setup_from_expression(rate_fraction)
+                            Expression.setup_from_expression(operational_setting.total_system_rate),
```
jsolaas, line 208: oops, forgot to mention I fixed a bug

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/base.py
```diff
@@ -21,7 +22,7 @@ def to_dto(self, **kwargs):
 
 
 class OperationalConditionBase(YamlBase):
-    condition: Optional[str] = Field(
```
TeeeJay, line 24: :O

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/pump_system.py
TeeeJay: nice to catch this..another "pros" of having a type/alias rather than typing the Union etc all the time...

TeeeJay: Ideally separate commit though...

TeeeJay (approved):

PR #164, opened 2023-09-12
fix: what fluid to recirculate when there is no rate entering a compressor stage in a multiple streams compressor train at a given time step

olelod:
This regards VairableSpeedCompressorTrainCommonShaftMultipleStreamsAndPressures only. In situations where there is both an additional ingoing stream and an additional ingoing stream between two compressors, one can end up with a situation where one tries to mix together two streams with zero mass rate (to mix the one stream passing from compressor 1 to compressor 2 and the additional stream entering the compressor train between compressor 1 and compressor 2). This means that if we look at this time step isolated there is an unknown composition to recirculate in the second compressor. This fluid to recirculate should have the same composition as the fluid passing through the second compressor at the previous time step.

- Includes a class attribute keeping track of the composition of the fluid passing through the different compressor stages in a compressor train, such that it will be available for the next time step (except for the first time step - there the code will fail) 
- A small fix giving the correct standard condition density for the resulting fluid with mixing two fluids (now using the correct pressure and temperature, instead of assuming that the given pressure and temperature is the standard condition pressure and standard condition temperature)
- Adding a test

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
@@ -35,8 +35,6 @@ class FluidStream:
             existing_fluid: Initialize FluidStream from an existing (NeqSim) fluid. Warning: Be mindful of keeping fluid_model and existing fluid consistent. If not the fluid properties may be incorrect.
         """
         self.fluid_model = fluid_model
-        self.initial_temperature_kelvin = temperature_kelvin
-        self.initial_pressure_bare = pressure_bara
 
         if not temperature_kelvin > 0:
             raise ValueError("FluidStream temperature needs to be above 0.")
@@ -60,8 +58,8 @@ class FluidStream:
         self._enthalpy_joule_per_kg = _neqsim_fluid_stream.enthalpy_joule_per_kg
 
         if (
-            pressure_bara == UnitConstants.STANDARD_PRESSURE_BARA
-            and temperature_kelvin == UnitConstants.STANDARD_TEMPERATURE_KELVIN
+            self._pressure_bara == UnitConstants.STANDARD_PRESSURE_BARA
+            and self._temperature_kelvin == UnitConstants.STANDARD_TEMPERATURE_KELVIN
         ):
             self.standard_conditions_density = _neqsim_fluid_stream.density
             self.molar_mass_kg_per_mol = _neqsim_fluid_stream.molar_mass
@@ -72,8 +70,8 @@ class FluidStream:
                 pressure_bara=UnitConstants.STANDARD_PRESSURE_BARA,
                 eos_model=self.fluid_model.eos_model,
             )
-
             self.standard_conditions_density = _neqsim_fluid_at_standard_conditions.density
+            self.molar_mass_kg_per_mol = _neqsim_fluid_stream.molar_mass
 
     @property
     def pressure_bara(self) -> float:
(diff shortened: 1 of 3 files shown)


src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
```diff
@@ -277,7 +278,5 @@ def mix_in_stream(
 
         return FluidStream(
             existing_fluid=mixed_neqsim_fluid_stream,
-            temperature_kelvin=temperature_kelvin,
```
TeeeJay, line 280: so, this was a bug?

TeeeJay: or just unnecessary?

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/variable_speed_compressor_train_common_shaft_multiple_streams_and_pressures.py
```diff
@@ -879,7 +892,7 @@ def calculate_compressor_train_given_rate_ps_speed(
             )
 
             mass_rate_previous_stage_kg_per_hour = mass_rate_this_stage_kg_per_hour
-
+            self.fluid_to_recirculate_in_stage[stage_number] = inlet_stream
```
TeeeJay, line 895: possibly to replace set and get to this member variable using methods to more clearly indicate that we are adding "leftover rate" in one timestep that is "captured" in next timestep? e.g.

`set_recirculate_fluid_in_next_timestep(stage, stream)`

and

`get_fluid_to_recirculate(stage)`

or some other names that make more sense? Does it make sense?

TeeeJay: But is it actually "next timestep"? Isnt it "next recirculation round" or whatever name is being used by experts? to not be confused with "ecalc timesteps"?

TeeeJay: unless, I am wrong ;)

src/ecalc/libraries/libecalc/common/tests/core/models/compressor_modelling/test_variable_speed_compressor_train_multiple_streams.py
```diff
@@ -219,6 +219,41 @@ def variable_speed_compressor_train_two_compressors_two_streams(
     )
 
 
+@pytest.fixture
```
TeeeJay, line 222: :+1:

TeeeJay: great with test. and it seems very minimal. Win! :)

We should have more tests like these, for all the edge cases we can think of. Just a reminder for future and improvement of tests coverage :)

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/fluid.py
olelod: Just unnecessary. When existing_fluid is defined, the pressure and temperature is taken from there. The bug was line 61 and 62. pressure_bara and temperature_kelvin was always defaulted to UnitConstants.STANDARD_PRESSURE_BARA and UnitConstants.STANDARD_TEMPERATURE_KELVIN, so the if statement would always evaluate to true...

src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/train/variable_speed_compressor_train_common_shaft_multiple_streams_and_pressures.py
olelod: > possibly to replace set and get to this member variable using methods to more clearly indicate that we are adding "leftover rate" in one timestep that is "captured" in next timestep? e.g.
> 
> `set_recirculate_fluid_in_next_timestep(stage, stream)`
> 
> and
> 
> `get_fluid_to_recirculate(stage)`
> 
> or some other names that make more sense? Does it make sense?

Yes, surely. Was a bit unsure how to approach it, but that sounds better.

TeeeJay (approved):

frodehk (approved):

PR #165, opened 2023-09-12
refactor: generate generator set schema

jsolaas:


src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/temporal_model.py
```diff
@@ -1,5 +1,11 @@
...
+from pydantic import ConstrainedStr
+
+
+class DatetimeString(ConstrainedStr):
+    regex = "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$"
+
 
 TModel = TypeVar("TModel")
-TemporalModel = Union[TModel, MutableMapping[datetime, TModel]]
+TemporalModel = Union[TModel, Dict[DatetimeString, TModel]]
```
jsolaas, line 11: We need to figure out how to do this, the schema for datetime does not enforce anything on the key, while a constr gives us the same regex as before. It still does not give a validation error in the editor, while in test it does, I'm assuming this is because we have 'additionalProperties' in the pydantic generated temporal model while the old schemas does not.

src/ecalc/libraries/libecalc/common/tests/input/yaml_types/test_temporal_model.py
```diff
@@ -0,0 +1,25 @@
...
+            "anyOf": [
+                {
+                    "type": "string",
+                },
+                {
+                    "type": "object",
+                    "patternProperties": {
+                        "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$": {"type": "string"}
+                    },
+                    "additionalProperties": {"type": "string"},
```
jsolaas, line 22: `additionalProperties` added by pydantic, it does not exist in the old schemas

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/generator_set.py
```diff
@@ -0,0 +1,50 @@
+from typing import Any, Dict, List, Type
+
+from libecalc.dto.base import ConsumerUserDefinedCategoryType
```
TeeeJay, line 3: Just want to say that I agree on this (I think), that the yaml-models depend on DTO (the common input layer that we have now...), instead of replicating it.

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/generator_set.py
```diff
@@ -0,0 +1,50 @@
...
+from libecalc.input.yaml_types import YamlBase
+from libecalc.input.yaml_types.components.category import CategoryField
+from libecalc.input.yaml_types.components.legacy.yaml_electricity_consumer import (
+    YamlElectricityConsumer,
+)
+from libecalc.input.yaml_types.placeholder_type import PlaceholderType
+from libecalc.input.yaml_types.schema_helpers import (
+    replace_temporal_placeholder_property_with_legacy_ref,
+)
+from libecalc.input.yaml_types.temporal_model import TemporalModel
```
TeeeJay, line 13: should we be consistent and rename _all_ yaml types to have **Yaml** as a prefix? WIP perhaps?

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/legacy/yaml_fuel_consumer.py
```diff
@@ -1,37 +1,14 @@
...
-from libecalc.input.yaml_types.placeholder_type import PlaceholderType
+from libecalc.input.yaml_types.components.legacy.yaml_electricity_consumer import (
+    YamlElectricityConsumer,
+)
 from libecalc.input.yaml_types.temporal_model import TemporalModel
 from pydantic import Field
 
 
-class YamlFuelConsumer(YamlBase):
+class YamlFuelConsumer(YamlElectricityConsumer):
```
TeeeJay, line 8:
```suggestion
class YamlFuelConsumer(YamlElectricityConsumer):
```
not sure what you are thinking here. Doesnt this add some fields from `YamlElectricityConsumer` that we do not want?

jsolaas: No, the only difference I found was fuel. We can split if that isn't the case

src/ecalc/libraries/libecalc/common/tests/input/yaml_types/test_temporal_model.py
TeeeJay: not sure if I understand this. Is it correct? Which case does it cover?

TeeeJay (approved):

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/legacy/yaml_fuel_consumer.py
TeeeJay: I think so, or alternatively add a new "top YamlConsumer" that is abstract...since this inheritance is very weird...

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/temporal_model.py
TeeeJay: ah, ok, so we do not really want the "additionalProperties" then?

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/temporal_model.py
```diff
@@ -1,5 +1,11 @@
-from datetime import datetime
-from typing import MutableMapping, TypeVar, Union
+from typing import Dict, TypeVar, Union
+
+from pydantic import ConstrainedStr
+
+
+class DatetimeString(ConstrainedStr):
+    regex = "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$"
```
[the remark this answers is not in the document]
jsolaas: regex copied from json-schema. do you mean if date is missing from the regex or that the datetime type is missing?

datetime is missing, that's why I added a comment. Currently we have to either get json-schema validation or datetime parsing.

I'm going to remove this change, then create an issue.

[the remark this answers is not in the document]
jsolaas: Or should I keep it?

src/ecalc/libraries/libecalc/common/tests/input/yaml_types/test_temporal_model.py
jsolaas: It describes the type of additional properties, but not the keys. So it accepts any key, while we want it to only accept the patternProperties. This is an issue we have with variables and temporal models. We'll have to look into it.

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/temporal_model.py
jsolaas: correct, at least if monaco-editor can handle that schema

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/generator_set.py
jsolaas: wip yes

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/components/legacy/yaml_fuel_consumer.py
jsolaas: a new top YamlLegacyConsumer could work I guess, then we'd get 
```
class YamlElectricityConsumer(YamlLegacyConsumer):
    pass
```

I copy-pasted, no common parent

src/ecalc/libraries/libecalc/common/libecalc/input/yaml_types/temporal_model.py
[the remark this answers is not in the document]
jsolaas: I'm keeping it!

PR #168, opened 2023-09-12
ECALC-21 chore: improve dto validation error message

frodehk:

Difficult to understand what the dto validation is related to when there is no reference to the consumer name.

Improve dto validation error to include consumer/component name.

https://equinor-ecalc.atlassian.net/browse/ECALC-21?atlOrigin=eyJpIjoiNjhhMzljMTg5YTk3NDViODgyZGM2ZWY0ZjM0YjM1ODIiLCJwIjoiaiJ9

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/input/validation_errors.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/input/validation_errors.py
@@ -7,6 +7,7 @@ import pydantic
 import yaml
 from libecalc.common.logger import logger
 from libecalc.input.yaml_entities import Resource, YamlDict, YamlList
+from libecalc.input.yaml_keywords import EcalcYamlKeywords
 from yaml import Dumper, Mark
 
 Loc = Tuple[Union[int, str], ...]
@@ -116,6 +117,7 @@ class DtoValidationError(DataValidationError):
     ):
         errors = validation_error.errors()
 
+        component_name = data[EcalcYamlKeywords.name]
         messages = []
         error_locs = []
         try:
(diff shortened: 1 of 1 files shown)


src/ecalc/libraries/libecalc/common/libecalc/input/validation_errors.py
```diff
@@ -116,6 +117,7 @@ def __init__(
     ):
         errors = validation_error.errors()
 
+        component_name = data[EcalcYamlKeywords.name]
```
TeeeJay, line 120: Are we sure that the `EcalcYamlKeywords.name` is always a part of this dict? Just asking... Better to:

```
component_name = data.get(EcalcYamlKeywords.name, "Missing Component Name")
```

?

TeeeJay (approved):

src/ecalc/libraries/libecalc/common/libecalc/input/validation_errors.py
jsolaas, line 120: I don't think it is guaranteed that name is available, so we should check if name is available then add it to the message if it is.

frodehk: I was thinking the same, but assumed it was safe when all tests passed. Will update.

TeeeJay:
perhaps "feat: " is a bit extreme in this case? More like a "chore:"? or we have another alternative? Maybe it is better and more visible that we annotate it as "feat." ?

src/ecalc/libraries/libecalc/common/libecalc/input/validation_errors.py
jsolaas: "Missing component name" does not make a lot of sense if the dto validation error is for fuel or facility input. We should probably figure out where to 'catch' these errors, but currently it can be something different than a component.

jsolaas (approved):
Looks good!

PR #169, opened 2023-09-13
Fix/use common period class

jsolaas:

FDE did not work with system v2

Refactor so we use Period classes from time_utils instead of TimeInterval defined in the flow_diagram module.
Add support for system v2 in FDE

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/common/time_utils.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/common/time_utils.py
@@ -52,19 +52,37 @@ class Periods:
     periods: List[Period]
 
     @classmethod
-    def create_periods(cls, times: List[datetime]) -> Periods:
+    def create_periods(cls, times: List[datetime], include_before: bool = True, include_after: bool = True) -> Periods:
+        """
+        Create periods from the provided datetimes
+        :param times: the datetimes to create periods from
+        :param include_before: whether to add a period that ends with the first provided datetime, i.e. define a period
+        before the earliest provided datetime.
+        :param include_after: whether to add a period that starts with the last provided datetime, i.e. define a period
+        after the latest provided datetime.
+        :return:
+        """
         if len(times) == 0:
             return cls([])
 
-        return cls(
-            [
+        periods = []
+
+        if include_before:
+            periods.append(
                 Period(
                     end=times[0],
-                ),
-                *[Period(start=times[index], end=times[index + 1]) for index in range(len(times) - 1)],
-                Period(start=times[-1]),
-            ]
-        )
+                )
+            )
+
+        periods.extend([Period(start=times[index], end=times[index + 1]) for index in range(len(times) - 1)])
+
+        if include_after:
+            periods.append(Period(start=times[-1]))
+
+        return cls(periods)
+
+    def __iter__(self):
+        return self.periods.__iter__()
 
     def get_period(self, time: datetime) -> Period:
         for period in self.periods:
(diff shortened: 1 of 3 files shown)


src/ecalc/libraries/libecalc/common/libecalc/output/flow_diagram/EcalcModelMapper.py
```diff
@@ -138,6 +129,56 @@ def _create_compressor_system_diagram(
...
+            title=pump_system.name,
+            start_date=min(timesteps),
+            end_date=max([*timesteps, global_end_date]),
+            nodes=[
+                Node(
+                    id=compressor.name,
+                    title=compressor.name,
+                    type=NodeType.COMPRESSOR,
+                )
+                for compressor in pump_system.pumps
```
jsolaas, line 174: All shown with compressor icons, legacy system looks at the model and displays compressor with turbine different than a compressor train for example.

src/ecalc/libraries/libecalc/common/libecalc/common/time_utils.py
```diff
@@ -52,19 +52,37 @@ class Periods:
     periods: List[Period]
 
     @classmethod
-    def create_periods(cls, times: List[datetime]) -> Periods:
+    def create_periods(cls, times: List[datetime], include_before: bool = True, include_after: bool = True) -> Periods:
```
TeeeJay, line 55: This will potentially fail with previous functionality, that defaulted to `False` for both `include_before` and `include_after`. Isnt it better to default to `False` then? Or do you have overview of all usage and that it will still work (or work better?) with the new functionality?

TeeeJay: maybe time to ask if we can enforce start and end time for ecalc models? ...

src/ecalc/libraries/libecalc/common/tests/common/test_time_utils.py
```diff
@@ -88,6 +88,46 @@ def test_three_dates(self):
             ]
         )
 
+    def test_three_dates_not_include_before(self):
+        first_date = datetime(2020, 1, 1)
+        second_date = datetime(2022, 1, 1)
+        third_date = datetime(2022, 1, 1)
```
TeeeJay, line 94: did you intend to set second and third date to the same date? I guess this can happen in the code..but imo. this will instead treat symptom and there is another underlying problem...? Might be that we have chosen to postpone this for v2 due to backwards compatibiltiy etc though...

TeeeJay:
Great changes and refactoring, but I do not quite understand the reason why this was needed when 2 periods were the same date...can we treat the problem instead of the symptom?

src/ecalc/libraries/libecalc/common/libecalc/common/time_utils.py
jsolaas: Default was True, right? So I played it safe. That's what I tried to do at least

src/ecalc/libraries/libecalc/common/tests/common/test_time_utils.py
jsolaas: No, I did not. Copy paste of the wrong test, I guess

src/ecalc/libraries/libecalc/common/libecalc/common/time_utils.py
TeeeJay: heh...good question...without tests its harder to say ...it was a bit hard to read the diff...as usual :p..but yes...you are right, then we agree :)

TeeeJay (approved):

PR #170, opened 2023-09-13
fix: component type set to component level in asset dto

frodehk:

componentType is set to component level in asset_result_dto (get_asset_result). Should be set to component type.

Update componentType in asset_result_dto.

https://equinor-ecalc.atlassian.net/browse/ECALC-179


jsolaas (approved):
Nice

PR #173, opened 2023-09-14
refactor: improve error message when wrong CURVE-keyword input to single speed compressor

frodehk:

Keyword `CURVES` is reserved for _variable_ speed compressors only. _Single_ speed compressors should use the keyword `CURVE`. Due to the obvious similarity it is a risk that the user mixes the two, e.g. using `CURVES` for single speed compressors. When this happens the error message is difficult to understand.

- Improve error message when `CURVES` is used for single speed compressor models
- Improve error message when the keyword CURVE is not found (typo etc.) in general, for single speed compressor models.

https://equinor-ecalc.atlassian.net/browse/ECALC-20?atlOrigin=eyJpIjoiOTBjOWU5M2VjOGQwNDdmOWFkOWIxYjlhM2ExYThmNTQiLCJwIjoiaiJ9

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/input/mappers/model.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/input/mappers/model.py
@@ -66,12 +66,31 @@ def _get_curve_data_from_resource(resource: Resource, speed: float = 0.0):
 def _single_speed_compressor_chart_mapper(model_config: Dict, resources: Resources) -> dto.SingleSpeedChart:
     units = get_units_from_chart_config(chart_config=model_config)
     curve_config: Optional[Dict] = model_config.get(EcalcYamlKeywords.consumer_chart_curve)
+    name = model_config.get(EcalcYamlKeywords.name)
+    curves_instead_of_curve = (
+        True if isinstance(model_config.get(EcalcYamlKeywords.consumer_chart_curves), dict) else False
+    )
+
+    # Check if user has used CURVES (reserved for variable speed compressors)
+    # instead of CURVE (should be used for single speed compressors),
+    # and give clear error message.
+    if curves_instead_of_curve:
+        raise DataValidationError(
+            data=model_config,
+            message=f"Compressor model {name}:\n"
+            f"The keyword {EcalcYamlKeywords.consumer_chart_curves} should only be used for "
+            f"variable speed compressor models.\n"
+            f"{name} is a single speed compressor model and should use the keyword "
+            f"{EcalcYamlKeywords.consumer_chart_curve}.",
+        )
 
     if not isinstance(curve_config, dict):
         raise DataValidationError(
             data=model_config,
-            message=f"{EcalcYamlKeywords.consumer_chart_curve}"
-            f" should be an object. Type given: {type(curve_config)}.",
+            message=f"The keyword {EcalcYamlKeywords.consumer_chart_curve} is not specified "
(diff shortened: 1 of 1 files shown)


src/ecalc/libraries/libecalc/common/libecalc/input/mappers/model.py
```diff
@@ -66,12 +66,31 @@ def _get_curve_data_from_resource(resource: Resource, speed: float = 0.0):
...
+            f"{name} is a single speed compressor model and should use the keyword "
+            f"{EcalcYamlKeywords.consumer_chart_curve}.",
+        )
 
     if not isinstance(curve_config, dict):
         raise DataValidationError(
             data=model_config,
-            message=f"{EcalcYamlKeywords.consumer_chart_curve}"
-            f" should be an object. Type given: {type(curve_config)}.",
+            message=f"The keyword {EcalcYamlKeywords.consumer_chart_curve} is not specified "
```
jsolaas, line 90: The check `if EcalcYamlKeywords.consumer_chart_curve not in model_config:` would make more sense here, we could move it up to the start of the function, then remove Optional on the curve_config type.

src/ecalc/libraries/libecalc/common/libecalc/input/mappers/model.py
```diff
@@ -66,12 +66,31 @@ def _get_curve_data_from_resource(resource: Resource, speed: float = 0.0):
...
     curve_config: Optional[Dict] = model_config.get(EcalcYamlKeywords.consumer_chart_curve)
+    name = model_config.get(EcalcYamlKeywords.name)
+    curves_instead_of_curve = (
+        True if isinstance(model_config.get(EcalcYamlKeywords.consumer_chart_curves), dict) else False
+    )
+
+    # Check if user has used CURVES (reserved for variable speed compressors)
+    # instead of CURVE (should be used for single speed compressors),
+    # and give clear error message.
+    if curves_instead_of_curve:
```
jsolaas, line 77:
```suggestion
   if EcalcYamlKeywords.consumer_chart_curves in model_config:
```

jsolaas (approved):
Looks good. It would be nice if we could figure out a way to do this in the pydantic models, but it also makes sense to do it in the yaml models (not yet used/created)

src/ecalc/libraries/libecalc/common/libecalc/input/mappers/model.py
frodehk: If the check `if isinstance(curve_config, dict):` is replaced with `if EcalcYamlKeywords.consumer_chart_curve not in model config:`, mypy fails with the following error:

> Unsupported right operand type for in ("Any | None")

frodehk: Seems like it is an issue with mypy (it doesn´t apply type narrowing within nested functions). Adding `# type: ignore` after the next check (`if EcalcYamlKeywords.file in curve_config:`) solves it...

jsolaas: Does it help to add a check for `if not isinstance(model_config, dict)` then raise a relevant error? Not sure why mypy complains when we have specified dict as type for model_config

frodehk: Yes it helps.

PR #174, opened 2023-09-14
refactor: use yaml prefix for yaml klasses/modules

jsolaas:
- add yaml prefix to modules
- add yaml prefix to classes
- add title in config


TeeeJay (approved):
It is as we discussed. Let's try and see if this makes it easier to use/find and improve automation. Next step would be to change the namespace... `libecalc.input.yaml` perhaps (didnt think properly yet). Might be "smør på flesk", but having the same names of files and modules all over the place makes it so hard to find what you are looking for, or knowing which context you are in...

jsolaas:
> It is as we discussed. Let's try and see if this makes it easier to use/find and improve automation. Next step would be to change the namespace... `libecalc.input.yaml` perhaps (didnt think properly yet). Might be "smør på flesk", but having the same names of files and modules all over the place makes it so hard to find what you are looking for, or knowing which context you are in...

Depends how you look! But this is at least very clear when you see a class used, we might catch some "Why is this Yaml-class used here?". 

How about libecalc.yaml instead?

PR #175, opened 2023-09-14
refactor: improve error message when wrong CURVES-keyword input to variable speed compressor

frodehk:

Keyword `CURVE` is reserved for _single_ speed compressors only. _Variable_ speed compressors should use the keyword `CURVES`. Due to the obvious similarity it is a risk that the user mixes the two, e.g. using `CURVE` for variable speed compressors. When this happens the error message is difficult to understand.

- Improve error message when `CURVE` is used for variable speed compressor models
- Improve error message when the keyword `CURVES` is not found (typo etc.) in general, for variable speed compressor models.

https://equinor-ecalc.atlassian.net/browse/ECALC-189?atlOrigin=eyJpIjoiZmMyMzkwNzAxNDAwNDdhMzk1NDM4ZDRmNTAyMmU1NTgiLCJwIjoiaiJ9

### DIFF
--- a/src/ecalc/libraries/libecalc/common/libecalc/input/mappers/model.py
+++ b/src/ecalc/libraries/libecalc/common/libecalc/input/mappers/model.py
@@ -117,8 +117,8 @@ def _variable_speed_compressor_chart_mapper(model_config: Dict, resources: Resou
 
     name = model_config.get(EcalcYamlKeywords.name)
 
-    # Check if user has used CURVES (reserved for variable speed compressors)
-    # instead of CURVE (should be used for single speed compressors),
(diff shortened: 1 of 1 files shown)


TeeeJay:
General comment. It is confusing for users when we have `CURVE` and `CURVES` used in different places. We solved with with `CONDITION` and `CONDITIONS` to allow both, I think. Should we allow both, both places? It is of course very confusing if `CURVE` would indicate a list of curves... We could alternatively discuss with user/users whether it is worth being consistent and change all to `CURVES` and hence have a breaking change...

jsolaas (approved):

PR #185, opened 2023-09-16
chore: add rate type to results where relevant

frodehk:

Missing rate type (stream or calendar) some places:

- compressor stage
- Inlet gas rate on compressor trains
- recirculation loss
- inlet actual rate on compressor stages

Add rate type (`rate_type_all_rates_incl_stages`) to `CompressorModelResult` in `get_asset_result` (`graph_result`). Instead of creating new time series for all different rates, the rate type is set one place and is valid for the whole compressor model and subsequent stages.

https://equinor-ecalc.atlassian.net/browse/ECALC-91?atlOrigin=eyJpIjoiNmU5Y2I5OGIyNjJjNDhmNWIwMTY3NTlhZmMyYmIzMTgiLCJwIjoiaiJ9


src/ecalc/libraries/libecalc/common/libecalc/dto/result/results.py
```diff
@@ -138,6 +139,7 @@ class CompressorModelResult(ConsumerModelResultBase, CompressorTrainResult):
     componentType: Literal[ComponentType.COMPRESSOR]
     requested_inlet_pressure: TimeSeriesFloat
     requested_outlet_pressure: TimeSeriesFloat
+    rate_type_all_rates_incl_stages: RateType
```
jsolaas, line 142: I think we should remove the CompressorTrainResult and use TimeSeries for those attributes instead. It makes it easier for consumers of the api (webapp) if we have a consistent way of representing the arrays/timeseries/results/values.

jsolaas (commented):
I would prefer using TimeSeries, in addition to being consistent for consumers, it also makes it easier if we want to include model results in csv, or generate a separate csv.

frodehk:
Trouble with rebasing. Creating new branch.

PR #186, opened 2023-09-18
chore: remove unnecessary folders

jsolaas:

The folder structure is something left-over from back when we had namespaced packages. This cleans it up a bit.

Rename/move files, the import path does not change.


TeeeJay (approved):
I cannot check everything, but assume that you are right that this doesnt change namespace/package, just makes the structure more compact than it was due to historical reasons :)

PR #187, opened 2023-09-18
chore: add rate type to compressor model results and convert to time series

frodehk:

Missing rate type (stream or calendar) some places for compressor models:

- Compressor stage
- Inlet gas rate on compressor trains
- Recirculation loss
- Inlet actual rate on compressor stages

- [x] Add rate type for compressor models in `get_asset_result()`
- [x] Add new classes for compressor stages, stream conditions and turbine result in dto results
- [x] Convert compressor model variables from list to `TimeSeriesRate`, `TimeSeriesFloat` or `TimeSeriesBoolean`
- [x] Update snapshots

https://equinor-ecalc.atlassian.net/browse/ECALC-91?atlOrigin=eyJpIjoiNmU5Y2I5OGIyNjJjNDhmNWIwMTY3NTlhZmMyYmIzMTgiLCJwIjoiaiJ9


src/ecalc/libraries/libecalc/common/libecalc/core/models/results/compressor.py
```diff
@@ -121,7 +122,7 @@ def create_empty(cls, number_of_timesteps: int) -> CompressorStageResult:
 class CompressorTrainResult(EnergyFunctionResult):
     """The compressor train result component."""
 
-    rate_sm3_day: Union[List[Optional[float]], List[List[Optional[float]]]]
+    rate_sm3_day: Union[List[Optional[float]], List[List[Optional[float]]], TimeSeriesRate]
```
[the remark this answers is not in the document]
frodehk: Done

src/ecalc/libraries/libecalc/common/libecalc/dto/result/results.py
```diff
@@ -134,10 +139,17 @@ class PumpModelResult(ConsumerModelResultBase):
...
 
 
-class CompressorModelResult(ConsumerModelResultBase, CompressorTrainResult):
+class CompressorModelResult(ConsumerModelResultBase):
     componentType: Literal[ComponentType.COMPRESSOR]
+    # max_standard_rate: Optional[Union[List[Optional[float]], List[List[Optional[float]]]]]
+    failure_status: List[Optional[CompressorTrainCommonShaftFailureStatus]]
     requested_inlet_pressure: TimeSeriesFloat
     requested_outlet_pressure: TimeSeriesFloat
+    rate_sm3_day: TimeSeriesRate
```
TeeeJay, line 148: I guess we need to name the attribute `rate_sm3_day` wrt API or legacy name? But with e.g. TimeSeriesRate, then information about unit is in that object, and we should therefore not need to provide units etc in the name any more...

```suggestion
    rate: TimeSeriesRate
```

? rate or `gas_rate` ? However, if possible, we should probably change this all over the place in one go, instead of here and there?

src/ecalc/libraries/libecalc/common/libecalc/dto/result/results.py
```diff
@@ -134,10 +141,25 @@ class PumpModelResult(ConsumerModelResultBase):
...
+    stage_results: List[CompressorStageResult]
+    turbine_result: Optional[TurbineResult] = None
+    energy_usage_unit: Unit
+    power_unit: Unit
+
+    def to_time_series(self, stage_result: list, unit: Unit) -> TimeSeriesRate:
```
frodehk, line 156: @jsolaas @TeeeJay : will it help to add this function? Less repeated code in graph_result.

TeeeJay: In general it is a good idea, and IMO and aligned with "core should know less about details, just numbers", but the problem here is that core depends on an outer layer, so to speak. It should be other way around. Difficult to say, because it seems that some of these classes are "both core and not core". So, good idea, and we will eventually do something like that, but I do not like the direction of the dependency. However, "damage" has already been done, as this direction of dependency is already there... I think we should solve the dependency issue first..

frodehk: Ok, we wait - and do everything in get_asset_result, like suggested?

src/ecalc/libraries/libecalc/common/libecalc/dto/result/results.py
frodehk: Have changed to rate in the dto, however, the dto get values from core (CompressorTrainResult) where it is still rate_sm3_day.

src/libecalc/core/graph_result.py
```diff
@@ -295,6 +295,104 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
...
 
+                    rate = TimeSeriesRate(
+                        timesteps=model.timesteps,
+                        values=model.rate_sm3_day,
+                        unit=Unit.STANDARD_CUBIC_METER_PER_DAY,
+                        rate_type=RateType.STREAM_DAY,
+                    )
+
+                    # Convert rates in stage results from lists to time series:
+                    for stage_result in model.stage_results:
```
jsolaas, line 306: Instead of manipulating the objects here, we should create a new ones. This way we don't end up with invalid data in model.stage_results, while also getting validation that the new objects are correct (pydantic validation).

When manipulating values pydantic does not validate, so here we get a situation where the type in the stage_result class says List[float] (or something) while the type has been altered to TimeSeriesRate.

jsolaas (approved):
Looks good! We should keep going with TurbineResult, and maybe look into a common FailureStatus object that could be moved to libecalc.common while we restructure. Me, @TeeeJay and @olelod had a discussion about failure status, so we should align before doing anything with that.

TeeeJay (approved):
Good effort :)

frodehk:
Forgot TurbineResult, will add it to the CompressorModelResult. Should we create a separate class for TurbineResult in dto.results as well, and convert to time series?

jsolaas:
> `TurbineModelResult` is now added as a new dto.result-class, and used in `graph_result` to add the turbine result to `CompressorModelResult`. @jsolaas and @TeeeJay: When time, see if it is in line with what we want.

Looks good!

PR #190, opened 2023-09-18
Ecalc 167 actual rate in compressor sampled

olelod:

The input rate (in Sm3) has been returned as the actual rate (in Am3) for compressor trains of type compressor_sampled. This is obviously not correct, since we need information about fluid density to calculate the actual rate.


src/ecalc/libraries/libecalc/common/libecalc/core/models/compressor/sampled/compressor_model_sampled.py
```diff
@@ -226,7 +226,6 @@ def evaluate_rate_ps_pd(
                     power_unit=Unit.MEGA_WATT,
                     inlet_stream_condition=CompressorStreamCondition(
                         pressure=list(suction_pressure) if suction_pressure is not None else None,
-                        actual_rate_m3_per_hr=list(rate) if rate is not None else None,
```
TeeeJay, line 229: Add a comment here or in `CompressorStreamCondition` to specify what you mentioned in the PR comment that knowing `fluid_density` is a requirement to set this, to avoid doing the same mistake in the future, and to bring awareness?

TeeeJay (approved):
Good fix :) Remember to use squash to change commit-message to `fix:`

jsolaas (approved):

PR #194, opened 2023-09-21
feat: add stream conditions to compressor v2

jsolaas:
Also expanded merge to handle recursive objects and lists.


src/tests/libecalc/integration/snapshots/test_consumer_system_v2/test_compressor_system_v2_results/consumer_system_v2-consumer_system_v20/consumer_system_v2.json
```diff
@@ -3916,6 +3916,68 @@
...
+                        ],
+                        "unit": "bara",
+                        "values": [
+                            125.0,
+                            125.0,
+                            125.0,
+                            125.0
+                        ]
+                    },
+                    "stream_day_rates": [
```
olelod, line 3950: Is the reason for calling it stream_day_rates here that it should always be in stream days? I mean it is a TimeSeriesRate, and there is a rate_type inside. Just curious.

TeeeJay (approved):

TeeeJay (approved):
ok :)

src/tests/libecalc/integration/snapshots/test_consumer_system_v2/test_compressor_system_v2_results/consumer_system_v2-consumer_system_v20/consumer_system_v2.json
jsolaas: Yes, that was the reason. I think that name came before we used TimeSeriesRate, but I'm not sure. Seems unnecessary now :)

src/libecalc/core/consumers/compressor/component.py
```diff
@@ -132,13 +151,30 @@ def evaluate(
...
+            outlet_pressure_before_choking=outlet_pressure_before_choke,
+            stages=[
+                Stage(
+                    name="inlet",
+                    stream_condition=StreamCondition(
+                        rate=total_requested_rate,
+                        pressure=operational_settings.inlet_pressure,
+                    ),
+                ),
+                Stage(
```
jsolaas, line 163: Should be easy to extend this list with more stages from models.

src/libecalc/common/stream.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class Stage(BaseModel):
+    class Config:
+        extra = Extra.forbid
+        alias_generator = to_camel_case
+        allow_population_by_field_name = True
+
+    name: Literal["inlet", "before_choke", "outlet"]
+    stream_condition: StreamCondition
```
jsolaas, line 26: Let me know if we have something similar already in models

src/libecalc/common/stream.py
```diff
@@ -0,0 +1,26 @@
+from typing import Literal
+
+from libecalc.common.string_utils import to_camel_case
+from libecalc.common.utils.rates import TimeSeriesFloat, TimeSeriesRate
+from pydantic import BaseModel, Extra
+
+
+class StreamCondition(BaseModel):
```
TeeeJay, line 8: I guess usage here is in consumer system (only), but the entity should be called `Stream`, and it is the `Stream` that should be used in consumer system imo

src/libecalc/common/stream.py
```diff
@@ -0,0 +1,26 @@
...
+    class Config:
+        extra = Extra.forbid
+        alias_generator = to_camel_case
+        allow_population_by_field_name = True
+
+    rate: TimeSeriesRate
+    pressure: TimeSeriesFloat
+    fluid_density: TimeSeriesFloat = None
```
TeeeJay, line 16: Is this for pump only? Or both compressor and pump? May easily lead to confusion if there is a difference. Not sure if it makes sense with a `Fluid` and `FluidProperties` for pump, as it is simpler..but may be that it is the same...or if we should call it `Fluid` wrt compressors since it is "mostly gas"...to be discussed

src/libecalc/common/stream.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class Stage(BaseModel):
```
TeeeJay, line 19: Name easily confused with `Compressor(Train)Stage`, but somehow related (I guess). To be discussed. Not sure what is the name being used to define the stage/phase/step/ that the stream goes through...in the system...

src/libecalc/core/consumers/compressor/component.py
TeeeJay: I like the start to more structured add the different stage/phases etc of the stream that you are adding though :)

src/libecalc/core/result/results.py
```diff
@@ -4,6 +4,7 @@
 from datetime import datetime
 from typing import Any, Dict, List, Optional, Union
 
+from libecalc.common.stream import Stage
```
TeeeJay, line 7: This should go to `core` (now) or `domain` (later?) project package though, imo. But we will need a representation in the `dto`/application layer as well wrt input/output I guess...

src/libecalc/common/stream.py
jsolaas: I guess we could have different types of streams, where pump expects a certain type, and compressor another. But I'm hoping we can find a common interface, where the stream itself can handle density etc. That means using streams as input though, so not yet. 

But I agree, to be discussed.

src/libecalc/common/stream.py
jsolaas: CompressorStage, ASVStage, ChokeStage. I was thinking the common name for all of these were stage, and that is what this represents, so stream is only one of the properties on a stage, might be the only one.. we'll see

src/libecalc/core/result/results.py
jsolaas: Yes, core/domain. Hopefully dto will be allowed to use it from core. It isn't all that clear to me anymore why we need a dto layer..We have a yaml layer, and can convert to something else there, why do we need a layer between yaml and core? I'm sure there is a reason, and I'm thinking it has something to do with evaluated vs not evaluated expressions..

It was introduced to allow changes, but in v2 we can make changes in core. I dunno

src/libecalc/common/stream.py
jsolaas: Only consumer system now, but I'm hoping/expecting this to be used for all new components, also in models if we want to introduce TimeSeries there..

It's used in compressor (and when this is approved I will add it to pump) so when we add a pump and compressor v2 (outside system) it will be used there also

src/libecalc/core/result/results.py
TeeeJay: Let's discuss .)

PR #196, opened 2023-09-21
chore: warn user about full recirculation of fluids in a compressor s…

olelod:
…tage in a multiple streams and pressures compressor train


src/libecalc/core/models/compressor/train/variable_speed_compressor_train_common_shaft_multiple_streams_and_pressures.py
```diff
@@ -872,6 +872,11 @@ def calculate_compressor_train_given_rate_ps_speed(
                     stage_number=stage_number
                 )
                 if fluid_to_recirculate:
+                    logger.warning(
+                        f"For stage number {stage_number}, there is no fluid entering the stage at at this time step. "
```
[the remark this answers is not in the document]
olelod: I agree, but I also think this is a more general problem that occurs many of the times we use `logger.warning`. This is at least not the only place. Should we address it in a more general setting?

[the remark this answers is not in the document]
TeeeJay: You are right...it is happening many times already, and we should instead find a better solution in a general setting. Later. (Unless you see and find a good solution for this and others now)

src/libecalc/core/models/compressor/results.py
```diff
@@ -290,15 +290,21 @@ def set_failure_status(cls, v, values):
 
     @property
     def chart_area_status(self) -> ChartAreaFlag:
-        chart_area_failure = [
-            stage.chart_area_flag
-            for stage in self.stage_results
-            if stage.chart_area_flag != ChartAreaFlag.INTERNAL_POINT
-        ]
+        """Checks where the operational points are placed in relation to the compressor charts in a compressor train.
```
TeeeJay, line 293: Set up docstring in PyCharm to use "Google" in "File > Settings > Tools > Python Integrated Tools > Docstring format" --> Google :)

src/libecalc/core/models/compressor/train/chart/variable_speed_compressor_chart.py
```diff
@@ -317,24 +317,25 @@ def _evaluate_point_validity_chart_area_flag_and_adjusted_speed_and_rate(
...
-                    point_is_valid = False
             elif rate > maximum_flow_rate_for_speed:
                 chart_area_flag = (
                     ChartAreaFlag.BELOW_MINIMUM_SPEED_AND_ABOVE_MAXIMUM_FLOW_RATE
                     if speed_is_increased
                     else ChartAreaFlag.ABOVE_MAXIMUM_FLOW_RATE
                 )
                 point_is_valid = False
-
+            if increase_rate_left_of_minimum_flow_assuming_asv:
```
TeeeJay, line 335: Is it on purpose that this moved outside the `elif`? One intendation out?

src/libecalc/core/models/compressor/train/chart/variable_speed_compressor_chart.py
```diff
@@ -317,24 +317,25 @@ def _evaluate_point_validity_chart_area_flag_and_adjusted_speed_and_rate(
...
+                rate_to_use = max(minimum_flow_rate_for_speed, rate_to_use)
```
TeeeJay, line 336: is `point_is_valid` = `False` here as well?

src/libecalc/core/models/compressor/train/chart/variable_speed_compressor_chart.py
olelod: Yes and no. Updating the rate to use can be seen as independent from finding the placement of the rate in the input relative to the compressor chart. Done some small updates now. Maybe more clear...?

src/libecalc/core/models/compressor/train/chart/variable_speed_compressor_chart.py
olelod: If `increase_rate_left_of_minimum_flow_assuming_asv == True `points will be move to minimum flow and considered to be valid.

src/libecalc/core/models/compressor/train/chart/variable_speed_compressor_chart.py
```diff
@@ -317,16 +325,21 @@ def _evaluate_point_validity_chart_area_flag_and_adjusted_speed_and_rate(
             # Todo: Need to QA this logic. Below minimum speed requires speed < minimum_speed, but this is not enforced?
             minimum_flow_rate_for_speed = self.minimum_rate_as_function_of_speed(speed_to_use)
             maximum_flow_rate_for_speed = self.maximum_rate_as_function_of_speed(speed_to_use)
-            if rate < minimum_flow_rate_for_speed:
+            # first update rate_to_use when below minimum flow if increase_rate_left_of_minimum_flow_assuming_asv
+            if rate_to_use < minimum_flow_rate_for_speed:
+                if increase_rate_left_of_minimum_flow_assuming_asv:
+                    rate_to_use = max(minimum_flow_rate_for_speed, rate_to_use)
```
TeeeJay, line 331: you already know that `rate_to_use < minimum_flow_rate_for_speed` ? Then just take `minimum_flow_rate_for_speed`?

TeeeJay (approved):
See comment. I think it is ok.

Do we have a test that catches these 2 scenarios somehow?

src/libecalc/core/models/compressor/train/chart/variable_speed_compressor_chart.py
olelod: 💡

olelod:
> See comment. I think it is ok.
> 
> Do we have a test that catches these 2 scenarios somehow?

Made a test for full recirculation in an earlier PR. Added check of the ChartAreaFlag also in that test now.

TeeeJay:
Remember to update changelog in docs as well (v8.5)

PR #200, opened 2023-09-25
refactor: remove rate_fractions from system v2

jsolaas:
Rate fractions seems to cause some confusion, and can be achieved with variables.


TeeeJay (approved):
sounds like a good improvement :)

PR #204, opened 2023-09-25
refactor: move crossover to component_conditions for system v2

jsolaas:
Only allow a single crossover, not change it over time.

Make it more clear that operational settings only contain stream conditions for consumers. Rename coming.

### DIFF
--- a/src/libecalc/core/consumers/consumer_system.py
+++ b/src/libecalc/core/consumers/consumer_system.py
@@ -42,9 +42,11 @@ ConsumerOperationalSettings = TypeVar(
 )
 
 
-class SystemOperationalSettings(Protocol):
+class SystemComponentConditions(Protocol):
     crossover: List[int]
 
+
+class SystemOperationalSettings(Protocol):
     @abstractmethod
     def get_consumer_operational_settings(
         self,
@@ -64,14 +66,14 @@ class ConsumerSystem(BaseConsumer):
     for a period of time -> Turn of compressor #2, and vice versa.
     """
 
-    def __init__(self, id: str, consumers: List[ConsumerComponent]):
+    def __init__(self, id: str, consumers: List[ConsumerComponent], component_conditions: SystemComponentConditions):
         self.id = id
         self._consumers = [create_consumer(consumer) for consumer in consumers]
+        self._component_conditions = component_conditions
 
     def _get_operational_settings_adjusted_for_crossover(
         self,
         operational_setting: Union[
-            SystemOperationalSettings,
             EvaluatedPumpSystemOperationalSettings,
             EvaluatedCompressorSystemOperationalSettings,
         ],
@@ -82,7 +84,7 @@ class ConsumerSystem(BaseConsumer):
         consumers.
         """
         sorted_consumers = ConsumerSystem._topologically_sort_consumers_by_crossover(
-            crossover=operational_setting.crossover,
+            crossover=self._component_conditions.crossover,
             consumers=self._consumers,
         )
         adjusted_operational_settings = []
@@ -92,7 +94,7 @@ class ConsumerSystem(BaseConsumer):
         }
 
         # Converting from index 1 to index 0.
-        crossover = [crossover_flow_to_index - 1 for crossover_flow_to_index in operational_setting.crossover]
+        crossover = [crossover_flow_to_index - 1 for crossover_flow_to_index in self._component_conditions.crossover]
 
         for consumer in sorted_consumers:
             consumer_index = self._consumers.index(consumer)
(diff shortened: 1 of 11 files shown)


src/libecalc/fixtures/cases/consumer_system_v2/data/consumer_system_v2.yaml
```diff
@@ -102,19 +102,18 @@ INSTALLATIONS:
               COMPRESSOR_MODEL: compressor_sampled_1d
             - NAME: compressor3
               COMPRESSOR_MODEL: compressor_sampled_1d
+          COMPONENT_CONDITIONS:
```
jsolaas, line 105: COMPONENT_PROPERTIES? Or use more specific naming? I'm thinking this can contain CONDITION (rename to ENABLED or something?), if we add a start and end we could have that here. TOTAL_SYSTEM_RATE could be here for system, but it (still) seems unnecessary to me.

src/libecalc/fixtures/cases/consumer_system_v2/data/consumer_system_v2.yaml
```diff
@@ -125,16 +124,15 @@ INSTALLATIONS:
             ENERGY_USAGE_MODEL: compressor_sampled_1d
           - NAME: compressor3
             ENERGY_USAGE_MODEL: compressor_sampled_1d
+        COMPONENT_CONDITIONS:
+          CROSSOVER: [ 0, 1, 1 ]
         OPERATIONAL_SETTINGS:
-          - RATES: [ 1000000, 5000000, 5000000 ]
+          - RATES: [ 1000000, 6000000, 6000000 ]
```
jsolaas, line 130: We should look into restructuring this, do the users prefer thinking in indices or should we just use a map where the key is the compressor name? 

```yaml
          CONSUMER_STREAM_CONDITIONS:
            - consumer1:
                STREAMS:
                  - RATE: 1000000
                    INLET_PRESSURE: 50
                    OUTLET_PRESSURE: 250
              consumer2:
                STREAMS:
                  - RATE: 6000000
                    INLET_PRESSURE: 50
                    OUTLET_PRESSURE: 250
              consumer3:
                STREAMS:
                  - RATE: 6000000
                    INLET_PRESSURE: 50
                    OUTLET_PRESSURE: 250
```

if we should support multiple streams, if not

```yaml
          CONSUMER_STREAM_CONDITIONS:
            - consumer1:
                STREAM:
                  RATE: 1000000
                  INLET_PRESSURE: 50
                  OUTLET_PRESSURE: 250
              consumer2:
                STREAM:
                  RATE: 6000000
                  INLET_PRESSURE: 50
                  OUTLET_PRESSURE: 250
              consumer3:
                STREAM:
                  RATE: 6000000
                  INLET_PRESSURE: 50
                  OUTLET_PRESSURE: 250
```

src/libecalc/core/consumers/consumer_system.py
```diff
@@ -42,9 +42,11 @@
 )
 
 
-class SystemOperationalSettings(Protocol):
+class SystemComponentConditions(Protocol):
```
TeeeJay, line 45: If this is more or less static, it will not be a condition. In any case, name to be decided imo. Something we need to know.

Could also be possibly improved in yaml to make it easier for user to see which crossover is open and which is closed with more "verbose" and explicit yaml, if it is more or less static, then if a crossover opens or closes, that is given as an "event" instead of repeating the "final state" for all every time...

However, I didnt check usage in practice, whether it is normal to change this depending on priority...

src/libecalc/fixtures/cases/consumer_system_v2/data/consumer_system_v2.yaml
TeeeJay: I think something like this is better, since it does not give the cognitiv overload of making sure that the index that you add sth to is correct, pretty sure it is easy to do a mistake...

however, quite much more verbose when writing yaml...so there is a trade off... also topic for discussion..

src/libecalc/fixtures/cases/consumer_system_v2/data/consumer_system_v2.yaml
```diff
@@ -125,16 +125,15 @@ INSTALLATIONS:
             ENERGY_USAGE_MODEL: compressor_sampled_1d
           - NAME: compressor3
             ENERGY_USAGE_MODEL: compressor_sampled_1d
+        COMPONENT_CONDITIONS:
```
TeeeJay, line 128: topic for discussion of course. I will check usage a few places as well.

TeeeJay (approved):

TeeeJay (approved):
good initiative. topic for discussion :)

PR #210, opened 2023-09-26
Feat/add stages to pumpv2

jsolaas:
Add stages to pump
Add pump result when calculating pump system


src/libecalc/core/consumers/compressor/component.py
```diff
@@ -72,12 +72,20 @@ def evaluate(
...
+            rate_type=operational_settings.stream_day_rates[0].rate_type,
+        )
+
         model_results = []
         evaluated_timesteps = []
 
-        # TODO: This is a false assumption and will be dealt with shortly (that the regularity is the same
-        #   for all timesteps, and only taken for the first timestep). Not the first timestep, the first rate
-        evaluated_regularity = operational_settings.stream_day_rates[0].regularity
+        evaluated_regularity = total_requested_rate.regularity
```
TeeeJay, line 88: so the one above to restrict to one stream is to 1. handle regularity correctly and 2. fix multiple streams problem later? (to be discussed?)

TeeeJay (approved):

src/libecalc/core/consumers/compressor/component.py
jsolaas: It's the same we have been doing. I moved the total_requested_rate to the top to make it a bit more clear, and reusable. So we don't have several places where we sum the rates. Regularity is kind of weird, because we are relying on it being set on installation, i.e. the same for all vectors. But that is a TODO. 

And yes, I have postponed dealing with multiple streams here, which will not work correctly with a single rate. To be discussed

src/libecalc/core/consumers/compressor/component.py
```diff
@@ -72,12 +72,20 @@ def evaluate(
         """
         self._operational_settings = operational_settings
 
+        # Creating a single input rate until we decide how to deal with multiple rates, multiple rates added because of
```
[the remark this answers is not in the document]
jsolaas: Well, I'm just aligning pump and compressor. So it depends on how strict we are with separation, it's fine in this PR IMO. It's just a refactor.

PR #213, opened 2023-09-28
fix: set_regularity fixture

jsolaas:
Set regularity should not raise error when value is wrong, and I think we need to raise ValueError for pydantic to pick it up correctly, if we wanted to raise an error.


TeeeJay (approved):
remember cherry-pick to patch AND update web patch/release branch...

PR #214, opened 2023-09-28
fix: rate when multiple streams model

jsolaas:
When multiple streams are used we sum the rates to fit the expected result.


TeeeJay (approved):
ok, change to first instead of sum

PR #215, opened 2023-09-28
chore: handle requested pressures for compressor systems

frodehk:

Extracting requested inlet- and outlet pressure does not work correctly for compressor systems:
When using `SUCTION_PRESSURES` or `DISCHARGE_PRESSURES` there are one pressure per component in the system. This list of pressures is not captured and the pressures are set to nan, and then nan to num (0).

- Update `get_requested_compressor_pressures` in `graph_result`, to select correct pressures for compressors in system
- Ensure correct operational setting is used for the compressors in the system

https://equinor-ecalc.atlassian.net/browse/ECALC-252?atlOrigin=eyJpIjoiYTJiZGUwNmY5MjlhNGMwMjg5YTc1Y2FiNTY0ZTkxNDQiLCJwIjoiaiJ9

### DIFF
--- a/src/libecalc/core/graph_result.py
+++ b/src/libecalc/core/graph_result.py
@@ -227,24 +227,41 @@ class GraphResult:
         :param energy_usage_model: Temporal energy model
         :param pressure_type: Compressor pressure type, inlet- or outlet
         :param name: name of compressor
+        :param operational_setting_index: if multiple operational settings for one model
         :return: Temporal model with pressures as expressions
         """
+
         evaluated_temporal_energy_usage_models = {}
+        operational_setting_index = 0
+
         for period, model in TemporalModel(energy_usage_model).items():
             if isinstance(model, CompressorSystemConsumerFunction):
-                for compressor, operational_setting in zip(model.compressors, model.operational_settings):
+                for compressor in model.compressors:
                     if compressor.name == name:
-                        pressures = operational_setting.suction_pressure
+                        # Challenge: do not handle variation in operational settings, just select first priority.
+                        # Possible to vary inlet- and outlet pressures between different operational settings
+                        # priorities, within one temporal model,
+                        operational_setting = model.operational_settings[operational_setting_index]
+                        compressor_nr = int(
+                            [i for i, compressor in enumerate(model.compressors) if compressor.name == name][0]
+                        )
 
-                        if pressure_type.value == CompressorPressureType.OUTLET_PRESSURE:
-                            pressures = operational_setting.discharge_pressure
+                        if pressure_type.value == CompressorPressureType.INLET_PRESSURE:
+                            if operational_setting.suction_pressures is not None:
+                                pressures = operational_setting.suction_pressures[compressor_nr]
+                            else:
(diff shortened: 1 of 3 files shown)


jsolaas (commented):
Can we drop this and say requested pressures are good enough until v2? Wouldn't this cause confusion? If I understand it correctly this is not necessarily correct?

TeeeJay:
Did you discuss this today? Should we just drop it for now? It just turns out to be "full of holes" because of all the different edge cases?

frodehk:
Discussed with @HeddaSvendsen, she reminded me of the `operational_settings_used`. The code is now updated, and the function should extract the correct requested pressures now - taking into account the relevant operational setting used, and picking the correct compressor in the system (if different pressures are specified within a temporal model)

src/libecalc/core/graph_result.py
```diff
@@ -215,50 +216,90 @@ def _parse_emissions(emissions: Dict[str, EmissionResult]) -> Dict[str, libecalc
             for key in emissions
         }
 
+    @staticmethod
+    def get_operational_setting_index(timestep: datetime, operational_settings_used: TimeSeriesInt):
```
TeeeJay, line 220:
```suggestion
    def get_operational_setting_index(timestep: datetime, operational_settings_used: TimeSeriesInt) -> int:
```

src/libecalc/core/graph_result.py
```diff
@@ -215,50 +216,90 @@ def _parse_emissions(emissions: Dict[str, EmissionResult]) -> Dict[str, libecalc
...
+        timestep_index = operational_settings_used.timesteps.index(timestep)
+
+        operational_setting_index = operational_settings_used.values[timestep_index]
```
TeeeJay, line 223: I would prefer to use a different name. This is easily confused with the index in the list. How about `operational_setting_used`? Or perhaps you have sth better?

TeeeJay: or Id?

frodehk: Perfect

frodehk: I like `operational_setting_used`, as it refers to one entry in the in the time series `operational_settings_used`, or?

TeeeJay:
This is getting overly complicated. Not your fault, it is just very difficult to get this good and right. I am afraid that we do not have enough overview of edge cases, where this will work and not etc, and whether the user can always trust the results? If we are to merge this, we should make sure that the code is isolated (doesnt break anything else), and possibly give a warning about this being  experimental (using Feature.experimental, in case it fails?) which also gives a warning....

Also, we do not have any other tests than the standard tests to test this.

What do you think?

frodehk:
I agree that it is complicated, and difficult to control edge cases - with few tests really checking. On the other hand, the code that is already merged in is trying to handle the requested pressures for systems in a wrong way - and this will not make it worse I think. Is it an idea to create som targeted tests for this, especially for operational settings in compressor systems?

TeeeJay:
So, also, we need to patch it then...

jsolaas (approved):
Looks good

PR #219, opened 2023-09-28
chore: add dependabot actions monitoring

jsolaas:


TeeeJay (approved):
Was "switched off" wrt vulnerabilities, and the github-actions part was forgotten (since we already have Snyk, and it was probably not needed). Snyk wont help us with gh actions though  :)

PR #224, opened 2023-09-29
refactor(libecalc.core): stream as input

jsolaas:


src/tests/libecalc/core/consumers/test_crossover.py
```diff
@@ -1,62 +1,96 @@
...
+            timesteps=timesteps,
+            values=[50] * len(rate),
+            unit=Unit.BARA,
+        ),
+    )
+
+
 class TestCrossover:
-    parameterized_crossover_rates = [
-        ([], [], [], []),  # All empty
```
jsolaas, line 8: No inlet streams should be handled elsewhere, right?

src/libecalc/core/consumers/compressor/component.py
```diff
@@ -49,8 +49,8 @@ def get_max_rate(self, operational_settings: CompressorOperationalSettings) -> L
             operational_settings_this_timestep = operational_settings.get_subset_for_timestep(timestep)
             results.extend(
                 compressor.get_max_standard_rate(
-                    suction_pressures=np.asarray(operational_settings_this_timestep.inlet_pressure.values),
-                    discharge_pressures=np.asarray(operational_settings_this_timestep.outlet_pressure.values),
+                    suction_pressures=np.asarray(operational_settings_this_timestep.inlet_streams[0].pressure.values),
```
jsolaas, line 52: Again, assuming first stream is the requested. We should probably make this a bit more clear. Maybe requested_inlet_stream should be a separate property than crossover. Need to look at how this works (or should work) in multiple streams also. Should we ask for max_rate per stream?

olelod: For multiple streams the max standard rate is calculated for each stream. When calculating the max standard rate for one stream, the rates for the other streams are kept constant.

src/libecalc/common/stream.py
```diff
@@ -15,6 +19,41 @@ class Config:
     pressure: TimeSeriesFloat
     fluid_density: TimeSeriesFloat = None
```
TeeeJay, line 20:
```suggestion
    fluid_density: Optional[TimeSeriesFloat] = None
```

src/libecalc/common/stream.py
```diff
@@ -15,6 +19,41 @@ class Config:
...
+            # TODO: return a warning object with the specific timesteps?
+            logger.warning("Increasing pressure when mixing stream. That should not happen.")
+        else:
+            return Stream(
+                rate=self.rate + otherStream.rate,
+                pressure=target_pressure,
+                fluid_density=self.fluid_density,  # Check that they are equal? Or handle it?
+            )
+
+    def get_subset_for_timestep(self, current_timestep: datetime) -> Self:
```
TeeeJay, line 44: docstrings?

src/libecalc/common/stream.py
```diff
@@ -15,6 +19,41 @@ class Config:
...
+        return Stream(
+            rate=self.rate.for_timestep(current_timestep),
+            pressure=self.pressure.for_timestep(current_timestep),
+            fluid_density=self.fluid_density.for_timestep(current_timestep) if self.fluid_density is not None else None,
+        )
+
+    @classmethod
+    def mix_all(cls, inlet_streams: List["Stream"]) -> "Stream":
+        if len(inlet_streams) == 0:
+            raise ValueError("No streams to mix")
```
TeeeJay, line 54: could warn and just return `self`?

src/libecalc/common/utils/rates.py
```diff
@@ -160,6 +160,12 @@ def convert_none_to_nan(cls, v: float, field: ModelField) -> TimeSeriesValue:
     def __len__(self) -> int:
         return len(self.values)
 
+    def __lt__(self, other) -> bool:
+        if not isinstance(other, TimeSeries):
+            return NotImplemented
```
TeeeJay, line 165: Should add a comment and reference why NotImplemented is used here instead of raising exception imo

src/libecalc/core/consumers/compressor/component.py
```diff
@@ -151,25 +145,19 @@ def evaluate(
...
+                        rate=total_requested_inlet_stream.rate,
                         pressure=outlet_pressure_before_choke,
                     ),
                 ),
                 Stage(
                     name="outlet",
                     stream=Stream(
-                        rate=total_requested_rate,
-                        pressure=operational_settings.outlet_pressure,
+                        rate=total_requested_inlet_stream.rate,  # Actual, not requested, different because of crossover
```
TeeeJay, line 159: supposed to be a TODO comment?

TeeeJay: and...inlet stream?

src/libecalc/core/consumers/pump/component.py
```diff
@@ -43,15 +44,17 @@ def get_max_rate(self, operational_settings: PumpOperationalSettings) -> List[fl
             operational_settings_this_timestep = operational_settings.get_subset_for_timestep(timestep)
             results.extend(
                 pump.get_max_standard_rate(
-                    suction_pressures=np.asarray(operational_settings_this_timestep.inlet_pressure.values),
-                    discharge_pressures=np.asarray(operational_settings_this_timestep.outlet_pressure.values),
-                    fluid_density=np.asarray(operational_settings_this_timestep.fluid_density.values),
+                    suction_pressures=np.asarray(operational_settings_this_timestep.inlet_streams[0].pressure.values),
```
TeeeJay, line 47: first or list of lists? Ref ole petters comment

src/tests/libecalc/core/consumers/test_crossover.py
TeeeJay: If so, then the docstring should say that that is assumed and not checked and handled in that method. Can it be handled? Probably not. ..

TeeeJay (approved):

TeeeJay (approved):
a few comments, see comment from ole petter wrt possibly changing logic?

src/libecalc/core/consumers/compressor/component.py
jsolaas: No todo, the total_inlet_stream is the actual stream being sent through the compressor, while the requested might be a lot less since we have crossover.

src/libecalc/core/consumers/pump/component.py
jsolaas: multistream isn't handled yet, first because that is the requested stream and where the target pressure is set.

PR #225, opened 2023-09-29
ECALC-19 - Check for zero efficiency

markusrf:
Added to give a more informative "division by zero" error when polytropic efficiency is 0.

### DIFF
--- a/src/ecalc_cli/commands/run.py
+++ b/src/ecalc_cli/commands/run.py
@@ -4,6 +4,12 @@ from pathlib import Path
 import libecalc.common.time_utils
 import libecalc.version
 import typer
+from libecalc.common.run_info import RunInfo
+from libecalc.core.ecalc import EnergyCalculator
+from libecalc.core.graph_result import GraphResult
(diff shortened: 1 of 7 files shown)


src/ecalc_cli/commands/run.py
```diff
@@ -4,6 +4,12 @@
 import libecalc.common.time_utils
 import libecalc.version
 import typer
+from libecalc.common.run_info import RunInfo
```
TeeeJay, line 7: resultat av pre-commit run -all regner jeg med?

TeeeJay (approved):

TeeeJay:
As mentioned, also check if this can be added when validating input. That would give user the warning/error immediately, instead of after calculations are done...

src/ecalc_cli/commands/run.py
markusrf: @TeeeJay  Yes, det ble litt endringer fra pre-commit pga den nye __init__ filen

PR #226, opened 2023-10-03
refactor: remove regularity out of core/domain

TeeeJay:

Regularity should not be transferred or handled in core. At the time data arrives in core, it should only be stream day. Therefore even the information that it is stream day and bringing in regularity adds up the complexity and risk of errors.

* Initial approach to make 2 fixed timeseries representations for fixed rate. The `TimeSeriesStreamDayRate` is the only one that should be used in core. However, it should most likely only be called `TimeSeriesRate`, as that is the real rate going through the system (real rate?). The other "rate" however, can have a more "report-friendly version" name, which is used to be able to convert to calendar day and do other calculations.
* Assuming that using rate as a float is ok in core, but that can be a bit dangerous if one adds functionality to Float, that should not work the same as Rate, e.g. adding 2 floats such as pressure, will yield the same,while adding Rate, will yield increased rate...(should we have a pressure type?...)
* 

## Remaining work

- [x] Make sure that the regularity lists correspond tot he length and indices of the timeseries it manipulates (in particular within models)
- [x] Verify results
- [ ] Add tests where regularity varies between different installations, to make sure correct regularity is used
- [x] Handle all special cases in application/dto layer instead of in domain/core? Such as default calendar rate etc.
- [x] Disallow to set a rate without explicitly set regularity or type (to avoid bugs, due to fallback to 1 and stream day)
- [x] TimeSeriesStreamDayRate should not know about TimeSeriesRate, just the other way around, switch that.
- [x] Handle legacy consumers, such as Direct. Handle regularity on the outside.
- [x] Verify that all conversion to stream day is done in preprocessing and correct
- [x] Verify that all convert to calendar day is done correctly in postprocessing
- [x] Check whether energy usage should be cal day or stream day in "intermediate" steps
- [x] Consider to restrict to only see stream day or calendar day in output for user (calendar day at high level, stream day at low level, with exceptions. consider for additional PR) (follow up PR)
- [x] Potentially clean up and make it clearer what is application (dto) and what is domain (core) in particular. (follow up PR)
- [x] Verify on init that self.values, self.timesteps (and self.regularity) have equal lengths? otherwise programmingerror...
- [x] Fix missing emissions
- [x] Fix missing consumer results (and failing corresponding test)
- [x] Verify length of timesteps and valuse (and regularity...that all correspond)...
- [x] LTP error, incorrect results ...


src/libecalc/common/utils/rates.py
```diff
@@ -624,6 +624,56 @@ def resample(self, freq: Frequency) -> TimeSeriesIntensity:
...
+        return TimeSeriesRate(
+            timesteps=self.timesteps,
+            values=self.values,
+            unit=self.unit,
+            rate_type=RateType.STREAM_DAY,
+            regularity=regularity,
+        )
+
+
+class TimeSeriesCalendarDayRate(TimeSeriesFloat):
```
TeeeJay, line 656: Probably not needed, was intended for HCExport etc that only resides in application layer, but then we can use the old TimeSeriesRate class (under a new name?)

src/libecalc/common/utils/rates.py
```diff
@@ -624,6 +624,56 @@ def resample(self, freq: Frequency) -> TimeSeriesIntensity:
...
+        return TimeSeriesRate(
+            timesteps=self.timesteps,
+            values=self.values,
+            unit=self.unit,
+            rate_type=RateType.CALENDAR_DAY,
+            regularity=regularity,
+        )
+
+
 class TimeSeriesRate(TimeSeries[float]):
```
TeeeJay, line 677: rename?

src/libecalc/core/consumers/compressor/component.py
```diff
@@ -107,16 +104,12 @@ def evaluate(
...
 
-        energy_usage = TimeSeriesRate(
+        energy_usage = TimeSeriesStreamDayRate(
             values=aggregated_result.energy_usage,
             timesteps=evaluated_timesteps,
             unit=aggregated_result.energy_usage_unit,
-            regularity=evaluated_regularity,
         )
 
-        if energy_usage.unit == Unit.STANDARD_CUBIC_METER_PER_DAY:
```
TeeeJay, line 117: to be handled in application layer? Needed? Nice to have what it was...but perhaps we shouldnt have different defaults?

src/libecalc/core/graph_result.py
```diff
@@ -271,12 +365,22 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
 
         installation_results = self._evaluate_installations(variables_map=self.variables_map)
 
+        regularities: Dict[str, TimeSeriesFloat] = {
```
TeeeJay, line 368: It is very cumbersome to collect, gather the regularities now...was the best (temporary) solution I found without tearing up everything

src/libecalc/dto/graph.py
```diff
@@ -27,17 +27,38 @@ def get_successors(self, component_id: str, recursively=False) -> List[str]:
...
             return self.graph.successors(component_id)
 
-    def get_predecessor(self, component_id) -> str:
+    def get_predecessor(self, component_id: str) -> str:
         predecessors = list(self.graph.predecessors(component_id))
         if len(predecessors) > 1:
             raise ValueError("Component with several parents encountered.")
         return predecessors[0]
 
+    def get_parent_installation_id(self, component_id: str) -> str:
```
TeeeJay, line 36: recursive, find first parent that is an installation. Not completly done. Might be better ways. Perhaps make more general.

src/libecalc/dto/result/emission.py
```diff
@@ -65,6 +67,26 @@ def create_empty(cls, name: str, timesteps: List[datetime]):
         )
 
 
+class PartialEmissionResult(TabularTimeSeries):
```
TeeeJay, line 70: We might want to have a temporary dto result, before calculating cumulatives....maybe we can remove it and calculate cumulatives right away

src/libecalc/core/graph_result.py
```diff
@@ -671,23 +799,31 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
                                 name=model.name,
                                 componentType=model.component_type,
                                 component_level=ComponentLevel.MODEL,
-                                energy_usage=model.energy_usage,
+                                energy_usage=model.energy_usage.to_timeseries_rate(regularity=regularity.values),
```
olelod, line 802: power can be none, so: 
```suggestion
energy_usage=model.energy_usage.to_timeseries_rate(regularity=regularity.values) if model.power is not None else None,
```

src/libecalc/core/graph_result.py
```diff
@@ -649,13 +770,20 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
...
-                                    model.power.to_volumes().to_unit(Unit.GIGA_WATT_HOURS).cumulative()
+                                    model.power.to_timeseries_rate(regularity=regularity.values)
+                                    .to_volumes()
+                                    .to_unit(Unit.GIGA_WATT_HOURS)
+                                    .cumulative()
                                     if model.power is not None
                                     else None
                                 ),
-                                power=model.power,
+                                power=model.power.to_timeseries_rate(regularity=regularity.values),
```
olelod, line 786: power can be None, so:
```suggestion
power=model.power.to_timeseries_rate(regularity=regularity.values) if model.power is not None else None,
```

src/libecalc/core/graph_result.py
```diff
@@ -671,23 +799,31 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
...
```
olelod, line 818: same here

src/ecalc_cli/cli_types.py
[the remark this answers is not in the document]
jsolaas: Did you mark ecalc_cli as sources root? I don't think I have this issue anymore when I only mark src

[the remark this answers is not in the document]
TeeeJay: I do not intend to include this in the PR. It was just done quickly because debugging didnt work with my current setup. Due to init file and which directory I start from. Didnt have time to debug at that time :)

src/libecalc/core/graph_result.py
```diff
@@ -45,6 +46,40 @@ class EnergyCalculatorResult(BaseModel):
...
+    Args:
+        emission_core_results:
+        regularities:
+
+    Returns:
+
+    """
+    dto_result: Dict[str, Dict[str, PartialEmissionResult]] = {}
+
+    for installation_name, emissions in emission_core_results.items():
```
olelod, line 67: emission_core_results are per consumer, not per installation?

TeeeJay: I just happened to got error on that in the test, so yes, it seems like you are right :)

TeeeJay: fixed

src/libecalc/core/consumers/compressor/component.py
jsolaas: Since users can change it in web it would be nice if we could have stream day for all. It would affect the csv though

src/libecalc/core/graph_result.py
```diff
@@ -724,50 +850,124 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
...
+                )
+                obj.rate_exceeds_maximum = consumer_result.component_result.rate_exceeds_maximum
+                obj.outlet_pressure_before_choking = consumer_result.component_result.outlet_pressure_before_choking
+            elif isinstance(obj, AssetResult):
+                obj.hydrocarbon_export_rate = TimeSeriesRate.from_timeseries_stream_day_rate(
+                    consumer_result.component_result.hydrocarbon_export_rate, regularity=regularity
+                )
+                obj.emission_intensities = consumer_result.component_result.emission_intensities
+            elif isinstance(obj, InstallationResult):
+                obj.regularity = consumer_result.component_result.regularity
```
jsolaas, line 956: TODO: don't use parse_obj above, instead create the object after the instance check core resultcomponent result?

src/libecalc/core/graph_result.py
```diff
@@ -724,50 +850,124 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
...
+            # )
 
         for installation in asset.installations:
+            regularity = regularities[installation.id]  # Already evaluated regularities
             for direct_emitter in installation.direct_emitters:
                 energy_usage = TimeSeriesRate(
                     timesteps=self.variables_map.time_vector,
                     values=[0.0] * self.variables_map.length,
-                    regularity=[1] * self.variables_map.length,
+                    regularity=[1] * self.variables_map.length,  # TODO: Correct? Dummy? Use calendar rate?
```
jsolaas, line 968: I think dummy, I'm assuming regularity was unused for DirectEmitters. It seems to be a feature that is not used anyway, so we probably won't get much feedback if we do something else...

src/libecalc/dto/result/results.py
```diff
@@ -66,21 +66,24 @@ class AssetResult(ComponentResultBase):
...
+    emission_intensities: Optional[List[EmissionIntensityResult]]
 
 
 class InstallationResult(AssetResult):
     """The installation result component."""
 
     componentType: Literal[ComponentType.INSTALLATION]
+    regularity: Optional[
+        TimeSeriesFloat
+    ]  # Regularity is currently set at per installation, send through. Possibly skip in output if confusing
```
jsolaas, line 79: Why send it through?

src/libecalc/dto/result/results.py
```diff
@@ -105,19 +108,19 @@ class GenericConsumerResult(EquipmentResultBase):
...
 class PumpResult(EquipmentResultBase):
     componentType: Literal[ComponentType.PUMP]
-    inlet_liquid_rate_m3_per_day: TimeSeriesRate
-    inlet_pressure_bar: TimeSeriesFloat
-    outlet_pressure_bar: TimeSeriesFloat
-    operational_head: TimeSeriesFloat
+    inlet_liquid_rate_m3_per_day: Optional[TimeSeriesRate]
+    inlet_pressure_bar: Optional[TimeSeriesFloat]
+    outlet_pressure_bar: Optional[TimeSeriesFloat]
+    operational_head: Optional[TimeSeriesFloat]
```
jsolaas, line 114: Why so many optional?

src/libecalc/output/results/queries.py
```diff
@@ -70,6 +74,16 @@ def query(
             time_steps=installation_time_steps,
         )
 
+        # TODO: Temp, get regularity - we should send in DTOs here instead
```
jsolaas, line 77: Should we?

jsolaas (commented):
Also remove regularity from components, and only store for installation? in dtos

src/libecalc/core/consumers/generator_set.py
```diff
@@ -37,7 +40,7 @@ def evaluate(
         """
         logger.debug(f"Evaluating Genset: {self.data_transfer_object.name}")
 
-        regularity = TemporalExpression.evaluate(
+        TemporalExpression.evaluate(
```
TeeeJay, line 43: remove

src/libecalc/output/results/queries.py
TeeeJay: lets see

TeeeJay:
already merged in a new PR

PR #229, opened 2023-10-04
chore: more robust surge control margin calculation

frodehk:

In some cases eCalc fails when the surge control margin is too high, for variable speed compressors. This happens because the new calculated minimum rate, based on the control margin, is outside the compressor chart for the given speed. 

The control margin is currently applied to the average of the minimum flow rate for all speed curves in the compressor chart, to find how much the minimum flow should be increased. The increase in minimum flow is the same for all speed curves in a variable speed compressor. Hence, for low speeds it is a risk that the updated minimum flow is outside the compressor chart - and eCalc fails.

- [x] Implement more robust surge margin calculation. The updated minimum flow is calculated individually for each speed, using the control margin as the increase in minimum flow, in percentage or fraction of the rate difference between minimum- and maximum flow, for the given speed.
- [x] Update tests
- [x] Update snapshots
- [x] Update docs

Comparing original vs. new method (green line) for an example case at 40% margin:
image

https://equinor-ecalc.atlassian.net/browse/ECALC-224?atlOrigin=eyJpIjoiNTJjZWZhMjI4NjI3NDk5NTlhZjAwNWM5YWZhYzYyNmIiLCJwIjoiaiJ9


src/libecalc/core/models/compressor/train/chart/variable_speed_compressor_chart.py
```diff
@@ -36,8 +36,8 @@ def _get_new_point(x: List[float], y: List[float], new_x_value) -> float:
             "Any usage of this functionality is at your own risk."
         )
         new_curves = []
-        adjust_minimum_rate_by = np.average([np.min(curve.rate) for curve in self.curves]) * control_margin
         for curve in self.curves:
+            adjust_minimum_rate_by = (np.max(curve.rate) - np.min(curve.rate)) * control_margin
```
TeeeJay, line 40: for single speed the above one will remain? Or is that implemented like this as well? (if implemented?)

TeeeJay:
Looks ok from my side. Need expertice. @olelod 

Remember to update changelog in docs as well

src/libecalc/core/models/compressor/train/chart/variable_speed_compressor_chart.py
frodehk: As far as I understand control margin is currently only implemented for variable speed compressors, am I right @olelod? It

olelod: When this was implemented it was only done for variable speed compressor charts. It should be implemented for single speed as well at some point, but in a different task, I guess.

olelod (approved):

PR #230, opened 2023-10-04
chore: move feature experimental to main method for requested pressures

frodehk:

Catch potential problems with new requested pressure feature, and warn user.

Move feature experimental to main method for requested pressures.

### DIFF
--- a/src/libecalc/core/graph_result.py
+++ b/src/libecalc/core/graph_result.py
@@ -218,9 +218,6 @@ class GraphResult:
         }
 
     @staticmethod
-    @Feature.experimental(
-        feature_description="Reporting requested pressures from compressor systems is an experimental feature."
-    )
     def operational_setting_used_id(timestep: datetime, operational_settings_used: TimeSeriesInt) -> int:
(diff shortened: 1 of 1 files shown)


src/libecalc/core/graph_result.py
```diff
@@ -229,6 +226,9 @@ def operational_setting_used_id(timestep: datetime, operational_settings_used: T
         return operational_setting_id - 1
 
     @staticmethod
+    @Feature.experimental(
+        feature_description="Reporting requested pressures from compressor systems is an experimental feature."
```
jsolaas, line 230: Er tanken her at vi ikke trenger å nevne compressor system? At `Reporting requested pressures is an experimental feature.`?

jsolaas (approved):

src/libecalc/core/graph_result.py
frodehk: Ja, det tror jeg ja - glemte den. Min tanke var opprinnelig at vi bare varslet ved bruk av system. Det er nå endret, bra spotta. Skal fikse.

PR #232, opened 2023-10-05
refactor: implement evaluate streams in models

jsolaas:
This gives models the option to handle streams separately instead of mixing.


src/libecalc/core/consumers/compressor/component.py
```diff
@@ -84,15 +81,11 @@ def evaluate(
...
-                raise NotImplementedError("Need to implement this")
-            elif issubclass(type(compressor), CompressorModel):
-                model_result = compressor.evaluate_rate_ps_pd(
-                    rate=np.asarray(total_requested_inlet_stream.rate.values),
-                    suction_pressure=np.asarray(total_requested_inlet_stream.pressure.values),
-                    discharge_pressure=np.asarray(operational_settings_for_timestep.outlet_stream.pressure.values),
-                )
-                model_results.append(model_result)
+            model_result = compressor.evaluate_streams(
+                inlet_streams=operational_settings.inlet_streams,
```
TeeeJay, line 85: streams are already mixed above in `total_requested_inlet_stream`. Should that be removed then?

src/libecalc/core/models/compressor/sampled/compressor_model_sampled.py
```diff
@@ -253,6 +254,18 @@ def evaluate_rate_ps_pd(
 
         return result
 
+    def evaluate_streams(
+        self,
+        inlet_streams: List[Stream],
+        outlet_stream: Stream,
+    ) -> CompressorTrainResult:
+        mixed_input_streams = Stream.mix_all(streams=inlet_streams)
```
TeeeJay, line 262: I think it makes sense to mix here, if we remove the one we already had, to not do it twice for different usage. @olelod what do you think?

src/libecalc/core/models/compressor/train/base.py
```diff
@@ -138,6 +139,27 @@ def evaluate_rate_ps_pd(
...
+        """
+        Evaluate model based on inlet streams and the expected outlet stream.
+        Args:
+            inlet_streams:
+            outlet_stream:
+
+        Returns:
+
+        """
+        mixed_inlet_stream = Stream.mix_all(inlet_streams)
```
TeeeJay, line 156: Do we ever need all input streams here? Maybe it is better to mix before we evaluate? Then always treat 1 in and output stream here?

src/libecalc/core/models/pump/pump.py
```diff
@@ -212,6 +221,16 @@ def evaluate_rate_ps_pd_density(
         return pump_result
 
 
+def evaluate_streams(self, inlet_streams: List[Stream], outlet_stream: Stream) -> PumpModelResult:
+    total_requested_stream = Stream.mix_all(inlet_streams)
```
TeeeJay, line 225: Should we create a class `Streams` that we use to handle multiple streams, like a "collection/container class"? And change method name to `Streams.mix(streams)`? Not a big change though..

TeeeJay (approved):
nice and clean. Would like input from @olelod , and would like @frodehk and @markusrf to also look at these PRs :)

olelod:
I think that this PR handles streams in the context of compressor trains where all additional streams are entering the compressor train at the first compressor stage (i.e. cross over streams from other trains in a compressor system). 

It is also a start for handling streams going in/out of compressor trains at any compressor stage.

src/libecalc/core/models/compressor/train/base.py
jsolaas: We should consider it, we could also look at the difference between the Streams container and operational_settings passed to compressor and pump. We could only pass Streams, then add separate arguments to evaluate for pump and compressor if we need to pass something else than streams.

src/libecalc/core/models/pump/pump.py
jsolaas: We should consider it, we could also look at the difference between the Streams container and operational_settings passed to compressor and pump. We could only pass Streams, then add separate arguments to evaluate for pump and compressor if we need to pass something else than streams.

src/libecalc/core/models/compressor/train/base.py
jsolaas: A compressor train could use the streams separately, so that is the point of this PR, to give that option when creating a model. Multiple streams does that. Currently we don't have an id for streams, so index must be used, but I think we should look into adding that id (name in yaml, from user?)

PR #236, opened 2023-10-12
feat: support name for crossover streams

jsolaas:

Allow handling of crossover streams in multiple streams.

Crossover syntax in system v2


src/libecalc/fixtures/cases/consumer_system_v2/consumer_system_v2_dto.py
```diff
@@ -231,7 +239,14 @@
     consumes=ConsumptionType.FUEL,
     fuel=fuel,
     component_conditions=SystemComponentConditions(
-        crossover=[0, 1, 1],
+        crossover=[
+            SystemCrossoverStream(
+                from_component_id=generate_id("compressor2"), to_component_id=generate_id("compressor1")
```
jsolaas, line 244: Using component_id as name is not something we use in core, but this interface isn't very nice when creating a crossover stream in python, in dtos the id is created automatically. We could do that for crossover stream also

src/libecalc/fixtures/cases/consumer_system_v2/data/consumer_system_v2.yaml
```diff
@@ -126,7 +130,11 @@ INSTALLATIONS:
           - NAME: compressor3
             ENERGY_USAGE_MODEL: compressor_sampled_1d
         COMPONENT_CONDITIONS:
-          CROSSOVER: [ 0, 1, 1 ]
+          CROSSOVER:
+            - FROM: compressor2
+              TO: compressor1
+            - FROM: compressor3
+              TO: compressor1
```
jsolaas, line 137: How can we support several crossovers from the same consumer? I guess we could use max_rate of the receiving consumer to decide, use crossover if receiving consumer can handle it, if not skip to next crossover or from-consumer becomes invalid.

olelod: I guess you could make TO: a list, and then pass whatever the first consumer has room for (using get_max_rate) to the first consumer on the list, and then move to the next one with the remaining volume. I don't know how many actual use cases we would have for that, but it should at least be possible to implement like that. Or?

src/libecalc/common/string_utils.py
```diff
@@ -34,4 +34,5 @@ def to_camel_case(string: str) -> str:
 
     """
     string_split = string.replace("__", "_").split("_")
+    string_split = [word for word in string_split if len(word) > 0]  # Allow names such as 'from_'
```
TeeeJay, line 37: Allow words that end with an underscore?

src/libecalc/core/consumers/consumer_system.py
```diff
@@ -47,8 +47,14 @@ def get_subset_for_timestep(self, current_timestep: datetime) -> Self:
         ...
 
 
+class SystemCrossoverStream(Protocol):
+    stream_name: Optional[str]
+    from_component_id: str
```
TeeeJay, line 52: synes at vi skal starte innføre IDer/Referanser som en egen type, og ikke bare en streng. Da kan variabelen heter `from` og typen er ComponentID ? eller ComponentReference?

src/libecalc/core/consumers/consumer_system.py
```diff
@@ -47,8 +47,14 @@ def get_subset_for_timestep(self, current_timestep: datetime) -> Self:
         ...
 
 
+class SystemCrossoverStream(Protocol):
+    stream_name: Optional[str]
```
TeeeJay, line 51: Burde vi tvinge navn? Er evnt fallbacken en autogenerering av `f{from}_{to}` eller lignende?

src/libecalc/core/consumers/consumer_system.py
```diff
@@ -47,8 +47,14 @@ def get_subset_for_timestep(self, current_timestep: datetime) -> Self:
         ...
 
 
+class SystemCrossoverStream(Protocol):
```
TeeeJay, line 50: Er det nok å kallen den `Crossover`?

src/libecalc/core/consumers/consumer_system.py
```diff
@@ -95,31 +101,28 @@ def _get_operational_settings_adjusted_for_crossover(
...
-        crossover_streams_map: Dict[int, List[Stream]] = {
-            consumer_index: [] for consumer_index in range(len(self._consumers))
+        crossover_streams_map: Dict[str, List[Stream]] = {consumer.id: [] for consumer in self._consumers}
+        crossover_definitions_map: Dict[str, SystemCrossoverStream] = {
+            crossover_stream.from_component_id: crossover_stream
+            for crossover_stream in self._component_conditions.crossover
         }
 
-        # Converting from index 1 to index 0.
-        crossover = [crossover_flow_to_index - 1 for crossover_flow_to_index in self._component_conditions.crossover]
```
TeeeJay, line 103: :)

src/libecalc/dto/components.py
```diff
@@ -190,8 +190,17 @@ class PumpSystemOperationalSetting(EcalcBaseModel):
     outlet_pressures: List[Expression]
 
 
+class SystemCrossoverStream(EcalcBaseModel):
```
TeeeJay, line 193: Will this not be the same all places `Crossover` is used, so just call it `Crossover`?

src/libecalc/dto/components.py
```diff
@@ -190,8 +190,17 @@ class PumpSystemOperationalSetting(EcalcBaseModel):
...
+    class Config:
+        allow_population_by_field_name = True
+
+    stream_name: Optional[str] = Field(None)
+    from_component_id: str
```
TeeeJay, line 198: I would want to have a special syntax on references, to make it clear that it is a reference, such as prefix with $ (which is basically a standard?!). Then it would make sense to define our own type for the places where we require "references"...

src/libecalc/fixtures/cases/consumer_system_v2/consumer_system_v2_dto.py
TeeeJay: If we take the unique name one step further, to enforce unique names across installations as well..it would be easier? If we can...then we can skip that `generate_id()` ...

src/tests/libecalc/core/consumers/test_consumer_utils.py
```diff
@@ -1,25 +1,53 @@
...
+from libecalc.common.string_utils import generate_id
 from libecalc.core.consumers.consumer_system import ConsumerSystem
 from libecalc.core.consumers.legacy_consumer.consumer_function.utils import (
     apply_condition,
 )
+from libecalc.dto.components import SystemCrossoverStream
+
+
+@dataclass
+class ConsumerMock:
```
TeeeJay, line 13: :)

src/libecalc/fixtures/cases/consumer_system_v2/consumer_system_v2_dto.py
jsolaas: We have unique names across installations, so id is just a hash of name, we could use name, but if we do that in core, then what is the point of the id? But it would make things easier, so I think we can use name as id.

src/libecalc/common/string_utils.py
jsolaas: ends and starts I think, I added the commented test cases. Needed for alias when having an attribute that is a reserved word in python

src/libecalc/core/consumers/consumer_system.py
jsolaas: Enig

src/libecalc/core/consumers/consumer_system.py
jsolaas: Fallback er indeks, eller at man ikke trenger å referere til den streamen direkte (man mikser alle inlet streams). Kan alltids generere et navn for output. 

Kan vurdere å tvinge navn.

src/libecalc/dto/components.py
jsolaas: Would a reference meaning "insert reference here" (i.e. yaml include-ish) be different from this reference where we mean "from the component with this name" (but not insert the yaml)

PR #237, opened 2023-10-12
fix: wrong data for boilers and heaters in ltp-results

frodehk:

Calculation of nmvoc-emissions for the `BOILER`- and `HEATER` ltp-categories are wrong. The emission query asks for nox instead of nmvoc.

Update emission query to ask for correct emission type.

https://equinor-ecalc.atlassian.net/browse/ECALC-313?atlOrigin=eyJpIjoiM2EwY2MwYzQ4ZTRiNGY5ZGEzZmMyMTgyNjAwYTJiYWMiLCJwIjoiaiJ9

### DIFF
--- a/docs/docs/changelog/next.md
+++ b/docs/docs/changelog/next.md
@@ -17,7 +17,7 @@ sidebar_position: 1
 
 
 ## Fixes
-
+- `nmvoc` emissions were incorrectly reported for the ltp categories `HEATER` and `BOILER`: The emission query filters included `nox`, and are now corrected to `nmvoc`.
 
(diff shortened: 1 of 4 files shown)


TeeeJay (approved):
This should have lead to changes in ltp ficture snapshot...but it doesnt...because it is not tested...can you add a simple test for that in LTP export test?

frodehk:
> This should have lead to changes in ltp ficture snapshot...but it doesnt...because it is not tested...can you add a simple test for that in LTP export test?

I will add a test.

PR #238, opened 2023-10-12
fix: ensure unique names in system v2

jsolaas:
Also fix error message when this fails.


src/libecalc/input/validation_errors.py
```diff
@@ -126,7 +126,7 @@ def __init__(
                 error_location_info = " -> ".join(
                     [str(s).capitalize().replace("__root__", "General error").replace("_", " ") for s in error_loc]
                 )
-                if component_name := data.get(EcalcYamlKeywords.name):
+                if data is not None and (component_name := data.get(EcalcYamlKeywords.name)):
```
jsolaas, line 129: Data is optional

TeeeJay (approved):

PR #242, opened 2023-10-17
feat: multiple streams in system

jsolaas:
Multiple streams can be used to handle crossover streams individually in a train. Currently no additional streams can be given, as the system does not support providing several rates/streams per consumer.


src/libecalc/common/stream.py
```diff
@@ -15,6 +15,7 @@ class Config:
         alias_generator = to_camel_case
         allow_population_by_field_name = True
 
+    name: Optional[str]
     rate: TimeSeriesRate
```
olelod, line 19: I guess we need either a (unique) name or an id to attach a stream to a compressor stage when we later allow for streams going into/out of any compressor stage. 


Also consider including temperature here and the full fluid model/composition, not just the density.

jsolaas: I agree, the plan is to implement that in another PR. I'm thinking a unique name for a stream within a component makes sense.

src/libecalc/fixtures/cases/consumer_system_v2_multiple_streams/data/consumer_system_v2.yaml
```diff
@@ -0,0 +1,128 @@
...
+              - NAME: compressor2
+                ENERGY_USAGE_MODEL: compressor_sampled_1d
+              - NAME: compressor3
+                ENERGY_USAGE_MODEL: compressor_sampled_1d
+            COMPONENT_CONDITIONS:
+              CROSSOVER:
+                - NAME: in_stream_crossover_1
+                  FROM: compressor2
+                  TO: compressor1
+                - FROM: compressor3
```
jsolaas, line 117: missing name

src/libecalc/fixtures/cases/consumer_system_v2_multiple_streams/data/consumer_system_v2.yaml
```diff
@@ -0,0 +1,128 @@
...
+          - NAME: compressor_system_v2
+            CATEGORY: COMPRESSOR
+            TYPE: COMPRESSOR_SYSTEM@v2
+            CONSUMERS:
+              - NAME: compressor1
+                ENERGY_USAGE_MODEL: compressor_with_turbine
+              - NAME: compressor2
+                ENERGY_USAGE_MODEL: compressor_sampled_1d
+              - NAME: compressor3
+                ENERGY_USAGE_MODEL: compressor_sampled_1d
```
jsolaas, line 111: Should only consume power, not fuel

src/tests/libecalc/integration/snapshots/test_consumer_system_v2/test_compressor_system_v2_results/consumer_system_v2-consumer_system_v20/consumer_system_v2.json
```diff
@@ -242,6 +242,7 @@
                                     2.0
                                 ]
                             },
+                            "name": null,
```
TeeeJay, line 245: I think we should enforce it ... or generate it

src/libecalc/core/models/compressor/train/variable_speed_compressor_train_common_shaft_multiple_streams_and_pressures.py
```diff
@@ -95,6 +96,54 @@ def __init__(
         # previous time step to recirculate. This will take care of that.
         self.fluid_to_recirculate_in_stage_when_inlet_rate_is_zero = [None] * len(self.stages)
 
+    def evaluate_streams(
```
TeeeJay, line 99: It seems that this is not in use yet, just making it ready? Is that so? Possible to use it, and e.g. enable it with a feature flag? Or is it not ready?

TeeeJay (approved):

src/libecalc/core/models/compressor/train/variable_speed_compressor_train_common_shaft_multiple_streams_and_pressures.py
jsolaas: Used in v2. I'm overwriting the parent method so there is no change where it is used.

PR #243, opened 2023-10-17
fix: update ecalc validation for yaml file in web

krisolba1:

 - Removed PRESSURE_CONTROL as required for single speed compressor train, as it defaults to DOWNSTREAM_CHOKE.
 - Added $ to expression regexp as this seems to be much used , e.g. " VALUE: 0.5 {*} $var.some_file"
 - Curve for chart can be either specified by object or FILE. 
 -


jsolaas (approved):
Very nice

PR #244, opened 2023-10-18
refactor: remove temporal operational settings system v2

jsolaas:
It is unclear if it is needed, and it causes confusion. We want users to use temporal variables instead in most cases. Removing ensures we actually capture the need when it would be used.


jsolaas:
> TODO: Show and test how we can use temporary variables instead

Where should we show and test? Examples folder? https://test.ecalc.equinor.com/project/112/models/user/913

PR #245, opened 2023-10-18
refactor: separate optimization from system

jsolaas:


src/libecalc/common/priotity_optimizer.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@dataclass
+class EvaluatorResult(Generic[TResult]):
+    id: str
+    result: TResult
+    is_valid: TimeSeriesBoolean
+
+
+class PriorityOptimizer(Generic[TResult]):
```
jsolaas, line 31: Add TPriority

src/libecalc/common/priotity_optimizer.py
```diff
@@ -0,0 +1,81 @@
...
+    def optimize(
+        self,
+        variables_map: VariablesMap,
+        priorities: List[TPriority],
+        evaluator: typing.Callable[[datetime, TPriority], List[EvaluatorResult]],
```
jsolaas, line 36:
```suggestion
        evaluator: typing.Callable[[datetime, TPriority], List[EvaluatorResult[TResult]]],
```

src/libecalc/common/priotity_optimizer.py
```diff
@@ -0,0 +1,81 @@
...
+from libecalc.dto import VariablesMap
+
+TResult = TypeVar("TResult")
+TPriority = TypeVar("TPriority")
+
+
+@dataclass
+class PriorityOptimizerResult(Generic[TResult]):
+    priorities_used: TimeSeriesInt
+    priority_results: Dict[datetime, Dict[int, Dict[str, TResult]]]
```
TeeeJay, line 21: Would prefer to have this as a type/class as well to better understand what the different keys and values represent :)

src/libecalc/common/priotity_optimizer.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@dataclass
+class EvaluatorResult(Generic[TResult]):
+    id: str
+    result: TResult
+    is_valid: TimeSeriesBoolean
+
+
+class PriorityOptimizer(Generic[TResult, TPriority]):
```
TeeeJay, line 31: perhaps a descritpion of how this optimizer works?

TeeeJay (approved):
I like that we separate this algorithm from consumer system, that is correct imo :) Please dont merge until I have merged regularity as it will mess up stuff afaik...

PR #248, opened 2023-10-20
chore: make sure no mismatch between timestemps and time series values

frodehk:

When there is a mismatch between number timesteps and number of time series values, this may be silently ignored. The danger is that a single value or few values are extrapolated, or mapped incorrectly, and therefore getting wrong results.
 
We should make sure that we always have a 1:1 relationshop between timesteps and values when working with timeseries to make sure that this cannot be a problem.

- [x] Add a root validator in `TimeSeries` to ensure there is no mismatch between timesteps and values.
- [ ] Update `test_adding_timeseriesrate`

https://equinor-ecalc.atlassian.net/browse/ECALC-310?atlOrigin=eyJpIjoiN2MyYjA5MzE4YmZhNGQ3M2JmZmU5ZDU1MmQ0MTEyYzIiLCJwIjoiaiJ9


src/libecalc/common/utils/rates.py
```diff
@@ -152,11 +152,28 @@ class Config:
...
         return v
 
+    @root_validator(pre=True)
+    def timesteps_values_one_to_one(cls, field_values):
+        nr_timesteps = len(field_values["timesteps"])
+        nr_values = len(field_values["values"])
+
+        if not cls.__name__ == TimeSeriesVolumes.__name__:
+            if nr_timesteps != nr_values:
+                if all(math.isnan(i) for i in field_values["values"]):
```
frodehk, line 167: Some situations: number of timesteps (e.g. 12) differ from number of timesteps values (e.g 2). The values are all NaN. To make things run, the values list is extended with NaN to have same dimension as the timesteps. However, this is probably not the right place to do this. And, ruff fails saying it is a pointless comparison.

src/libecalc/common/utils/rates.py
```diff
@@ -152,11 +152,28 @@ class Config:
...
         if field.outer_type_ is float and v is None:
             return math.nan
         return v
 
+    @root_validator(pre=True)
+    def timesteps_values_one_to_one(cls, field_values):
+        nr_timesteps = len(field_values["timesteps"])
+        nr_values = len(field_values["values"])
+
+        if not cls.__name__ == TimeSeriesVolumes.__name__:
```
TeeeJay, line 165: is it possible to override this in the TimeSeriesVolumes class? Just asking

TeeeJay: but good that you think of all the consequences :)

src/libecalc/common/utils/rates.py
```diff
@@ -152,11 +152,28 @@ class Config:
...
+            if nr_timesteps != nr_values:
+                if all(math.isnan(i) for i in field_values["values"]):
+                    # TODO: This should be solved another place. Temporary solution to make things run
+                    field_values["values"] == [math.nan] * len(field_values["timesteps"])
+                else:
+                    raise ProgrammingError(
+                        "Time series: number of timesteps do not match number " "of values. Most likely a bug."
```
TeeeJay, line 172: Report to eCalc Dev Team :)

TeeeJay (changes requested):
Please create snapshots, if you are able to, so that we can see how much that is affected :) How many tests are failing?

frodehk:
Trouble with rebasing

src/libecalc/common/utils/rates.py
frodehk: Done (had to create a new PR, rebasing stuff...)

src/libecalc/common/utils/rates.py
frodehk: Done (another PR)

src/libecalc/common/utils/rates.py
frodehk: New PR: https://github.com/equinor/ecalc/pull/251

frodehk:
> Please create snapshots, if you are able to, so that we can see how much that is affected :) How many tests are failing?

No tests are failing. Ok to create a simple test without snapshots?

PR #250, opened 2023-10-24
refactor: use Graph object to build graph

jsolaas:
Proper graph implementation, isolate networkx and implementation details.


TeeeJay (approved):
Nice, one step on the way to make core agnostic about dependencies :)

We need to give both graph, node and edge proper names related to the domain though (imo). Such as `EnergyNetwork` (PowerNetwork, fuel network etc) or `Piping System` (TBD), whatever it is called, but then internally we use a graph, so that the `Graph` is never exposed to users.

```
pipingSystem.addPipeline(..,..)
pipeline.connect(...,...)
```

etc?

PR #251, opened 2023-10-24
chore: make sure no mismatch between timestemps and time series values

frodehk:

When there is a mismatch between number timesteps and number of time series values, this may be silently ignored. The danger is that a single value or few values are extrapolated, or mapped incorrectly, and therefore getting wrong results.
 
We should make sure that we always have a 1:1 relationshop between timesteps and values when working with timeseries to make sure that this cannot be a problem.

- [x] Add a regular validator in `TimeSeries` to ensure there is no mismatch between timesteps and values.
- [x] Add test.

### DIFF
--- a/src/libecalc/common/utils/rates.py
+++ b/src/libecalc/common/utils/rates.py
@@ -580,6 +580,18 @@ class TimeSeriesVolumesCumulative(TimeSeries[float]):
 
 
 class TimeSeriesVolumes(TimeSeries[float]):
+    @validator("values", pre=True)
+    def check_length_timestep_values(cls, v: list, values, field: ModelField):
+        # Initially timesteps for volumes contains one more item than values
+        # After reindex number of timesteps equals number of values
+        # TODO: Ensure periodical volumes are handled in a consistent way. Why different after reindex?
+        if len(v) not in [len(values["timesteps"]), len(values["timesteps"]) - 1]:
+            raise ProgrammingError(
+                "Time series: number of timesteps do not match number "
+                "of values. Most likely a bug, report to eCalc Dev Team."
+            )
+        return v
+
     def resample(self, freq: Frequency):
         msg = (
             f"{self.__repr_name__()} does not have an resample method."
--- a/src/libecalc/common/utils/rates.py
+++ b/src/libecalc/common/utils/rates.py
@@ -9,6 +9,7 @@ from enum import Enum
 from typing import (
     Any,
     DefaultDict,
+    Dict,
     Generic,
     Iterable,
     Iterator,
@@ -162,7 +163,7 @@ class TimeSeries(GenericModel, Generic[TimeSeriesValue], ABC):
         return v
 
     @validator("values", pre=True)
-    def timesteps_values_one_to_one(cls, v: list, values, field: ModelField):
+    def timesteps_values_one_to_one(cls, v: List[Any], values: Dict[str, Any]):
         nr_timesteps = len(values["timesteps"])
         nr_values = len(v)
 
@@ -581,7 +582,7 @@ class TimeSeriesVolumesCumulative(TimeSeries[float]):
 
 class TimeSeriesVolumes(TimeSeries[float]):
     @validator("values", pre=True)
-    def check_length_timestep_values(cls, v: list, values, field: ModelField):
+    def check_length_timestep_values(cls, v: List[Any], values: Dict[str, Any]):
         # Initially timesteps for volumes contains one more item than values
         # After reindex number of timesteps equals number of values
         # TODO: Ensure periodical volumes are handled in a consistent way. Why different after reindex?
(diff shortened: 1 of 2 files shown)


src/libecalc/common/utils/rates.py
```diff
@@ -156,6 +156,23 @@ def convert_none_to_nan(cls, v: float, field: ModelField) -> TimeSeriesValue:
             return math.nan
         return v
 
+    @validator("values", pre=True)
+    def timesteps_values_one_to_one(cls, v: list, values, field: ModelField):
+        nr_timesteps = len(values["timesteps"])
+        nr_values = len(v)
+
+        if not cls.__name__ == TimeSeriesVolumes.__name__:
```
TeeeJay, line 164: Can you add an additional test for TimeSeriesVolumes, that the `len(values) == len(timesteps) -1` ?

I think it was where i saw a lot of errors when working with regularity. I dont think i fixed anything

src/libecalc/common/utils/rates.py
```diff
@@ -156,6 +156,23 @@ def convert_none_to_nan(cls, v: float, field: ModelField) -> TimeSeriesValue:
...
+            if nr_timesteps != nr_values:
+                if all(math.isnan(i) for i in v):
+                    # TODO: This should probably be solved another place. Temporary solution to make things run
```
TeeeJay, line 167: yes, this is a symptom that nr of timesteps are off...for some reason. It shouldnt be...right? It should be fixed where error arises, to make sure we have the same nr of values all the time...

src/libecalc/common/utils/rates.py
frodehk: TimeSeriesVolumes: initially len(values) == len(timesteps) - 1. 

However, after TimeSeriesVolumes.reindex: len(values) == len(timesteps).

Not sure why we have this difference, suggestions?

TeeeJay: It is because volumes uses "periods" while the other timeseries uses "timesteps". We should change the timeseries that use periods to do that directly, instead of indirectly as the period between previous timestep and next timestep. In a timeseries with 10 timesteps, there are 9 periods (=which represent volumes). It is easy to do mistakes, misunderstand etc, so reasons to fix that. However, as it is now, that is why I suggested the alternative check above.

src/libecalc/common/utils/rates.py
frodehk: This is most likely a bug, it happens in `EnergyModelBaseResult.extend()`, related to handling of None etc. Will be reported as separate issue.

src/libecalc/common/utils/rates.py
```diff
@@ -531,6 +548,18 @@ def to_volumes(self) -> TimeSeriesVolumes:
 
 
 class TimeSeriesVolumes(TimeSeries[float]):
+    @validator("values", pre=True)
+    def check_length_timestep_values(cls, v: list, values, field: ModelField):
+        # Initially timesteps for volumes contains one more item than values
+        # After reindex number of timesteps equals number of values
+        # TODO: Ensure periodical volumes are handled in a consistent way. Why different after reindex?
+        if len(v) not in [len(values["timesteps"]), len(values["timesteps"]) - 1]:
```
frodehk, line 556: The periodical volumes initially have one more timestep than values. After reindexing this changes. This is why the check is looking for two values. This should be looked into. Note that periodical volumes are dependent on the actual timesteps (one more than number of periods), if we should continue to calculate cumulatives from periodical volumes in the same way as we do.

src/libecalc/common/utils/rates.py
```diff
@@ -156,6 +156,23 @@ def convert_none_to_nan(cls, v: float, field: ModelField) -> TimeSeriesValue:
             return math.nan
         return v
 
+    @validator("values", pre=True)
+    def timesteps_values_one_to_one(cls, v: list, values, field: ModelField):
```
TeeeJay, line 160:
```suggestion
    def timesteps_values_one_to_one(cls, v: List[Any], values: Dict[str, Any], field: ModelField):
```
typehint as detailed as possible. Is the suggested change correct? I guess this is before the `v` becomes and is parsed to `TimeSeriesValue` but not sure...

src/libecalc/common/utils/rates.py
```diff
@@ -531,6 +548,18 @@ def to_volumes(self) -> TimeSeriesVolumes:
 
 
 class TimeSeriesVolumes(TimeSeries[float]):
+    @validator("values", pre=True)
+    def check_length_timestep_values(cls, v: list, values, field: ModelField):
```
TeeeJay, line 552: typehints here as well

src/libecalc/common/utils/rates.py
TeeeJay: Similar to StreamDay "problem", the timesteps should be in core, while volume should only be handled in the layer where we are reporting etc

TeeeJay (approved):

PR #252, opened 2023-10-25
chore: only calculate max standard rate for time steps with valid inp…

olelod:
…ut operational conditions

The operating conditions (rate, suction/intermediate/discharge pressures) are validated. Invalid suction pressure input is flagged and the rate is set to zero such that no calculations are performed. The pressure input is not changed so the current code will try to calculate the max standard rate also in situations where i.e. the discharge pressure is less than the suction pressure. 

Instead of calculating valid indexes inside get_max_standard_rate for each type of compressor train, the max standard rate is calculated only for the time steps that are already validated to be valid (that does not have a input_failure_status).

[ECALC 302](https://equinor-ecalc.atlassian.net/browse/ECALC-302)

### DIFF
--- a/src/libecalc/core/models/compressor/train/base.py
+++ b/src/libecalc/core/models/compressor/train/base.py
@@ -7,7 +7,11 @@ from libecalc.common.feature_flags import Feature
 from libecalc.common.logger import logger
 from libecalc.common.stream import Stream
 from libecalc.common.units import Unit
-from libecalc.core.models import ModelInputFailureStatus, validate_model_input
+from libecalc.core.models import (
+    INVALID_INPUT,
+    ModelInputFailureStatus,
+    validate_model_input,
+)
 from libecalc.core.models.compressor.base import CompressorModel
 from libecalc.core.models.compressor.results import CompressorTrainResultSingleTimeStep
 from libecalc.core.models.compressor.train.fluid import FluidStream
@@ -23,6 +27,7 @@ from libecalc.dto.models.compressor.train import (
 from numpy.typing import NDArray
 
 TModel = TypeVar("TModel", bound=CompressorTrainDTO)
+INVALID_MAX_RATE = INVALID_INPUT
 
 
 class CompressorTrainModel(CompressorModel, ABC, Generic[TModel]):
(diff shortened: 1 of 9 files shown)


src/libecalc/core/models/compressor/train/base.py
```diff
@@ -101,17 +101,20 @@ def evaluate_rate_ps_pd(
         )
 
         if self.data_transfer_object.calculate_max_rate:
+            # only calculate max standard rate for time steps with valid input
+            valid_idx = np.where([failure_status is None for failure_status in input_failure_status], 1, 0)
```
TeeeJay, line 105: can we create a failure_status enum: `NO_FAILURE = None` to make sure that is consistent and use that as check?

TeeeJay, line 105:
```suggestion
            valid_indices = np.where([failure_status is FailureStatus.NO_FAILURE for failure_status in input_failure_status], 1, 0)
```

src/libecalc/core/models/compressor/train/base.py
```diff
@@ -101,17 +101,20 @@ def evaluate_rate_ps_pd(
...
             if isinstance(self.data_transfer_object, dto.VariableSpeedCompressorTrainMultipleStreamsAndPressures):
-                max_standard_rate = self.get_max_standard_rate_per_stream(
-                    suction_pressures=suction_pressure,
-                    discharge_pressures=discharge_pressure,
-                    rates_per_stream=rate,
+                max_standard_rate_for_valid_idx = self.get_max_standard_rate_per_stream(
```
TeeeJay, line 107:
```suggestion
                max_standard_rate_for_valid_indices = self.get_max_standard_rate_per_stream(
```

src/libecalc/core/models/compressor/train/base.py
```diff
@@ -101,17 +101,20 @@ def evaluate_rate_ps_pd(
...
                 )
             else:
-                max_standard_rate = self.get_max_standard_rate(
-                    suction_pressures=suction_pressure,
-                    discharge_pressures=discharge_pressure,
+                max_standard_rate_for_valid_idx = self.get_max_standard_rate(
+                    suction_pressures=suction_pressure[valid_idx],
+                    discharge_pressures=discharge_pressure[valid_idx],
                 )
+            max_standard_rate = np.where(valid_idx, max_standard_rate_for_valid_idx, np.nan)
```
TeeeJay, line 117: should we start also referring to invalid values via constant or enum, to make sure it is consistent?

TeeeJay, line 117:
```suggestion
            INVALID_INPUT = np.nan
            INVALID_MAX_RATE = INVALID_INPUT
            max_standard_rate = np.where(valid_idx, max_standard_rate_for_valid_idx, INVALID_MAX_RATE)
```

open for other suggestions

TeeeJay: INVALID_INPUT should then be in a errors.py file or similar, where we have code related to invalid data etc

TeeeJay (approved):
Good. But no tests are failing, and no tests provided... Possible to indicate the change with changing an old test or adding anew test for that?

olelod:
A small refactor of model input validation first (https://github.com/equinor/ecalc/pull/256), then some small changes here. Maybe you want to take a new look @TeeeJay?

PR #254, opened 2023-10-26
fix: don't require HCEXPORT in editor

jsolaas:

schema wrong

schema correct


TeeeJay (commented):
What happens when hcexport is not set and user looks as intensity? 0/? will it give NaN or 0? Is that explicitly handled?

jsolaas:
> What happens when hcexport is not set and user looks as intensity? 0/? will it give NaN or 0? Is that explicitly handled?

hcexport is set to 0 and intensity plots will be empty. Not sure what the values of intensity are, nan converted to None maybe.

TeeeJay (approved):

PR #256, opened 2023-10-27
refactor: move validate operational conditions from compressor train

olelod:
Rename validate_operational_conditions to validate_model_input because it is validation of model input.

Move the validation out of compressor train, since model input of other models (such as pumps) should also be validated in the same way.

Adding ModelInputFailureStatus enum to keep track of valid/invalid model input.


src/libecalc/core/models/model_input_validation.py
```diff
@@ -0,0 +1,209 @@
...
+            validated_rate[indices_to_validate] = tmp_rate
+        validated_suction_pressure[indices_to_validate] = tmp_suction_pressure
+        validated_discharge_pressure[indices_to_validate] = tmp_discharge_pressure
+        if intermediate_pressure is not None:
+            validated_intermediate_pressure[indices_to_validate] = tmp_intermediate_pressure
+        for i, failure in enumerate(tmp_failure_status):
+            validated_failure_status[indices_to_validate[i]] = failure
+
+    # any remaining zero or negative suction/discharge pressures (for unvalidated time steps, others are already changed)
+    # must be set to 1 (for neqsim to initiate fluid streams)
```
jsolaas, line 64: Why do we have unvalidated timesteps, and why is the pressure values relevant then? Unvalidated has zero rate or negative as far as I understand. I assume it is related to multiple streams, but not sure

olelod: I actually think it is the current implementation of simplified train that requires it. 

(https://github.com/equinor/ecalc/blob/c7f5a9955c55fdc970f60b6aade8a0793acab27a/src/libecalc/core/models/compressor/train/simplified_train.py#L162)

PR #261, opened 2023-10-30
fix: wrong handling of values and timesteps in temporal models

frodehk:

Problem when extending time series for compressors across temporal models in `EnergyModelBaseResult.extend()`. This is basically related to how None-values are treated. Two bugs:

- Loss of data: When first temporal model contains `None` and next temporal model contains real values, the merge/extend function evaluates to `None` - ignoring data from second temporal model.
- Mismatch between timesteps and values: When first temporal model contains real values, and next temporal model contains `None`, the merge/extend function evaluates to the values of the first temporal model. The timesteps in the second model are ignored, hence a mismatch between number of timesteps and number of values are introduced for the actual time series.

- [x] `CompressorTrainResultSingleStep`: Ensure that all time series have values or NaN for all valid timesteps, even if input is `None`
- [x] `CompressorModelSampled`: Ensure that all time series have values or NaN for all valid timesteps, even if input is `None`
- [x] Update snapshots
- [x] Add test

https://equinor-ecalc.atlassian.net/browse/ECALC-376?atlOrigin=eyJpIjoiNzY4MzVlMWU2ZjEzNDMwNzhiYzNhMDZkMmExYzljOTQiLCJwIjoiaiJ9

### DIFF
--- a/src/libecalc/core/models/compressor/results.py
+++ b/src/libecalc/core/models/compressor/results.py
@@ -316,22 +316,22 @@ class CompressorTrainResultSingleTimeStep(BaseModel):
             compressor_stage_result[i].rate_has_recirculation = [
                 bool(result_list[t].stage_results[i].rate_has_recirculation)
                 for t in range(len(result_list))
-                if result_list[t].stage_results[i].rate_has_recirculation is not None
+                if bool(result_list[t].stage_results[i].rate_has_recirculation) is not None
             ]
             compressor_stage_result[i].rate_exceeds_maximum = [
                 bool(result_list[t].stage_results[i].rate_exceeds_maximum)
                 for t in range(len(result_list))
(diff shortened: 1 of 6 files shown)


TeeeJay (commented):
it looks good, but all snapshots are expanding `null` to `NaN`..I guess we have no cases where real values are actually overwritten? Good with test. Need UAT to have a look to QA this :) Good job though :)

frodehk:
> it looks good, but all snapshots are expanding `null` to `NaN`..I guess we have no cases where real values are actually overwritten? Good with test. Need UAT to have a look to QA this :) Good job though :)

Yes, I think None is resolved to null in json (?). And, None is now filled with nans to ensure correct number of values vs timesteps.

TeeeJay (approved):

PR #263, opened 2023-10-30
fix: make iteration loops for simplified train consistent

olelod:

https://github.com/equinor/ecalc/blob/883b7e6888d5ff4ddca41cbeac0f7c7dd96e60a6/src/libecalc/core/models/compressor/train/simplified_train.py#L481 and https://github.com/equinor/ecalc/blob/883b7e6888d5ff4ddca41cbeac0f7c7dd96e60a6/src/libecalc/core/models/compressor/train/utils/enthalpy_calculations.py#L20 both have the same iterative loop changing z and kappa, but the algorithm for finding the max rate has an extra update in the end (https://github.com/equinor/ecalc/blob/883b7e6888d5ff4ddca41cbeac0f7c7dd96e60a6/src/libecalc/core/models/compressor/train/simplified_train.py#L563). Removing this extra update solves the numerical problems reported in this issue.

ECALC 381

### DIFF
--- a/src/tests/libecalc/core/models/compressor_modelling/test_simplified_compressor_train.py
+++ b/src/tests/libecalc/core/models/compressor_modelling/test_simplified_compressor_train.py
@@ -194,7 +194,7 @@ def test_calculate_maximum_rate_given_outlet_pressure_all_calculation_points(
     pressure_ratios = [1, 2, 3, 4, 5, 10, 100, 1000]
 
     # These expected max rates are here just to assure stability in the results. They are not assured to be correct!
-    approx_expected_max_rates = [1116990, 1359255, 1536130, 1052085, 1052085, 1052085, 1052085, 1052085]
+    approx_expected_max_rates = [1116990, 1359022, 1536130, 1052085, 1052085, 1052085, 1052085, 1052085]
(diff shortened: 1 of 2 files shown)


src/tests/libecalc/core/models/compressor_modelling/test_simplified_compressor_train.py
```diff
@@ -194,7 +194,7 @@ def test_calculate_maximum_rate_given_outlet_pressure_all_calculation_points(
     pressure_ratios = [1, 2, 3, 4, 5, 10, 100, 1000]
 
     # These expected max rates are here just to assure stability in the results. They are not assured to be correct!
-    approx_expected_max_rates = [1116990, 1359255, 1536130, 1052085, 1052085, 1052085, 1052085, 1052085]
+    approx_expected_max_rates = [1116990, 1359022, 1536130, 1052085, 1052085, 1052085, 1052085, 1052085]
```
olelod, line 197: The iteration loop stops when the change in actual rate less than 0.1% from one iteration to the next. This PR basically removes the next update here. In the time step that changed here, the change was 0.09% when the iteration stops - the change in the next step (which is now removed for consistency) is 0.017%. For all other time steps the change in actual rate when the iterations stops are much less than 0.1%, and no visible difference in the test.

TeeeJay (approved):

PR #267, opened 2023-11-01
refactor: use common consumer system dto class

jsolaas:

No need to have separate systems for pump and compressor

Creates ConsumerSystem instead of Pump- and CompressorSystem in dto.


TeeeJay (approved):
I liked that this was doable now :)

PR #268, opened 2023-11-03
Refactor/single system yaml

jsolaas:

DRY, separation of concern. Consumer system should not be pump/compressor specific

Merge COMPRESSOR_SYSTEM@v2 and PUMP_SYSTEM@v2 types into CONSUMER_SYSTEM@v2
Add TYPE to CONSUMERS in system.


src/libecalc/fixtures/cases/consumer_system_v2/data/consumer_system_v2.yaml
```diff
@@ -71,13 +71,16 @@ INSTALLATIONS:
                   CROSSOVER: [ 0, 1, 1 ]
           - NAME: pump_system_v2
             CATEGORY: PUMP
-            TYPE: PUMP_SYSTEM@v2
+            TYPE: CONSUMER_SYSTEM@v2
             CONSUMERS:
               - NAME: pump1
+                TYPE: PUMP@v2
```
olelod, line 77: Could we find the TYPE here from the ENERGY_USAGE_MODEL instead of having to define it? I mean pump_single_speed has TYPE PUMP_CHART_SINGLE_SPEED making it a system of pumps. I guess that would introduce other complications....just a thought.

jsolaas: We probably could, and it is what we are doing for v1. Setting the type makes it easier to know what the user is trying to do when the input is wrong, so it is easier to give good feedback.

jsolaas: `_get_component_type` in `input.mappers.component_mapper.py` sets the component_type for v1 components

olelod (approved):

PR #269, opened 2023-11-06
Ecalc 125 resampling when first date in resampling frequency is different from first date in global time vector

olelod:

Currently, if a reporting frequency other than data defined is chosen, the global time vector is forced to start and end at dates which is part of the requested reporting frequency (e.g. 1st of January for YEARLY). If the start date is in e.g. August 2023, the resampled time series will start 1st of January 2023, and we will start extrapolating the results, before resampling - which is not straightforward. What should be set to nan, what should be set to zero.

This PR adds the start date and the end date from the initial global time vector to the resampled time vector of any frequency. This means that cumulative volumes are kept intact, no volumes are lost. The start and end of the data defined global time vector and the resampled time vector will be the same, but the dates in between will be different.

In a LTP reporting context the volumes from the start date to the end of that year will be reported (e.g. from 1st of Aug to 31st of December), also the volumes from 1st of January the last year until the end date (e.g. 31st of July).

ECALC-125

### DIFF
--- a/src/libecalc/common/time_utils.py
+++ b/src/libecalc/common/time_utils.py
@@ -166,30 +166,60 @@ class Frequency(str, enum.Enum):
 def resample_time_steps(
     time_steps: List[datetime],
     frequency: Frequency,
-    remove_last: bool = False,
+    include_start_date: bool = True,
+    include_end_date: bool = True,
 ) -> List[datetime]:
-    """Makes a time vector, based on the first and last date in time_vector and the frequency.
+    """Makes a time vector, based on the first and last date in time_vector and the frequency
 
-    :param time_steps: The original time vector
-    :type time_steps: List[datetime]
-    :param frequency: The reporting frequency
-    :type frequency: Frequency
-    :param remove_last: Decides whether the final date should be returned
-    :type remove_last: bool
-    :return: Time vector with dates according to frequency, start and end date
-    "rtype: List[datetime]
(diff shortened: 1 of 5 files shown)


src/libecalc/common/time_utils.py
```diff
@@ -166,30 +166,60 @@ def formatstring(self) -> str:
...
+    Returns:
+        A list of dates (and possibly including the start/end dates) between the given start and end dates following
+        the requested frequency
+
+    """
+    date_range = pd.date_range(start=start, end=end, freq=frequency.value)
 
+    time_steps = [clear_time(time_step) for time_step in date_range]
+    if include_start_date:
+        time_steps = time_steps + [clear_time(start)]
```
TeeeJay, line 218: I know that we sort, but this looks better in my eyes:
```suggestion
        time_steps = [clear_time(start)] + time_steps
```

not sure what you and the rest think...

src/libecalc/dto/result/tabular_time_series.py
```diff
@@ -25,6 +25,13 @@ def to_dataframe(
         self,
         prefix: Optional[str] = None,
     ) -> pd.DataFrame:
+        """
```
TeeeJay, line 28: not much of use this docstring...

TeeeJay (approved):
Very good work. Looks correct to me :)

PR #273, opened 2023-11-09
Refactor/component graph

jsolaas:

Reusable generic Graph class

ComponentGraph is used to represent the "is part of" connections. We
also need to represent the stream connections between consumers in
a graph (as seen in system v2). Renaming graph to be able to create
a more generic Graph class that can be reused for StreamGraph.

Move common graph methods to Graph, while keeping Component specific methods in renamed ComponentGraph

### DIFF
--- a/src/libecalc/common/graph.py
+++ b/src/libecalc/common/graph.py
@@ -1,0 +1,73 @@
+from __future__ import annotations
+
+from typing import Dict, Generic, List, Protocol, TypeVar
+
+import networkx as nx
+from typing_extensions import Self
+
+NodeID = str
+
+
+class NodeWithID(Protocol):
+    @property
+    def id(self) -> NodeID:
+        ...
+
+
+TNode = TypeVar("TNode", bound=NodeWithID)
+
+
+class Graph(Generic[TNode]):
+    def __init__(self):
+        self.graph = nx.DiGraph()
+        self.nodes: Dict[NodeID, TNode] = {}
+
+    def add_node(self, node: TNode) -> Self:
+        self.graph.add_node(node.id)
+        self.nodes[node.id] = node
+        return self
+
+    def add_edge(self, from_id: NodeID, to_id: NodeID) -> Self:
+        if from_id not in self.nodes or to_id not in self.nodes:
+            raise ValueError("Add node before adding edges")
+
+        self.graph.add_edge(from_id, to_id)
+        return self
+
+    def add_subgraph(self, subgraph: Graph) -> Self:
+        self.nodes.update(subgraph.nodes)
+        self.graph = nx.compose(self.graph, subgraph.graph)
+        return self
+
+    def get_successors(self, node_id: NodeID, recursively=False) -> List[NodeID]:
+        if recursively:
+            return [
+                successor_id
+                for successor_id in nx.dfs_tree(self.graph, source=node_id).nodes()
+                if successor_id != node_id
+            ]
+        else:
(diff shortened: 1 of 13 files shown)


src/libecalc/common/graph.py
```diff
@@ -0,0 +1,66 @@
...
+class NodeWithID(Protocol):
+    @property
+    def id(self) -> NodeID:
+        ...
+
+
+TNode = TypeVar("TNode", bound=NodeWithID)
+
+
+class Graph(Generic[TNode]):
```
TeeeJay, line 19: You _may_be_ right that I want more domain specific names, but perhaps not here? Depends on how we use this graph class. If this graph class is never exposed to users, but just an internal interface because we use a "graph-like structure", then it is ok, but that the implementation (networkx) is hidden.

However, in that case this is a data structure that could be "outside" the domain, in a "helper library, such as seedwork that is used similarily as a part of the language", such as pydantic , since Python itself doesnt have a graph-like structure in its batteries (afaik)

src/libecalc/common/graph.py
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(self):
+        self.graph = nx.DiGraph()
```
TeeeJay, line 21: I like this implementation, handling the 2 different perspectives of a graph in different implementations" :) Makes it easier to deal with data I think :)

src/libecalc/common/graph.py
```diff
@@ -0,0 +1,66 @@
+from __future__ import annotations
+
+from typing import Dict, Generic, List, Protocol, TypeVar
+
+import networkx as nx
+
+NodeID = str
+
+
+class NodeWithID(Protocol):
```
TeeeJay, line 10: should probably make sure that the `id` is immutable/frozen, or even the node? but that the graph may change?

src/libecalc/common/graph.py
```diff
@@ -0,0 +1,66 @@
+from __future__ import annotations
```
TeeeJay, line 1: Important that we have good and covered tests for this structure, since it will be the backbone of eCalc

src/libecalc/dto/component_graph.py
```diff
@@ -0,0 +1,64 @@
...
+from typing import List
+
+from libecalc import dto
+from libecalc.common.component_info.component_level import ComponentLevel
+from libecalc.common.graph import Graph, NodeID
+from libecalc.dto.base import ComponentType
+from libecalc.dto.node_info import NodeInfo
+
+
+class ComponentGraph(Graph):
```
TeeeJay, line 12: Here I would want to use composition instead of inheritance, and make this "domain specific" wrt naming.

src/libecalc/core/ecalc.py
```diff
@@ -12,20 +12,20 @@
...
-from libecalc.dto.graph import Graph
+from libecalc.dto.component_graph import ComponentGraph
 from libecalc.dto.types import ConsumptionType
 
 
 class EnergyCalculator:
     def __init__(
         self,
-        graph: Graph,
+        graph: ComponentGraph,
```
TeeeJay, line 22: EnergyEquipmentNetwork? @olelod what is a good name to name the overall network (?) of energy equipment in an installation?

olelod: Process Flow Diagram perhaps

jsolaas: Can we also come up with a name for the network where the edges are streams?

PR #274, opened 2023-11-09
feat: allow expressions for pressure drop ahead of stage

frodehk:

Use case: It can be challenging to use PRESSURE_DROP_AHEAD_OF_STAGE in calibration because it can only be a fixed number, and if such a model is run with PRESSURE_DROP_AHEAD_OF_STAGE for prediction phase, the historical part must be excluded.

In general: allowing for expressions/time series in models can increase flexibility for users.

- [x] Allow time series for PRESSURE_DROP_AHEAD_OF_STAGE. Evaluating expression before sending to core.
- [x] Update core to use actual time step to read correct pressure drop
- [x] Update tests

https://equinor-ecalc.atlassian.net/browse/ECALC-371?atlOrigin=eyJpIjoiZjMxMDEzYmY4NmUxNGFkNDk3NGE1NjNjM2M5MWFjOWQiLCJwIjoiaiJ9


src/libecalc/dto/models/compressor/stage.py
```diff
@@ -10,7 +10,7 @@
 class CompressorStage(EcalcBaseModel):
     compressor_chart: CompressorChart
     inlet_temperature_kelvin: confloat(ge=0)
-    pressure_drop_before_stage: confloat(ge=0)
+    pressure_drop_before_stage: Union[confloat(ge=0), str, List[float]]
```
jsolaas, line 13: I don't like the `str` here. When in core and looking at this type you don't want that to be an option, it should probably be List[float] (if we think of this object as something that belongs to core).

The 'problem' is that we are using the same object in two different 'layers' I guess. I think we need the YamlCompressorStage class (currently not implemented) and another object that core expects. This conversion can happen from YamlCompressorStage to CompressorStage (domain class?). This is related to the same stuff we are trying to figure out for v2. I think we are leaning towards the dto-later being unnecessary, i.e. we should have a yaml-layer, and domain (core) objects.

PR #275, opened 2023-11-09
refactor: rename Stream to StreamConditions

jsolaas:
We should use StreamConditions because the Stream is defined as the from, to definition of a stream.


TeeeJay (commented):
I dont understand the intent of this..imo `Stream` is better and covers the domain correctly. Or perhaps, we should recompose the object? Does it make sense that a "StreamCondition" has a name? Is `StreamState` better? or that a `Stream` consists of `StreamCondition` and a name (and possibly a State?)

jsolaas:
The name in StreamCondition is a reference to the stream, so "this is the conditions for the stream with name"

PR #278, opened 2023-11-10
refactor: collect results in priority optimizer

jsolaas:

Don't want to deal with priority results outside PriorityOptimizer. 

This only works if we can merge results from core in a generic way. I'm thinking we should introduce ids for all objects to make sure we don't merge objects that should not be merged (i.e. merging on index, but the objects are not the same)

Collect results based on priority used before returning from optimize in PriorityOptimizer

### DIFF
--- a/src/libecalc/common/priority_optimizer.py
+++ b/src/libecalc/common/priority_optimizer.py
@@ -92,14 +92,16 @@ class PriorityOptimizer(Generic[TResult, TPriority]):
 
         We process each timestep separately.
 
+        It will default to the last priority if all settings fails
+
         Args:
             timesteps: The timesteps that we want to figure out which priority to use for.
             priorities: Dict of priorities, key is used to identify the priority in the results.
             evaluator: The evaluator function gives a list of results back, each result with its own unique id.
 
         Returns:
-            PriorityOptimizerResult: result containing priorities used and a map of the calculated results. The keys of
-                the results map are the timestep used, the priority index and the id of the result.
+            PriorityOptimizerResult: result containing priorities used and a list of the results merged on priorities
+            used,
 
         """
         is_valid = TimeSeriesBoolean(timesteps=timesteps, values=[False] * len(timesteps), unit=Unit.NONE)
--- a/src/libecalc/common/priority_optimizer.py
+++ b/src/libecalc/common/priority_optimizer.py
@@ -18,21 +18,68 @@ from libecalc.common.utils.rates import (
 TResult = TypeVar("TResult")
 TPriority = TypeVar("TPriority")
 
+ComponentID = str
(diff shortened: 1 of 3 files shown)


src/libecalc/common/priority_optimizer.py
```diff
@@ -18,21 +18,68 @@
...
 TPriority = TypeVar("TPriority")
 
+ComponentID = str
+
 
 @dataclass
 class PriorityOptimizerResult(Generic[TResult]):
     priorities_used: TimeSeriesString
-    priority_results: Dict[datetime, Dict[PriorityID, Dict[str, TResult]]]
+    priority_results: List[typing.Any]  # TODO: typing. This is the consumer results merged based on priorities used
```
TeeeJay, line 27:
```suggestion
    priority_results: List[Any]  # TODO: typing. This is the consumer results merged based on priorities used
```

?

src/libecalc/common/priority_optimizer.py
```diff
@@ -18,21 +18,68 @@
 TResult = TypeVar("TResult")
 TPriority = TypeVar("TPriority")
 
+ComponentID = str
```
TeeeJay, line 21: Good start, and hopefully this will lead to a class at some point :)

src/libecalc/common/priority_optimizer.py
```diff
@@ -18,21 +18,68 @@
...
-    id: str
+    id: ComponentID
     result: TResult
     is_valid: TimeSeriesBoolean
 
 
 class PriorityOptimizer(Generic[TResult, TPriority]):
+    @staticmethod
+    def get_component_ids(
+        priority_results: Dict[datetime, Dict[PriorityID, Dict[ComponentID, TResult]]]
```
TeeeJay, line 40: not a TemporalModel? Should create a new class for this (soon?), as it gets (too) many layers of dicts ...difficult to grasp

src/libecalc/core/consumers/consumer_system.py
```diff
@@ -106,65 +98,51 @@ def _get_stream_conditions_adjusted_for_crossover(
 
         return adjusted_stream_conditions
 
-    def evaluate(
+    def evaluator(
```
TeeeJay, line 101: methods usually have a verb, unless they return a class or an instance? Not sure why/how this is better than `evaluate`? I guess you want to differentiate since `evaluate` is taken, and since it does not inherit/implement from `baseConsumer` any more?

TeeeJay:
I agree on the reasons behind this PR, and there are many reasons why we should have IDs on objects (domain entities), since they need to be tracked... :)

TeeeJay (approved):

PR #284, opened 2023-11-15
refactor: calculate timesteps separately

jsolaas:
Process one timestep at a time in v2


src/libecalc/common/priority_optimizer.py
```diff
@@ -2,147 +2,70 @@
...
 TResult = TypeVar("TResult")
-TPriority = TypeVar("TPriority")
 
 ComponentID = str
 
 
 @dataclass
 class PriorityOptimizerResult(Generic[TResult]):
-    priorities_used: TimeSeriesString
+    priority_used: PriorityID
```
TeeeJay, line 17: <3

src/libecalc/domain/stream_conditions.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+@dataclass
+class Pressure:
+    value: float
+    unit: Unit
+
+    def __lt__(self, other: Pressure) -> bool:
+        if self.unit != other.unit:
+            raise ValueError("Unit mismatch")
```
TeeeJay, line 24: We should use domain specific exceptions, but they can perhaps extend ValueError in order to easily catch higher up?

TeeeJay (approved):

src/libecalc/core/consumers/compressor/component.py
```diff
@@ -1,147 +1,106 @@
...
         self.id = id
-        self._temporal_model = TemporalModel(
-            data={timestep: create_compressor_model(model) for timestep, model in energy_usage_model.items()}
-        )
-
+        self._compressor_model = compressor_model
         self._operational_settings: Optional[CompressorOperationalSettings] = None
 
-    def get_max_rate(self, inlet_stream: StreamConditions, target_pressure: TimeSeriesFloat) -> List[float]:
+    def get_max_rate(self, inlet_stream: StreamConditions, target_pressure: Pressure) -> float:
```
olelod, line 30: Remove "For each timestep" in docstring.

PR #287, opened 2023-11-20
Feat/support bfs tree graph

jsolaas:

- bfs_tree of a graph representing the connection between variables gives the relations in order of when the variables are used
- Exposing the yaml object for variables gives us the option to dump the variables as yaml again

### DIFF
--- a/src/libecalc/common/graph.py
+++ b/src/libecalc/common/graph.py
@@ -69,5 +69,18 @@ class Graph(Generic[TNode]):
     def sorted_node_ids(self) -> List[NodeID]:
         return list(nx.topological_sort(self.graph))
 
+    def breadth_first_search_tree(self, source_id: NodeID) -> List[NodeID]:
+        """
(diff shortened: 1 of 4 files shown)


frodehk (approved):
Looks good!

src/libecalc/common/graph.py
```diff
@@ -69,5 +69,8 @@ def get_node(self, node_id: NodeID) -> TNode:
     def sorted_node_ids(self) -> List[NodeID]:
         return list(nx.topological_sort(self.graph))
 
+    def bfs_tree(self, source_id: NodeID) -> List[NodeID]:
```
TeeeJay, line 72:
```suggestion
    def breadth_first_search_tree(self, source_id: NodeID) -> List[NodeID]:
```
Please write a small docstring of what this is, and why this is an important way of retrieving data.
I would prefer to have it as a property/method to the network/graph, as a representation in which order it must be evaluated.

TeeeJay (approved):

PR #290, opened 2023-11-21
chore: attempt to automatically upgrade to pydantic v2

TeeeJay:
a few places with a bit advanced functionality did not work. Needs to be looked into in detail.

There is still a fallback to pydantic v1 functionality through pydantic.v1 package ...


TeeeJay:
should first just import v1 from v2, and then update bit by bit :)

PR #296, opened 2023-11-22
fix: bug in compressor with turbine models with multiple streams and …

olelod:
…only one date

A 2d-array with dates and rates for different streams was squeezed into a 1d-array when there was only one date. There was a check for the compressor model being a MultipleStreamsAndPressures compressor train, but the case where a CompressorWithTurbineModel with a MultipleStreamsAndPressures compressor train was not covered.


TeeeJay (approved):
I actually have a hard time understnading the nested test/if statement. Not sure if it is me, or if it is better to add it sequentially ...

PR #300, opened 2023-11-23
refactor: change emission rate type to calendar day

frodehk:

Emission rate type should be `CALENDAR_DAY`, the same as for fuel rate.

- [x] Convert emission rate results (in dto) to `CALENDAR_DAY`
- [x] Update snapshots

https://equinor-ecalc.atlassian.net/browse/ECALC-495?atlOrigin=eyJpIjoiZGY1YTA5OThkYjc1NGVmNjgxNjIxOTNiYzAwMzBmZjAiLCJwIjoiaiJ9


src/libecalc/dto/utils/aggregators.py
```diff
@@ -48,7 +48,7 @@ def aggregate_emissions(
         emissions_aggregated[emission_name] = PartialEmissionResult(
             name=emission_name,
             timesteps=emissions[0].timesteps,
-            rate=reduce(operator.add, [emission.rate for emission in emissions]),
+            rate=reduce(operator.add, [emission.rate.to_calendar_day() for emission in emissions]),
```
TeeeJay, line 51: here the rate should already be calendar day I guess, but it doesn hurt to be sure at least ;)

TeeeJay (approved):
nice, but also here no snapshots change...probably due to regularity = 1 all over the place ...

src/libecalc/dto/utils/aggregators.py
frodehk: When I checked (I think), it seemed to be stream:-)

PR #302, opened 2023-11-24
chore(ecalc-neqsim-wrapper): update to neqsim 2.5.9

olelod:

To include the latest development in NeqSim


TeeeJay:
this is ok I think. Just go forward and remove the failing test and make a note. It is only relevant for parallell usage

PR #304, opened 2023-11-28
fix: do not merge model results

jsolaas:
Unless it's a system

Refs: ECALC-51

Missing compressor chart when combining trains/compressors

Returns all model results without using `extend` for consumers (except system)


src/tests/libecalc/dto/results/model/test_requested_pressures_compressors.py
```diff
@@ -77,17 +77,17 @@ def test_requested_pressures_compressor_system_temporal_model(result: EcalcModel
...
-    requested_inlet_pressure_train2 = get_inlet_pressure(3, date_temporal_1, models)
-    requested_inlet_pressure_train2_upgr = get_inlet_pressure(4, date_temporal_2, models)
-    requested_inlet_pressure_train3 = get_inlet_pressure(5, date_temporal_1, models)
-
-    requested_outlet_pressure_train1 = get_outlet_pressure(1, date_temporal_1, models)
-    requested_outlet_pressure_train1_upgr = get_outlet_pressure(2, date_temporal_2, models)
-    requested_outlet_pressure_train2 = get_outlet_pressure(3, date_temporal_1, models)
-    requested_outlet_pressure_train2_upgr = get_outlet_pressure(4, date_temporal_2, models)
-    requested_outlet_pressure_train3 = get_outlet_pressure(5, date_temporal_1, models)
+    requested_inlet_pressure_train1 = get_inlet_pressure(2, date_temporal_1, models)
```
TeeeJay, line 80: why change this? as a result fo the bug?

src/tests/libecalc/integration/test_all_consumer_with_time_slots_models.py
```diff
@@ -46,7 +46,7 @@ def test_time_slots_with_changing_model(time_slot_electricity_consumer_with_chan
     # train-stages are also included as CONSUMER_MODELs
     model_results = result.models
     assert isinstance(model_results, list)
-    assert len(model_results) == 3
+    assert len(model_results) == 5
```
TeeeJay, line 49: funny that we were testing and verifying on something that was obviously wrong?

TeeeJay (approved):

TeeeJay (approved):
based on results I think this looks, good. Too difficult to read the diff, but I think I understand the change . Was the initial change a bug or intentional?

src/tests/libecalc/integration/test_all_consumer_with_time_slots_models.py
jsolaas: It's not that obvious to me. A bit confusing maybe, but it makes sense to merge the results in a lot of cases. We are solving it by rewriting the yaml currently though, and that is a better approach IMO.

frodehk (approved):

PR #306, opened 2023-11-30
Ecalc 510 improve error message spelling

jsolaas:

Bad error message when reference does not exis

Error message

### DIFF
--- a/src/libecalc/common/errors/exceptions.py
+++ b/src/libecalc/common/errors/exceptions.py
@@ -62,3 +62,10 @@ class IllegalStateException(EcalcError):
 
     def __init__(self, message: str):
         super().__init__("Illegal state", message, error_type=EcalcErrorType.SERVER_ERROR)
+
+
+class InvalidReferenceException(EcalcError):
+    """The data provided is missing a required reference."""
+
+    def __init__(self, message: str):
+        super().__init__("Invalid reference", message, error_type=EcalcErrorType.CLIENT_ERROR)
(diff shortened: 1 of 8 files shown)


src/libecalc/common/errors/exceptions.py
```diff
@@ -62,3 +62,10 @@ class IllegalStateException(EcalcError):
 
     def __init__(self, message: str):
         super().__init__("Illegal state", message, error_type=EcalcErrorType.SERVER_ERROR)
+
+
+class InvalidReferenceException(EcalcError):
+    """This exception should hopefully never occur, and indicates a bug in the code."""
```
TeeeJay, line 68: really? I think we should blame the user ;)

src/libecalc/presentation/yaml/mappers/utils.py
```diff
@@ -61,16 +58,10 @@ def resolve_reference(
 
 
 def resolve_and_validate_reference(value: str, references: Dict[str, ReferenceValue]) -> ReferenceValue:
```
TeeeJay, line 60: can we replace `resolve_reference` with `resolve_and_validate_reference` everywhere? no need to specify that we validate as well, it is just a natural thing of doing (in a submethod called `validate_reference` within `resolve_reference`?

TeeeJay (approved):

src/libecalc/presentation/yaml/mappers/utils.py
jsolaas: I have replaced everywhere, so yes. Should I rename it to resolve_reference?

PR #307, opened 2023-11-30
chore: define yaml classes for remaining elements in MODELS

frodehk:

Add classes for remaining elements in MODELS, for use with V2.

- [ ] Add yaml class for `SINGLE_SPEED_COMPRESSOR_TRAIN`

https://equinor-ecalc.atlassian.net/browse/ECALC-505?atlOrigin=eyJpIjoiYzMwMGYwZTgyMDZhNGUwOTkzODAzZTliNWU0ZWNjMDAiLCJwIjoiaiJ9

### DIFF
--- a/src/libecalc/presentation/yaml/yaml_types/models/single_speed_compressor_train.py
+++ b/src/libecalc/presentation/yaml/yaml_types/models/single_speed_compressor_train.py
@@ -1,0 +1,77 @@
+import enum
+from typing import List, Literal
+
+from pydantic import Field
+
+from libecalc.presentation.yaml.yaml_types import YamlBase
+from libecalc.presentation.yaml.yaml_types.models.yaml_compressor_chart import (
+    YamlCompressorChart,
+)
+from libecalc.presentation.yaml.yaml_types.models.yaml_enums import (
+    YamlModelType,
+)
+
+
+class YamlPressureControl(enum.Enum):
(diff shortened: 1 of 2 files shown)


src/libecalc/presentation/yaml/yaml_types/models/yaml_compressor_chart.py
```diff
@@ -76,6 +76,25 @@ def to_dto(self):
...
+        YamlModelType.COMPRESSOR_CHART,
+        description="Defines the type of model. See documentation for more information.",
+        title="TYPE",
+    )
+    chart_type: Literal[YamlChartType.VARIABLE_SPEED] = YamlChartType.VARIABLE_SPEED
+    curves: YamlCurve = Field(..., description="One single compressor chart curve.", title="CURVE")
+    units: YamlUnits = None
+
+    def to_dto(self):
+        raise NotImplementedError
```
jsolaas, line 95: I think it is a mistake to define this as needed in YamlBase, as I have done currently.

TeeeJay: no, it doesnt matter that much. We want to skip `dtos` completely, and map directly to domain. dtos are no longer needed, but we will need some clever ways to map to domain. WIP :)

PR #308, opened 2023-12-01
feat: init train v2

jsolaas:

Initialize the work on a common train, where the features of all current train-models should be handled.

### DIFF
--- a/src/libecalc/common/graph.py
+++ b/src/libecalc/common/graph.py
@@ -27,11 +27,11 @@ class Graph(Generic[TNode]):
         self.nodes[node.id] = node
         return self
 
-    def add_edge(self, from_id: NodeID, to_id: NodeID) -> Self:
+    def add_edge(self, from_id: NodeID, to_id: NodeID, **kwargs) -> Self:
         if from_id not in self.nodes or to_id not in self.nodes:
             raise ValueError("Add node before adding edges")
 
-        self.graph.add_edge(from_id, to_id)
+        self.graph.add_edge(from_id, to_id, **kwargs)
         return self
 
     def add_subgraph(self, subgraph: Graph) -> Self:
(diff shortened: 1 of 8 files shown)


src/libecalc/common/graph.py
```diff
@@ -27,11 +27,11 @@ def add_node(self, node: TNode) -> Self:
         self.nodes[node.id] = node
         return self
 
-    def add_edge(self, from_id: NodeID, to_id: NodeID) -> Self:
+    def add_edge(self, from_id: NodeID, to_id: NodeID, **kwargs) -> Self:
```
TeeeJay, line 30: docstring? intention behind kwargs ... different properties based on type of edge etc?

src/libecalc/common/graph.py
```diff
@@ -49,6 +49,15 @@ def get_successors(self, node_id: NodeID, recursively=False) -> List[NodeID]:
         else:
             return list(self.graph.successors(node_id))
 
+    def get_successor(self, node_id: NodeID) -> NodeID:
```
TeeeJay, line 52: Current limitation that we want to have now, or always, to use "crossovers" when there are more than 1?

src/libecalc/common/graph.py
```diff
@@ -58,6 +67,10 @@ def get_predecessor(self, node_id: NodeID) -> NodeID:
             )
         return predecessors[0]
 
+    def get_predecessors(self, node_id: NodeID) -> List[NodeID]:
```
TeeeJay, line 70: why > 1 here?

src/libecalc/common/graph.py
jsolaas: Not always, currently it's used when it's supposed to only be one, but we could of course just use get_successors()[0]. I was thinking we could have both get_successor and get_successors. I think we already have both for either successors or predecessors.

PR #311, opened 2023-12-05
fix: deep copy when aggregating model results

jsolaas:
Bug where the first model result became the aggregated result because we were mutating the first model result instead of creating a new/copying.


TeeeJay (approved):

src/tests/libecalc/integration/snapshots/test_all_energy_usage_models/test_all_results/all_energy_usage_models_v3.json
```diff
@@ -22145,48 +22145,34 @@
                 {
                     "energy_usage": {
                         "timesteps": [
-                            "2018-01-01 00:00:00",
-                            "2019-01-01 00:00:00",
-                            "2020-01-01 00:00:00"
+                            "2018-01-01 00:00:00"
```
jsolaas, line 22148: Probably easiest to see the fix here. This is the first model result which was previously mutated to the aggregated result, now it's only defined for 2018, and the next model result continues in 2019.

TeeeJay: Yes, now i understand :)

PR #312, opened 2023-12-05
feat: add emitter rate type to venting emitters

frodehk:

Should be possible to specify input rate type for `DIRECT`/`VENTING_EMITTERS`. These emitters are normally set up with calendar day input in `FUELCONSUMERS` and it should be possible to do the same when moving them to DIRECT/VENTING_EMITTERS.

- [x] Create and use Yaml-classes for venting emitters
- [x] Add `RATE_TYPE` and `UNIT` to venting emitters
- [x] Remove fuel dependency for venting emitters, i.e. where emission rate previously has been modelled as a proxy for fuel rate.
- [x] Convert output to `t/d`
- [x] Convert to stream days
- [x] Remove venting emitters from core
- [x] Remove `EmitterModelMapper` and`VentingEmittersMapper`
- [x] Regularity and variables map sent from installation to yaml-classes, in `EnergyCalculator`
- [x] Update tests and dtos
- [x] Update snapshots
- [x] Add test for venting emitters ltp export

https://equinor-ecalc.atlassian.net/browse/ECALC-524?atlOrigin=eyJpIjoiZjhjYzViMDJkMjM0NGEyMWI1YmRjZGY2MWQwMTlkYzEiLCJwIjoiaiJ9


src/libecalc/core/consumers/venting_emitter.py
```diff
@@ -58,19 +51,64 @@ def evaluate(
...
+
+        for period, model in TemporalModel(emitter_model).items():
+            if model.convert_to_stream_day:
+                start_index, end_index = period.get_timestep_indices(variables_map.time_vector)
+                regularity_this_period = regularity[start_index:end_index]
+                emission_rate[start_index:end_index] = Rates.to_stream_day(
+                    emission_rate[start_index:end_index], regularity=list(regularity_this_period)
+                )
+
+        emission_rate_kg_per_day = emission_rate
```
TeeeJay, line 99: always kg? stated in docs?

src/libecalc/dto/components.py
```diff
@@ -137,7 +138,8 @@ class EmitterModel(EcalcBaseModel):
     emission_rate: Expression
 
     regularity: Dict[datetime, Expression]
-
+    emission_rate_type: RateType = RateType.STREAM_DAY
+    convert_to_stream_day: bool = False
```
TeeeJay, line 142: why is this needed?

TeeeJay: we should always convert to stream day...or?

src/libecalc/presentation/yaml/mappers/component_mapper.py
```diff
@@ -241,13 +242,16 @@ def create_model(model: Dict, regularity: Dict[datetime, Expression]):
         name = model.get(EcalcYamlKeywords.name, "")
         user_defined_category = model.get(EcalcYamlKeywords.user_defined_tag, "")
         emission_rate = model.get(EcalcYamlKeywords.installation_venting_emitter_emission_rate)
-
+        emission_rate_type = model.get(EcalcYamlKeywords.venting_emitter_rate_type) or RateType.STREAM_DAY
+        convert_to_stream_day = emission_rate_type == RateType.CALENDAR_DAY
```
TeeeJay, line 246: just do:

`emission_rate.to_stream_day()` ?

TeeeJay:
Very nice initiative! Great observation!

Does it make sense to think of this in terms of v2 as well? How would that be?

Should be quite straight forward...lets discuss tomorrow ...

src/libecalc/dto/components.py
frodehk: Not if the user has given stream day as input.

TeeeJay: no, but the "stream day timeseries" also has a `to_stream_day()` and basically returns the same value without conversion...ie the correct timeseries should be created based on the rate_type ...

frodehk: I was looking for that, but could not find it.

src/libecalc/presentation/yaml/mappers/component_mapper.py
frodehk: At this stage, the emission_rate is e.g. a string (refering to e.g. a time series rate in input file). Do you think we should set up and evaluate expressions here, and convert to stream day?

src/libecalc/core/consumers/venting_emitter.py
```diff
@@ -2,75 +2,67 @@
...
 from libecalc.common.temporal_model import TemporalExpression, TemporalModel
-from libecalc.core.models.fuel import FuelModel
+from libecalc.common.units import Unit
+from libecalc.common.utils.rates import TimeSeriesStreamDayRate
 from libecalc.core.result.emission import EmissionResult
-from libecalc.dto import Emission
-from libecalc.dto.types import FuelType
 from libecalc.dto.variables import VariablesMap
 from libecalc.expression import Expression
+from libecalc.presentation.yaml.yaml_types.emitters.yaml_venting_emitter import (
```
jsolaas, line 12: This import is the wrong way, yaml should 'know' about core, not the other way around.

We are moving towards core not importing from anything else than libecalc.common, it is not the case currently.

We also want to get rid of expressions in core.

Considering what's left in this class if we do that, I suggest just removing this class and put the logic in EnergyCalculator instead. Some of the expression conversion could be done in the yaml model, i.e. yaml_model.get_emission_rate(variables_map) -> returns a TImeSeriesStreamDayRate. Then we can call to_unit(Units.tons) and create an EmissionResult.

src/libecalc/presentation/yaml/mappers/component_mapper.py
```diff
@@ -229,37 +233,41 @@ def from_yaml_to_dto(
...
+#         self, data: Optional[Dict], regularity: Dict[datetime, Expression]
+#     ) -> Dict[datetime, dto.EmitterModel]:
+#         time_adjusted_model = define_time_model_for_period(data, target_period=self._target_period)
+#         return {
+#             start_date: EmitterModelMapper.create_model(model, regularity=regularity)
+#             for start_date, model in time_adjusted_model.items()
+#         }
 
 
 class VentingEmittersMapper:
```
jsolaas, line 273: Do we need a mapper?

src/libecalc/presentation/yaml/yaml_types/emitters/yaml_venting_emitter.py
```diff
@@ -1,36 +1,139 @@
...
+    def get_model(
+        self, data: Optional[Dict], regularity: Dict[YamlDefaultDatetime, Expression]
+    ) -> Dict[YamlDefaultDatetime, YamlEmitterModel]:
+        time_adjusted_model = define_time_model_for_period(data, target_period=self._target_period)
+        return {
+            YamlDefaultDatetime(
+                start_date.year, start_date.month, start_date.day, start_date.hour, start_date.second
+            ): self.create_model(model, regularity=regularity)
+            for start_date, model in time_adjusted_model.items()
+        }
```
jsolaas, line 91: Do we need this?

jsolaas: The `YamlTemporalEmitterModel`, the full code-block was what I meant

frodehk: Will have a look.

src/libecalc/core/consumers/venting_emitter.py
frodehk: This venting emitter class in core is removed. Expressions are evaluated in `YamlVentingEmitter` class. A method is generated to return `TimeSeriesStreamDayRate`.

src/libecalc/presentation/yaml/mappers/component_mapper.py
frodehk: If you could have a look again, several changes. See if some of the `VentingEmittersMapper` should be moved? Now some of the "parsing" of the temporal emitter model is done here (earlier in the `YamlVentingEmitter` class)

src/libecalc/presentation/yaml/mappers/component_mapper.py
frodehk: This is now done in the `YamlVentingEmitter.get_emission_rate(variables_map)` -> `TimeSeriesStreamDayRate`

src/libecalc/core/consumers/venting_emitter.py
frodehk: This will be solved in another PR. Allowing for input units for venting emitters. Converting to kg per day before core, converting to t/d for results.

src/libecalc/presentation/yaml/mappers/component_mapper.py
frodehk: The mapper is removed

src/libecalc/fixtures/cases/all_energy_usage_models/data/all_energy_usage_models.yaml
```diff
@@ -615,7 +615,10 @@ INSTALLATIONS:
...
-        EMISSION_NAME: CH4
         CATEGORY: COLD-VENTING-FUGITIVE
-        EMITTER_MODEL:
-          EMISSION_RATE: FLARE;METHANE_RATE
+        EMISSION:
+          NAME: CH4
+          RATE:
+            VALUE: FLARE;METHANE_RATE
+            UNIT: kg/d
+            RATE_TYPE: STREAM_DAY
```
jsolaas, line 624: Do we need `RATE_` first?

src/libecalc/presentation/yaml/mappers/component_mapper.py
```diff
@@ -327,13 +266,29 @@ def from_yaml_to_dto(self, data: Dict) -> dto.Installation:
...
+                        emission=YamlVentingEmission(
+                            name=emission.get(EcalcYamlKeywords.name),
+                            rate=YamlRate(
+                                value=emission_rate.get(EcalcYamlKeywords.value),
+                                unit=emission_rate.get(EcalcYamlKeywords.unit),
+                                rate_type=emission_rate.get(EcalcYamlKeywords.rate_type),
+                            ),
+                        ),
+                        user_defined_category=venting_emitter.get(EcalcYamlKeywords.user_defined_tag),
+                    )
```
jsolaas, line 288:
```suggestion
                    YamlVentingEmitter(
                        **venting_emitter
                    )
```

We should be able to do this?

src/tests/libecalc/core/consumers/test_direct_emitter.py
```diff
@@ -1,27 +1,48 @@
 from datetime import datetime
 
 import pytest
-from libecalc import dto
-from libecalc.core.consumers.venting_emitter import VentingEmitter
+from libecalc.common.units import Unit
```
jsolaas, line 4: This test should probably be moved. presentation.yaml.test_venting_emitter.py?

src/tests/libecalc/dto/test_categories.py
```diff
@@ -3,96 +3,91 @@
...
-                user_defined_category={datetime(2000, 1, 1): "fuel-gas"},
+                emission=emission,
+                user_defined_category="fuel-gas",
             )
 
         exception: ValidationError = typing.cast(ValidationError, exc_info.value)
         assert (
             exception.errors()[0]["msg"]
-            == "CATEGORY: fuel-gas is not allowed for VentingEmitter with the name test. Valid categories are: ['BASE-LOAD', 'COLD-VENTING-FUGITIVE', 'COMPRESSOR', 'FIXED-PRODUCTION-LOAD', 'FLARE', 'MISCELLANEOUS', 'PUMP', 'GAS-DRIVEN-COMPRESSOR', 'TURBINE-GENERATOR', 'POWER-FROM-SHORE', 'OFFSHORE-WIND', 'LOADING', 'STORAGE', 'STEAM-TURBINE-GENERATOR', 'BOILER', 'HEATER']"
+            == "CATEGORY: fuel-gas is not allowed for YamlVentingEmitter with the name test. Valid categories are: ['BASE-LOAD', 'COLD-VENTING-FUGITIVE', 'COMPRESSOR', 'FIXED-PRODUCTION-LOAD', 'FLARE', 'MISCELLANEOUS', 'PUMP', 'GAS-DRIVEN-COMPRESSOR', 'TURBINE-GENERATOR', 'POWER-FROM-SHORE', 'OFFSHORE-WIND', 'LOADING', 'STORAGE', 'STEAM-TURBINE-GENERATOR', 'BOILER', 'HEATER']"
```
jsolaas, line 53: Is there a way we could keep the VentingEmitter name here? Are we using `type`when creating the error? In pydantic we can specify 
```
class Config:
  title = "VentingEmitter"
```
to manipulate the title when generating json-schema, but not sure how we can use that name when generating errors.

src/tests/libecalc/input/validation/snapshots/test_validation_json_schemas/test_json_schema_changed/schemas.json
```diff
@@ -1494,55 +1506,70 @@
...
+                        "NAME",
+                        "RATE"
+                    ],
+                    "title": "YamlVentingEmission",
+                    "type": "object"
+                },
+                "YamlVentingEmitter": {
+                    "additionalProperties": false,
+                    "properties": {
+                        "COMPONENT_TYPE": {
```
jsolaas, line 1537: This is not something the user should provide, right? If we are going to have a type in yaml, it should be called `TYPE`?

src/libecalc/fixtures/cases/all_energy_usage_models/data/all_energy_usage_models.yaml
frodehk: No, agreed.

src/libecalc/presentation/yaml/mappers/component_mapper.py
frodehk: I will check!

src/libecalc/fixtures/cases/all_energy_usage_models/data/all_energy_usage_models.yaml
frodehk: `RATE_` is removed.

src/tests/libecalc/core/consumers/test_direct_emitter.py
frodehk: Moved

src/tests/libecalc/input/validation/snapshots/test_validation_json_schemas/test_json_schema_changed/schemas.json
frodehk: The user should not provide this. Compoent type is just set internally.

src/libecalc/presentation/yaml/mappers/component_mapper.py
frodehk: Done. It works, but need to convert `CATEGORY` to `user_defined_category` for matching to work. This is a workaround to avoid renaming user_defined_category to category all places. This should probably be done later though.

frodehk: Changed it, now user_defined_category is implemented as a property in YamlVentingEmitter.

src/tests/libecalc/dto/test_categories.py
frodehk: This is fixed now, the title is used when generating the error.

src/libecalc/common/units.py
```diff
@@ -108,6 +108,7 @@ def _unit_registry() -> Dict[Unit, Dict[Unit, Callable]]:
 
         unit_registry[Unit.TONS_PER_DAY][Unit.KILO_PER_DAY] = lambda a: a * 1000
         unit_registry[Unit.KILO_PER_DAY][Unit.TONS_PER_DAY] = lambda a: a / 1000
+        unit_registry[Unit.KILO_PER_DAY][Unit.KILO_PER_DAY] = lambda a: a
```
TeeeJay, line 111: I assume that this is to avoid having to check when we "accidentally" have the same as we are trying to convert to? I guess this would be the same for "any" unit, in different situations, so why not just add this supprt before we convert, check if the units are the same, then just return the same value?

TeeeJay: actualyl, I was going to check the code, and I now see that there is a `lambda a:a` if the to and from unit is the same, so I think this line here can be skipped :)

src/libecalc/core/ecalc.py
```diff
@@ -16,12 +16,14 @@
 from libecalc.core.consumers.factory import create_consumer
 from libecalc.core.consumers.generator_set import Genset
 from libecalc.core.consumers.legacy_consumer.component import Consumer
-from libecalc.core.consumers.venting_emitter import VentingEmitter
 from libecalc.core.models.fuel import FuelModel
 from libecalc.core.result import ComponentResult, EcalcModelResult
 from libecalc.core.result.emission import EmissionResult
 from libecalc.dto.component_graph import ComponentGraph
 from libecalc.dto.types import ConsumptionType
+from libecalc.presentation.yaml.yaml_types.emitters.yaml_venting_emitter import (
```
TeeeJay, line 24: so; I see that the dto is skipped, which is good, but in that case the yaml should be converted to the domain object directly, instead of referring to the yaml object in core... as a rule, core should only depend on core (and common), not presentation, application or infrastructure layers...

src/libecalc/dto/components.py
```diff
@@ -335,7 +319,7 @@ class Installation(BaseComponent):
     user_defined_category: Optional[InstallationUserDefinedCategoryType] = None
     hydrocarbon_export: Dict[datetime, Expression]
     fuel_consumers: List[Union[GeneratorSet, FuelConsumer, ConsumerSystem]] = Field(default_factory=list)
-    venting_emitters: List[VentingEmitter] = Field(default_factory=list)
+    venting_emitters: List[YamlVentingEmitter] = Field(default_factory=list)
```
TeeeJay, line 322: Here, it should be also the domain objcet `VentingEmitter`, and not the "yaml version". It may be that there is sth that I am missing, ie. that this is one step on the way, and a reason to keep it like this temporarly, but I dont see that looking at the code :)

src/libecalc/presentation/yaml/mappers/component_mapper.py
```diff
@@ -327,13 +264,13 @@ def from_yaml_to_dto(self, data: Dict) -> dto.Installation:
...
-                venting_emitters,
-                regularity=regularity,
-            )
-            for venting_emitters in data.get(EcalcYamlKeywords.installation_venting_emitters, [])
-        ]
+
+        venting_emitters = []
+        for venting_emitter in data.get(EcalcYamlKeywords.installation_venting_emitters, []):
+            try:
+                venting_emitters.append(YamlVentingEmitter(**venting_emitter))
```
TeeeJay, line 271: so, this is where I basically expect the `YamlVentingEmitter(**venting_emitter).to_domain()`, if that is possible to do in one go, or if factories etc are needed, to do it in successive steps. The yamlventingemitter is just sth we want to validate the yaml (syntax), and parse the data, before mapping to our internal domain model(s)...

src/libecalc/presentation/yaml/yaml_types/emitters/yaml_venting_emitter.py
```diff
@@ -1,36 +1,104 @@
...
+            variables_map=variables_map,
+        )
+
+        emission_rate = (
+            Expression.setup_from_expression(value=self.emission.rate.value)
+            .evaluate(variables=variables_map.variables, fill_length=len(variables_map.time_vector))
+            .tolist()
+        )
+
+        # emission_rate = Unit.to(self.emission.rate.unit, Unit.KILO_PER_DAY)(emission_rate).tolist()
```
TeeeJay, line 93: remove? or keep?

TeeeJay (changes requested):
I really like to see how thorough you are with fixtures, tests, and thinking of all aspects :)

however, I see that we have different ideas/thoughts of the mapping and which "type of objects to use in different circumstances". Lets discuss. There may be things that I am not seeing, and I know that it is not necessarily easy to follow the clean archicture rules, given the existing code ...

src/libecalc/common/units.py
frodehk: Agree, should be skipped:-)

src/libecalc/core/ecalc.py
frodehk: When I removed the venting emitter from core, the remaining logic was put in the presentation layer (yaml class). Basically only evaluating expressions and handling stream days/calendar days (`get_emission_rate`). It may be possible to argue that there is no domain specific logic for venting emitters (except for rate type check). We may need a domain object, regardless of how much it does?

TeeeJay: Good point!

I think yes. 3 arguments from top of my head; 1. we may want to use the domain object directly, when using the libeCalc as a (library) API. 2. the domain may grow, and it should grow naturally from the e.g. basic logic we initially have, instead of having to change on a lot of the rest of the structure. 3. when the dependencies are "tangled", it gets much more difficult to change the code, side effects etc. My 3 cents :p

src/libecalc/presentation/yaml/mappers/component_mapper.py
frodehk: Are you sure we should do it here? If we are to avoid evaluating expressions in core/domain, we need to give a rate that is evaluated to domain. This means YamlVentingEmitter(**venting_emitter).to_dto(variables_map, regularity). Do we want to send variables_map to component_mapper?

TeeeJay: No! :) And that is why I gave an answer with reservations based on what is needed to be done. I think we should do the conversion of regularity and variables to what the core should have, BEFORE we convert to domain models. This can happen e.g. in a `factory` or request a `service` to do the conversion, as a part of the `to_domain()` ...

TeeeJay: In general I do not think to_domain should have any parameters

src/libecalc/dto/components.py
frodehk: This also means that we need to send `variables_map` into `component_mapper`, so probably we will keep the `YamlVentingEmitter` here for now?

src/libecalc/common/units.py
frodehk: Skipped now

src/libecalc/presentation/yaml/yaml_types/emitters/yaml_venting_emitter.py
frodehk: I think it should be removed. But keep the commented line for a little while, to be really sure:

1. There is no core calculations for venting emitters. So I cannot see a need to convert to `kg/d`.
2. The only thing I can think of, is aggregation - but that is done at a later stage, when emissions should be in `t/d`, I think?

src/libecalc/dto/components.py
frodehk: As discussed, for now we keep the approach "from yaml to result".

jsolaas (approved):

TeeeJay (approved):

src/tests/libecalc/presentation/yaml/test_venting_emitter.py
```diff
@@ -0,0 +1,48 @@
...
+from libecalc.dto.base import ConsumerUserDefinedCategoryType
+from libecalc.expression import Expression
+from libecalc.presentation.yaml.yaml_types.emitters.yaml_venting_emitter import (
+    YamlVentingEmission,
+    YamlVentingEmitter,
+)
+from libecalc.presentation.yaml.yaml_types.yaml_stream_conditions import YamlRate
+
+
+def test_venting_emitter(variables_map):
```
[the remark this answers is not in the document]
frodehk: Done

PR #318, opened 2023-12-11
refactor: move simple result to presentation

jsolaas:
Also reverse the imports/dependency

### DIFF
--- a/src/libecalc/presentation/simple_result/simple.py
+++ b/src/libecalc/presentation/simple_result/simple.py
@@ -6,9 +6,11 @@ from pydantic import Extra, root_validator
 from scipy.interpolate import interp1d
 
 from libecalc.common.component_info.component_level import ComponentLevel
+from libecalc.common.errors.exceptions import ProgrammingError
 from libecalc.common.logger import logger
 from libecalc.common.units import Unit
 from libecalc.dto.base import ComponentType
+from libecalc.dto.result import ComponentResult, EcalcModelResult
 from libecalc.dto.result.base import EcalcResultBaseModel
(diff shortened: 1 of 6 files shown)


src/libecalc/presentation/simple_result/simple.py
```diff
@@ -133,6 +134,10 @@ class SimpleComponentResult(SimpleBase):
     energy_usage_unit: Unit
     power: Optional[List[opt_float]]
 
+    @classmethod
+    def from_dto(cls, component_result) -> "SimpleComponentResult":
```
TeeeJay, line 138: typehint

src/tests/libecalc/dto/results/test_simple.py
```diff
@@ -5,7 +5,7 @@
 from libecalc.common.component_info.component_level import ComponentLevel
 from libecalc.common.units import Unit
 from libecalc.dto.base import ComponentType
-from libecalc.dto.result.simple import (
+from libecalc.presentation.simple_result.simple import (
```
TeeeJay, line 8: this test should be moved to tests/libecalc/presentation? I know I havent been consistent and updated tests in general based on the rules we want to adhere to, just to check if you thought of that

src/libecalc/presentation/simple_result/simple.py
jsolaas: So a protocol then? BaseModel from pydantic? Import type from dto.results?

jsolaas: Nevermind, I was thinking simple shouldn't import from dto, but that is what I'm fixing here..

PR #319, opened 2023-12-12
fix: support mismatching timesteps in ltp delta profile

jsolaas:
Refs: https://equinor-ecalc.atlassian.net/browse/ECALC-64

### DIFF
--- a/src/libecalc/presentation/exporter/dto/dtos.py
+++ b/src/libecalc/presentation/exporter/dto/dtos.py
@@ -1,7 +1,7 @@
 from collections import defaultdict
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Dict, List, Type, Union
+from typing import Dict, List, Set, Type, Union
 
 import pandas as pd
 from pydantic import Field, root_validator
@@ -10,6 +10,7 @@ from libecalc.common.errors.exceptions import (
     DifferentLengthsError,
     IncompatibleDataError,
     MissingKeyError,
+    ProgrammingError,
 )
 from libecalc.common.math.math_utils import MathUtil
 from libecalc.common.time_utils import Frequency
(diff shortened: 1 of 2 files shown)


src/libecalc/presentation/exporter/dto/dtos.py
```diff
@@ -100,6 +116,18 @@ def check_equal_lengths_of_timeseries(cls, values):
 
         return values
 
+    def fit_to_timesteps(self, timesteps: List[datetime]) -> "TSVPrognosis":
+        return TSVPrognosis(
+            time_steps=TimeSteps(
+                values=timesteps,
+                frequency=self.time_steps.frequency,  # Should this be provided?
```
jsolaas, line 123: What's the frequency here @TeeeJay? Just output frequency used for the run? Is it needed?

TeeeJay: https://en.wikipedia.org/wiki/What%27s_the_Frequency,_Kenneth%3F

jsolaas: nice

src/libecalc/presentation/exporter/dto/dtos.py
```diff
@@ -69,6 +69,22 @@ class TimeSeries(EcalcResultBaseModel):
     title: str
     unit: Unit
 
+    def fit_to_timesteps(self, timesteps: List[datetime]) -> "TimeSeries":
+        """
+        Fit TimeSeries to timesteps, in this case it is expected that the provided timesteps is a subset of the current
```
TeeeJay, line 74: I think we should check that is actually is a subset?

src/libecalc/presentation/exporter/dto/dtos.py
TeeeJay: but yes, I think it is just "metainfo" , but possibly used to know what to expect from the timeseries interval. shouldnt be needed now i think, as we can change to whatever we want, irrelevant of original frequency

src/libecalc/presentation/exporter/dto/dtos.py
```diff
@@ -178,12 +206,32 @@ class AssetTSVPrognosis(EcalcResultBaseModel):
 
     tsv_prognoses: Dict[str, TSVPrognosis] = Field(default_factory=dict)
 
+    @property
+    def timesteps(self) -> Set[datetime]:
```
TeeeJay, line 210:
```suggestion
    def common_timesteps(self) -> Set[datetime]:
```
? vs union, if that is suddenly needed (call it sth with intersection?)

TeeeJay (approved):

src/libecalc/presentation/exporter/dto/dtos.py
jsolaas: Yes, I wasn't sure, as all the timesteps should be the same for each installation right?

TeeeJay: not sure if it has or should, as I have experienced mismatches between mobile and fixed before ...

PR #328, opened 2023-12-20
refactor: use time series collection yaml classes

jsolaas:
Validate when getting time_series from yaml model.

### DIFF
--- a/src/libecalc/presentation/yaml/mappers/variables_mapper/variables_mapper.py
+++ b/src/libecalc/presentation/yaml/mappers/variables_mapper/variables_mapper.py
@@ -112,7 +112,8 @@ def map_yaml_to_variables(
     result_options: dto.ResultOptions,
 ) -> dto.VariablesMap:
     timeseries_collections = [
-        TimeSeriesCollectionMapper(resources).from_yaml_to_dto(timeseries) for timeseries in configuration.time_series
+        TimeSeriesCollectionMapper(resources).from_yaml_to_dto(timeseries.dict(by_alias=True))
+        for timeseries in configuration.time_series
     ]
 
     global_time_vector = get_global_time_vector(
--- a/src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
+++ b/src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
@@ -5,6 +5,10 @@ from typing import Any, Dict, Iterator, List, Optional, TextIO, Type, Union
 import yaml
 from yaml import SafeLoader
 
+from libecalc.presentation.yaml.yaml_types.time_series.yaml_time_series import (
+    YamlTimeSeriesCollection,
+)
+
 try:
     from pydantic.v1 import parse_obj_as
 except ImportError:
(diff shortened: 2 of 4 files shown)


src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
```diff
@@ -269,8 +273,15 @@ def facility_inputs(self):
         return self._internal_datamodel.get(EcalcYamlKeywords.facility_inputs, [])
 
     @property
-    def time_series(self):
-        return self._internal_datamodel.get(EcalcYamlKeywords.time_series, [])
+    def time_series(self) -> List[YamlTimeSeriesCollection]:
```
jsolaas, line 276: Should do this for all values in yaml model IMO. Doing it here when the properties are used makes it possible to work with the yaml file without everything being valid

src/libecalc/presentation/yaml/yaml_types/time_series/yaml_time_series.py
```diff
@@ -11,6 +13,7 @@
 class YamlTimeSeriesCollectionBase(YamlBase):
     class Config:
         title = "TimeSeries"
+        use_enum_values = True
```
jsolaas, line 16: Makes enum a str when calling dict()

src/libecalc/presentation/yaml/mappers/variables_mapper/variables_mapper.py
```diff
@@ -112,7 +112,8 @@ def map_yaml_to_variables(
     result_options: dto.ResultOptions,
 ) -> dto.VariablesMap:
     timeseries_collections = [
-        TimeSeriesCollectionMapper(resources).from_yaml_to_dto(timeseries) for timeseries in configuration.time_series
+        TimeSeriesCollectionMapper(resources).from_yaml_to_dto(timeseries.dict(by_alias=True))
```
jsolaas, line 115: We should look into removing this mapper and the classes, should be able to handle it in the yaml classes?

Converting back to dict as that is expected in TimeSeriesCollectionMapper

TeeeJay: you mean to implicitly and automatically handle the conversion?

jsolaas: https://equinor-ecalc.atlassian.net/browse/ECALC-586 Explained a bit here. 
TimeSeriesCollectionMapper should not need to deal with validation of the yaml at least, it could be simplified to only deal with reading the resources and creating something that can easily be added to variables map

src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
```diff
@@ -269,8 +273,15 @@ def facility_inputs(self):
         return self._internal_datamodel.get(EcalcYamlKeywords.facility_inputs, [])
 
     @property
-    def time_series(self):
-        return self._internal_datamodel.get(EcalcYamlKeywords.time_series, [])
```
TeeeJay, line 273: what was returned here?  a "string dict"?

src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
TeeeJay: so, to only actually parse accordign to schema when we use a property?

TeeeJay (approved):

src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
jsolaas: Previously it was a list of dicts, each dict was a time series entry from yaml

src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
jsolaas: yes, we can have a separate validate function if we need to make sure the full model is valid.

jsolaas: We could also require the model to be valid. 

It would be nice if this class took care of it though. This class would then read the model, then we can get the properties in the model as yaml classes, removing the need to validate a lot of stuff in other places.

PR #347, opened 2024-01-15
refactor: remove need for id on components etc

TeeeJay:

- [x] Considered to add tests

Id and name is being used for different purposes, and it might be difficutl to know when to use what. in particular now as name should be unique everywhere. no need for id at the moment.

make id work as name, almost.


src/libecalc/common/string/string_utils.py
```diff
@@ -14,13 +13,18 @@ def get_duplicates(names: Iterable[str]) -> Set[str]:
 
 
 def generate_id(*args: str) -> str:
-    """Generate an id from one or more strings. The string is encoded to avoid it being used to get other info than
+    """
+    Deprecated: When names were not unique, this was necessary in order to make names unique based on context/hierarchy. Now names should
+    be unique for any part of the eCalc model that supports names, and it should therefore not be needed any more.
```
jsolaas, line 18: Not any part, we use name inside models also, i.e. consumer system v1, which aren't unique across temporal systems

jsolaas (approved):

PR #348, opened 2024-01-15
refactor: pump v2 support for single speed

TeeeJay:

- [x] Considered to add tests

### DIFF
--- a/src/libecalc/application/graph_result.py
+++ b/src/libecalc/application/graph_result.py
@@ -162,6 +162,7 @@ class GraphResult:
                 [
                     emission_dto_results[fuel_consumer_id]
                     for fuel_consumer_id in self.graph.get_successors(installation.id)
+                    if fuel_consumer_id in emission_dto_results
                 ]
             )
 
@@ -828,7 +829,11 @@ class GraphResult:
                             )
                         ]
                     )
-            elif consumer_node_info.component_type in [ComponentType.PUMP, ComponentType.PUMP_SYSTEM]:
+            elif consumer_node_info.component_type in [
+                ComponentType.PUMP,
+                ComponentType.PUMP_SYSTEM,
+                ComponentType.PUMP_V2,
+            ]:
                 component = self.graph.get_node(consumer_id)
                 for model in consumer_result.models:
                     models.extend(
@@ -970,12 +975,12 @@ class GraphResult:
                     id=consumer_result.component_result.id,
                     is_valid=consumer_result.component_result.is_valid,
                 )
-            elif consumer_node_info.component_type == ComponentType.PUMP:
+            elif consumer_node_info.component_type in [ComponentType.PUMP, ComponentType.PUMP_V2]:
                 obj = dto.result.results.PumpResult(
                     name=consumer_node_info.name,
                     parent=self.graph.get_predecessor(consumer_id),
                     component_level=consumer_node_info.component_level,
-                    componentType=consumer_node_info.component_type,
+                    componentType=ComponentType.PUMP,  # TODO: Since v1 and v2 currently has same structure, we specify as v1 to avoid having to update Web
                     emissions=self._parse_emissions(self.emission_results[consumer_id], regularity)
                     if consumer_id in self.emission_results
                     else [],
@@ -1211,7 +1216,8 @@ class GraphResult:
                     },
                 )
 
-            sub_components.append(obj)
+            if obj:
+                sub_components.append(obj)
 
         for installation in asset.installations:
             regularity = regularities[installation.id]  # Already evaluated regularities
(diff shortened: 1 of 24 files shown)


TeeeJay:
rebased etc. will go through a final check

src/libecalc/application/graph_result.py
```diff
@@ -1211,7 +1216,8 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
                     },
                 )
 
-            sub_components.append(obj)
+            if obj:
```
jsolaas, line 1219: When is it None?

src/libecalc/core/result/base.py
```diff
@@ -33,7 +33,16 @@ def extend(self, other: Self) -> Self:
...
-                if isinstance(other_values, list):
+                # Temporary v2 only
+                # We are not able to currently import PumpModelResult due to circular import, therefore we use duck typing...
+                if (
+                    len(values) == 1
+                    and len(other_values) == 1
+                    and hasattr(values[0], "extend")
+                    and hasattr(other_values[0], "extend")
+                ):
+                    values[0].extend(other_values[0])
```
jsolaas, line 44: Why was this added?

src/libecalc/dto/components.py
```diff
@@ -293,7 +295,10 @@ class GeneratorSet(BaseEquipment):
     component_type: Literal[ComponentType.GENERATOR_SET] = ComponentType.GENERATOR_SET
     fuel: Dict[datetime, FuelType]
     generator_set_model: Dict[datetime, GeneratorSetSampled]
-    consumers: List[Union[ElectricityConsumer, ConsumerSystem]] = Field(default_factory=list)
+    consumers: List[Union[ElectricityConsumer, ConsumerSystem, Any]] = Field(
+        default_factory=list
+    )  # Any here is Pump, that cannot be explicitly specified due to circular import ...temporalequipment?
+    # Any must be set in order for pydantic to accept/validate the consumer
```
jsolaas, line 301: How do we fix this?

src/libecalc/presentation/yaml/mappers/component_mapper.py
```diff
@@ -134,6 +139,16 @@ def from_yaml_to_dto(
                 except ValidationError as e:
                     raise DtoValidationError(data=data, validation_error=e) from e
 
+            elif component_type == ComponentType.PUMP_V2:
+                try:
+                    pump_yaml = YamlPump(**data)
+                    return pump_yaml.to_temporal_domain_models(
+                        references=self.__references, timesteps=timevector, fuel=fuel
+                    )
```
jsolaas, line 147: Can't this be done in energy calculator instead? So we don't have to send timevector here?

src/libecalc/presentation/yaml/yaml_types/components/yaml_pump.py
```diff
@@ -33,6 +55,10 @@ class Config:
 
     energy_usage_model: YamlTemporalModel[str]
 
+    # TODO: Currently we share the stream conditions between consumer system and pump. Might change if they deviate ...
+    # TODO: We may also want to enforce the names of the streams, and e.g. limit to 1 in -and output stream ...? At least we need to know what is in and out ...
+    stream_conditions: Optional[YamlConsumerStreamConditions]
```
jsolaas, line 60: I still think we should fix the schema here, by not making stream_conditions optional.

src/tests/libecalc/dto/test_generator_set.py
```diff
@@ -66,6 +66,9 @@ def test_valid(self):
             )
         }
 
+    @pytest.mark.skip(
+        reason="Due to circular dependencies we cannot specify a pump as an alternative as a consumer for genset, therefore we must set Any, which makes this not fail."
```
jsolaas, line 70: How do we fix it?

src/libecalc/core/result/base.py
TeeeJay: Due to the comment above...circular import. Next step with v2 will have to be to sort out a few ofthe ciruclar import issues..

src/libecalc/dto/components.py
TeeeJay: not a quick fix, but solve all the "hacks" where we have incorrect directions of dependencies.

src/libecalc/presentation/yaml/mappers/component_mapper.py
TeeeJay: not sure, didnt check, but since we are moving away from energy calculator?

src/tests/libecalc/dto/test_generator_set.py
TeeeJay: every case must be considered separately. Here we must figure out the problem with the pumpcharts I think. We can however not move those to domain yet, as that will cause serialization to not work (unless we add that, but it doesnt make sense to serialize domain objects...). So we must avoid need for serialization of the ecalcmodel first...I think.

jsolaas:
no more v2

PR #355, opened 2024-01-24
Chore/fix serialization

jsolaas:

- [x] Considered to add tests

Fix serialization of results in web

- Serialization of ndarray
- fixes conversion of None to NaN when validating/parsing serialized core result (TimeSeries[float]) 
- document _type_handler in unit.py
- convert numpy to python by using tolist. This also converts the items to python native types, instead of only converting the array to list


markusrf (approved):
I'm not too familiar with this code, but looks good to me

PR #361, opened 2024-01-29
fix: expression error messages

jsolaas:


src/libecalc/expression/expression.py
```diff
@@ -10,6 +10,7 @@
 from pydantic_core import CoreSchema, core_schema
 from pydantic_core.core_schema import ValidationInfo
 
+from libecalc.common.errors.exceptions import EcalcError
```
jsolaas, line 13: This is wrong if we are trying to make expression module independent of ecalc.

ketwik (approved):
Consider handling this as a Bad Request.

PR #374, opened 2024-02-05
fix: get_max_rate for single/variable speed compressor trains

olelod:
Inside get_max_rate for variable speed compressor trains, we find the results for maximum speed and minimum speed for the shaft. If the target discharge pressure is between the discharge pressure at minimum speed and at maximum speed, we iterate on speed between the minimum and maximum speeds until a solution is found giving the correct target discharge pressure. There has been no check that the results and minimum and maximum speeds are actually valid. The result have ended up in situations where the results at minimum speed is invalid, and hence the final result also potentially end up as invalid (when in fact valid solutions exists). Also there has been no check to make sure that the found maximum standard rate requires less power than the maximum power defined for the compressor train.

This PR adds two things:

1. For variable speed compressor trains (including multiple streams and pressures), we find the minimum and maximum speeds that produce a valid result for the compressor train given rate and suction pressure. 
2. For single/variable speed compressor trains we make sure that the max rate is below maximum power

- [x] Considered to add tests

### DIFF
--- a/docs/docs/changelog/next.md
+++ b/docs/docs/changelog/next.md
@@ -15,6 +15,8 @@ sidebar_position: 1
 
 ## Fixes
 
+* Make sure that maximum standard rate calculations takes a potential maximum power constraint into account for single speed compressor trains and variable speed compressor trains. 
+
 
 ## Breaking changes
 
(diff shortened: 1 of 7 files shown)


frodehk (approved):
Looks good to me. Although I don´t have the full overview of all function calls and iterations. The overall logic seems fine.

PR #377, opened 2024-02-08
fix: validation using yaml classes

jsolaas:
- Yaml classes can now be used for validation
- Implemented validate for YamlValidator


frodehk (approved):
Nice!

PR #379, opened 2024-02-12
chore: allow venting emitters without genset and fuelconsumer

frodehk:

eCalc will fail if one of the keywords `GENERATORSETS` or `FUELCONSUMERS` is missing. Hence, it is not possible to have a model with only `VENTING_EMITTERS`.

Modify graph_result to accept a model with `VENTING_EMITTERS` only. 

https://equinor-ecalc.atlassian.net/browse/ECALC-772?atlOrigin=eyJpIjoiNjkyMjQ1YmZmMjI4NDAyZjhjMGRlZjc2YmJlY2NlMTAiLCJwIjoiaiJ9


src/libecalc/application/graph_result.py
```diff
@@ -128,6 +128,33 @@ def _evaluate_installations(self, variables_map: dto.VariablesMap) -> List[libec
...
+                        timesteps=self.variables_map.time_vector,
+                        is_valid=TimeSeriesBoolean(
+                            timesteps=self.variables_map.time_vector,
+                            values=[True] * self.variables_map.length,
+                            unit=Unit.NONE,
+                        ),
+                        energy_usage=energy_usage,
+                        energy_usage_cumulative=energy_usage.to_volumes().cumulative(),
+                    )
+                )
```
jsolaas, line 156: This causes venting emitter to be treated as a consumer right? I'd prefer it if we could keep venting emitters asset result and instead handle the case of no consumers here.

frodehk:
Close due to rebase issues. Continued in this PR: https://github.com/equinor/ecalc/pull/392

PR #392, opened 2024-02-23
chore: allow venting emitters without fuelconsumers

frodehk:

eCalc will fail if one of the keywords `GENERATORSETS` or `FUELCONSUMERS` is missing. Hence, it is not possible to have a model with only `VENTING_EMITTERS`.

Modify graph_result to accept a model with `VENTING_EMITTERS` only. 

https://equinor-ecalc.atlassian.net/browse/ECALC-772?atlOrigin=eyJpIjoiNjkyMjQ1YmZmMjI4NDAyZjhjMGRlZjc2YmJlY2NlMTAiLCJwIjoiaiJ9

### DIFF
--- a/src/libecalc/application/graph_result.py
+++ b/src/libecalc/application/graph_result.py
@@ -145,13 +145,17 @@ class GraphResult:
                 ),  # Initial value, handle no power output from components
             )
 
-            energy_usage = reduce(
-                operator.add,
-                [
-                    TimeSeriesRate.from_timeseries_stream_day_rate(component.energy_usage, regularity=regularity)
-                    for component in sub_components
-                    if component.energy_usage.unit == Unit.STANDARD_CUBIC_METER_PER_DAY
-                ],
+            energy_usage = (
+                reduce(
+                    operator.add,
+                    [
+                        TimeSeriesRate.from_timeseries_stream_day_rate(component.energy_usage, regularity=regularity)
+                        for component in sub_components
+                        if component.energy_usage.unit == Unit.STANDARD_CUBIC_METER_PER_DAY
+                    ],
+                )
+                if sub_components
+                else 0
             )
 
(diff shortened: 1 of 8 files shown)


src/libecalc/dto/components.py
```diff
@@ -364,15 +363,6 @@ def check_user_defined_category(cls, user_defined_category, info: ValidationInfo
 
         return user_defined_category
 
-    @field_validator("fuel_consumers", mode="before")
```
jsolaas, line 367: Check that either fuel_consumer or venting_emitters is set? Would be nice to also add that in the yaml class

jsolaas (commented):
Can we add a test with only venting emitters also?

src/libecalc/dto/components.py
frodehk: @jsolaas , you mean in the yaml_installation right? So, add a check here and in yaml class (even if it is not in use yet)?

frodehk:
> Can we add a test with only venting emitters also?

Tests are added

src/libecalc/application/graph_result.py
```diff
@@ -1233,27 +1244,41 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
...
-        asset_hydrocarbon_export_rate_core = reduce(
-            operator.add,
-            [installation.hydrocarbon_export_rate for installation in installation_results],
+        asset_hydrocarbon_export_rate_core = (
+            reduce(
+                operator.add,
+                [installation.hydrocarbon_export_rate for installation in installation_results],
+            )
+            if installation_results
+            else venting_emitter_zero_rate
```
jsolaas, line 1253: Confusing name of zero_rate here

jsolaas (approved):
Nice!

PR #397, opened 2024-03-01
Resolve potential numerical issues with generic from input compressor charts

olelod:
Since the compressor chart generated is also written to the results, the compressor curves are validated/rounded to six significant digits. This have shown to lead to numerical issues in some cases. This PR multiplies the maximum values for head in the input with a small fraction to make sure points are covered also after validation/rounding.


src/libecalc/core/models/compressor/train/chart/chart_creator.py
```diff
@@ -72,9 +75,11 @@ def from_rate_and_head_values(
             maximum_head_on_maximum_speed_curve_unified - minimum_head_on_maximum_speed_curve_unified
         )
 
-        # Maximum values from input
+        # Maximum values from input - scaling head values slightly to avoid having points exactly at the ends of the
+        # maximum speed curve, which potentially leads to numerical issues with the current implementation (rounding
+        # chart curves to a certain number of significant digits since they are also exported in results)
         maximum_actual_rate = np.max(actual_volume_rates_m3_per_hour)
-        maximum_head_value = np.max(heads_joule_per_kg)
+        maximum_head_value = np.max(heads_joule_per_kg) * (1 + PRESSURE_CALCULATION_TOLERANCE)
```
jsolaas, line 82: I really think we should try to postpone the rounding instead of this, but if we are mergin this we should at least create a ticket for handling of rounding and link this change so it can be reverted, 

The issue would be "round results explicitly after calculations" or something like that.

jsolaas (approved):

PR #401, opened 2024-03-07
feat: add file context to model validation error

jsolaas:


src/libecalc/presentation/yaml/validation_errors.py
```diff
@@ -203,6 +215,24 @@ def yaml(self) -> Optional[str]:
 
         return yaml.dump(self.data, sort_keys=False).strip()
 
+    @property
+    def file_context(self) -> Optional[FileContext]:
+        if hasattr(self.data, "end_mark") and hasattr(self.data, "start_mark"):
```
TeeeJay, line 220: might it only have one of them? if so, why and is that not relevant? or sufficient to test on only one of them? replace with a method that indicates that this is something that is set by ruamel, pyyaml or pydantic (I dont know/remember). Can or should self.data be able to contain other content as well, depending on validator etc (ie be another datatype/protocol?)

src/libecalc/presentation/yaml/validation_errors.py
```diff
@@ -179,6 +179,18 @@ def as_pydantic_compatible(self) -> PydanticLoc:
         return tuple(keys)
 
 
+@dataclass
+class FileMark:
+    line: int
```
TeeeJay, line 184:
```suggestion
    line_number: int
    column_number: int
```

? Or is that given? datatypes in python are just type hints...

TeeeJay (approved):

src/libecalc/presentation/yaml/validation_errors.py
jsolaas: Maybe a comment? This is set by pyyaml, and if we use ruamel instead the info would be elsewhere, but self.data would also be another type. We are missing our own definition of the yaml data, with file context. The way I see it we should improve the YamlValidator interface to return our own types for both read yaml and file context, we would then be able to move this logic into Pyyaml and ruamel specific YamlReaders/YamlValidators.

src/libecalc/presentation/yaml/validation_errors.py
jsolaas: It's obvious to me, now, since I just implemented this...

PR #402, opened 2024-03-07
refactor: move rounding to result objects

jsolaas:
Rounding where we return results makes it easier to be consistent for all usages of libecalc. Currently we have no need for configurable precision.


frodehk (approved):

src/libecalc/application/energy_calculator.py
```diff
@@ -188,4 +189,4 @@ def evaluate_emissions(
                     )
                 }
                 emission_results[consumer_dto.id] = emission_result
-        return emission_results
+        return Numbers.format_results_to_precision(emission_results, precision=6)
```
ketwik, line 192: Just a question of curiosity. Why is there a fixed precision for the returned number. Should it not be possible for the caller to decide the precision by doing the rounding or through passing a precision parameter?

ketwik (approved):

src/libecalc/application/energy_calculator.py
jsolaas: The reason we introduced rounding was because ecalc reported numbers that seemed much more precise than they actually are (mostly because of issues with changes and when doing UAT). So I think of it as something we always want to do in ecalc, we are not more precise than 6..Then the user can round the numbers more if they want. 

How we round could also be discussed, it could for example be dependent on the precision of the input.

But this isn't something we have spent much time on, this was the easiest way to do the rounding in an explicit way without having to repeat the rounding for cli, api, engine and tests.

PR #406, opened 2024-03-14
chore: sum and split electrical and mechanical power

frodehk:
ECALC-881

Power demand for directly driven compressors (FUELCONSUMERS) should be included in the total power of an installation, if not there is a risk of misinterpreting data, both in the eCalc Web plots and the csv file (the LTP file is not providing the total power so that one is OK).

The following changes are made in graph_result:
- [x] Sum electrical power (`ConsumptionType.ELECTRICITY`) at installation level, named `power_electrical`. 
- [x] Sum mechanical power (`ConsumptionType.FUEL`) at installation level, named `power_mechanical`
- [x] Sum electrical- and mechanical power at installation level, named `power`

https://equinor-ecalc.atlassian.net/browse/ECALC-881?atlOrigin=eyJpIjoiNDk5YTI2Yzg0NjI2NDQwY2E0ZTljMTIwNTc0N2RiOTYiLCJwIjoiaiJ9


src/libecalc/application/graph_result.py
```diff
@@ -89,6 +90,25 @@ def _compute_intensity(
             )
         return emission_intensities
 
+    def _compute_aggregated_power(
+        self, sub_components: list, regularity: TimeSeriesFloat, components: List[ComponentType]
```
jsolaas, line 94:
```suggestion
        self, sub_components: list, regularity: TimeSeriesFloat, component_types: List[ComponentType]
```

src/libecalc/application/graph_result.py
```diff
@@ -130,22 +150,27 @@ def _evaluate_installations(self, variables_map: dto.VariablesMap) -> List[libec
...
+            fuel_consumer_nodes = self.graph.get_fuel_consumer_nodes()
+            fuel_consumer_sub_components = [
+                sub_component for sub_component in sub_components if sub_component.id in fuel_consumer_nodes
+            ]
+
+            power_mechanical = self._compute_aggregated_power(
+                sub_components=fuel_consumer_sub_components,
+                regularity=regularity,
+                components=[ComponentType.COMPRESSOR, ComponentType.COMPRESSOR_V2],
+            )
```
jsolaas, line 167: Will this work if we have a compressor that consumes electrical power?

jsolaas (commented):
I think the consumer-dtos has a `consumes` property, could we use that instead of filtering on types?

src/libecalc/application/graph_result.py
frodehk: Yes, because we have ensured that only fuel consumers are sent in (fuel_consumer_sub_components)

frodehk:
> I think the consumer-dtos has a `consumes` property, could we use that instead of filtering on types?

At this stage the genset is already aggregated (genset has no `consumes`-property)

> I think the consumer-dtos has a `consumes` property, could we use that instead of filtering on types?

Yes they have. However, at this stage the genset-results are already aggregated (in consumer_results). If we should use the consumes-property, we need to go through each individual consumer under genset again. Do you suggest to do it earlier, before graph_result, when creating consumer_results?

src/libecalc/application/graph_result.py
frodehk: Have changed to consumption type instead

frodehk:
> I think the consumer-dtos has a `consumes` property, could we use that instead of filtering on types?

Done

src/libecalc/application/graph_result.py
```diff
@@ -130,22 +164,38 @@ def _evaluate_installations(self, variables_map: dto.VariablesMap) -> List[libec
...
+                consumption_type=ConsumptionType.FUEL,
+                exclude_components=exclude_components,
+            )
+
+            power_sum = [
+                electrical + mechanical
+                for electrical, mechanical in zip(power_electrical.values, power_mechanical.values)
+            ]
+            power = copy.deepcopy(power_electrical)
+            power.values = power_sum
```
jsolaas, line 198: add for TimeSeriesRate is implemented, I think. If not we could implement it. So that we can write 
`power = power_electrical + power_mechanical`

src/libecalc/application/graph_result.py
```diff
@@ -130,22 +164,38 @@ def _evaluate_installations(self, variables_map: dto.VariablesMap) -> List[libec
...
+                consumption_type=ConsumptionType.ELECTRICITY,
+                exclude_components=exclude_components,
             )
 
+            power_mechanical = self._compute_aggregated_power(
+                sub_components=sub_components,
+                regularity=regularity,
+                consumption_type=ConsumptionType.FUEL,
+                exclude_components=exclude_components,
+            )
```
jsolaas, line 191: This still seems a bit too likely to fail in the future. Could we implement a get_leaf_nodes or get_individual_consumers in either Graph or ComponentGraph?

Or instead of summing leafs we could ask the graph for direct successors. Then check if generator set or not. That seems more correct to me. Don't we have some power_loss or something that could adjust the power in generator set? So power requirement from consumers is not equal to the power generated in generator set.

src/libecalc/application/graph_result.py
frodehk: Done

src/libecalc/application/graph_result.py
frodehk: Done

jsolaas (approved):

PR #407, opened 2024-03-14
chore: add new minimal model test fixture

markusrf:
Refs: ECALC-337

- [x] Considered to add tests

Need to easily create two different models in some tests

Adds a new test fixture

### DIFF
--- a/src/libecalc/fixtures/cases/minimal/minimal_yaml.py
+++ b/src/libecalc/fixtures/cases/minimal/minimal_yaml.py
@@ -37,3 +37,34 @@ INSTALLATIONS:
         )
 
     return minimal_model_yaml
+
+
+@pytest.fixture
+def minimal_model_yaml_factory_2():
+    def minimal_model_yaml(fuel_rate: Union[int, str] = 50):
+        return YamlModel(
+            name="minimal_model_2",
+            source=f"""
+FUEL_TYPES:
+  - NAME: fuel_2
+    EMISSIONS:
+      - NAME: co2
+        FACTOR: 2
+
+START: 2020-01-01
+END: 2023-01-01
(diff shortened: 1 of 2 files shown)


src/libecalc/fixtures/cases/minimal/minimal_yaml.py
```diff
@@ -37,3 +37,34 @@ def minimal_model_yaml(fuel_rate: Union[int, str] = 50):
...
+      - NAME: co2
+        FACTOR: 2
+
+START: 2020-01-01
+END: 2023-01-01
+
+INSTALLATIONS:
+  - NAME: minimal_installation_2
+    HCEXPORT: 0
+    FUEL: fuel
```
TeeeJay, line 60: invalid model, refer to `fuel`but has `fuel_2`. should be valid imo

src/libecalc/fixtures/cases/minimal/minimal_yaml.py
```diff
@@ -37,3 +37,34 @@ def minimal_model_yaml(fuel_rate: Union[int, str] = 50):
         )
 
     return minimal_model_yaml
+
+
+@pytest.fixture
+def minimal_model_yaml_factory_2():
```
TeeeJay, line 43: reuse the model above but add extra parameters for fuel and installation name (with default values as before)? To easier spot and know the difference between the models?

src/libecalc/fixtures/cases/minimal/minimal_yaml.py
markusrf: good point, it somehow seemed to pass the validation when I ran it 🤷‍♂️

PR #412, opened 2024-03-19
fix: use yaml expression type for installation

jsolaas:
Improves error messages for hcexport and regularity

### DIFF
--- a/src/tests/libecalc/input/mappers/variables_mapper/test_variables_mapper.py
+++ b/src/tests/libecalc/input/mappers/variables_mapper/test_variables_mapper.py
@@ -9,8 +9,6 @@ from libecalc.presentation.yaml.mappers.variables_mapper.variables_mapper import
 )
 from libecalc.presentation.yaml.yaml_types.yaml_variable import YamlSingleVariable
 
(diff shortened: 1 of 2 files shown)


src/tests/libecalc/input/mappers/variables_mapper/test_variables_mapper.py
```diff
@@ -9,8 +9,6 @@
 )
 from libecalc.presentation.yaml.yaml_types.yaml_variable import YamlSingleVariable
 
-# from libecalc.presentation.yaml.yaml_types import YamlSingleVariable
```
jsolaas, line 12: YamlVariables still do not use YamlExpressionType, this is because we expect the type to be parsed into an expression. YamlExpressionType only validates. We should look into parsing into an expression in YamlExpressionType I think, if we go that route we probably don't need to implement pydantic specific methods in expression.py (`__get_pydantic_core_schema__` etc.). The alternative is to make expression.py validate similar to YamlExpressionType and use that in the Yaml classes.

TeeeJay (approved):

PR #417, opened 2024-03-20
fix: support control margin for single speed compressor stage

jsolaas:


TeeeJay (approved):
Seems that this nicely gathers duplicate attributes in not needed classes, but I do not see how this fixes control margin ...

jsolaas:
CONTROL_MARGIN for single speed is not yet implemented, this should wait for that feature

PR #420, opened 2024-03-22
docs: update generic workflow description in documentation

equinor-schen:
I don't have the sketches made previously, so some of the comments are hard to address. Tried to fix most things.

- [ ] It's a bit difficult to see the color of the arrows in the sketch. Maybe you should make them wider?
- [ ] I'm not sure oil export pump should be shown in the sketch, as we rarely model them (and you've listed it under base load in the workflow)
- [x] Consider removing "Oil produced" in the workflow. We rarely model the oil export pumps, the oil production is only needed for HCEXPORT, meaning that oil production is not really needed for the power/emissions calculations
- [x] Change from "produced" to "production" as eCalc is not only dealing with historical data (but also prediction)
- [x] Isn't it a bit strange to list ESPs under other emissions (as they're electrical and need power)? There's not that many ESPs, so I suggest that you just leave it out. The alternative would be to have the separated, as for compressors and water injection.
- [ ] The split between facility input & consumer input is a bit unclear to me. Maybe you should merge them? Max capacity is for instance listed twice for some components.
- [ ] Seems like the only thing to do if exceeding max genset power is to adjust the genset curve. That's something that should be done with care! And there should be a warning like for adjusting compressor/pump charts. Other measures could be performed instead.
- [x] Seems strange to list validation after checking match in the calibration diagram (that comes after the validation diagram). The first thing should be to check the validation, and then the power/fuel match. I think this will be capture if you just remove validation from the calibration diagram.
- [x] Consider changing from Sm³/day to Sm³/cd to align with eCalc.
- [x] Not entirely correct to state that the base load is not dependent on production. It's correct that it's not dependent on variations in production (& injection!), but parts of the base load are dependent on there being prod/inj (i.e. on/off).
- [x] The explanation of the oil export pumps should probably be rephrased to say something about these pumps normally being modeled as constant load (because there's little variation in actual load for these pumps - and they're small consumers).
- [x] Don't understand why it says that validation should be done "Before running..." when the description is discussing invalid results...?
- [x] I think you should remove "often" from the calibration, as I don't think there's any other ways of doing the calibration than "history matching".

### DIFF
--- a/docs/docs/about/modelling/workflow/generic_workflow.md
+++ b/docs/docs/about/modelling/workflow/generic_workflow.md
@@ -16,11 +16,10 @@ The [workflow](#workflow) below will outline what is necessary to obtain for eac
 ```mermaid
   flowchart TD;
       subgraph ide1 ["Required Subsurface Profiles [All in Sm3/d]"]
-          ide1_A[Oil Produced];
-          ide1_B[Gas Produced];
-          ide1_C[Water Produced];
-          ide1_D[Gas Injected];
-          ide1_E[Water Injected];
+          ide1_A[Gas Production];
+          ide1_B[Water Production];
+          ide1_C[Gas Injection];
+          ide1_D[Water Injection];
       end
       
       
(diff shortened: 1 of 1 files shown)


docs/docs/about/modelling/workflow/generic_workflow.md
```diff
@@ -76,11 +76,11 @@ The [workflow](#workflow) below will outline what is necessary to obtain for eac
 
           ide4_A -. yes .-> ide4_A_1{"`Invalid Compressors?`"};
           ide4_A_1 -. yes .-> ide4_A_1_1(["`Either head or rate is too high`"]);
-          ide4_A_1_1 --> ide4_A_1_2(["`Plot operational points and adjust charts to fit historical data`"]);
+          ide4_A_1_1 --> ide4_A_1_2(["`Based on the qaulity and uncertianty of data, user can either adjust history operational points or charts to fit history data`"]);
```
olelod, line 79:
```suggestion
          ide4_A_1_1 --> ide4_A_1_2(["`Based on the quality and the uncertainty of the data, the user can either adjust the historical operational points or adjust the charts to fit the historical data`"]);
```

docs/docs/about/modelling/workflow/generic_workflow.md
```diff
@@ -76,11 +76,11 @@ The [workflow](#workflow) below will outline what is necessary to obtain for eac
...
     
           ide4_A -. yes .-> ide4_A_3{"`Invalid Pumps?`"};
           ide4_A_3 -. yes .-> ide4_A_3_1(["`Either head or rate is too high`"]);
-          ide4_A_3_1 --> ide4_A_3_2(["`Plot operational points and adjust charts to fit historical data`"]);
+          ide4_A_3_1 --> ide4_A_3_2(["`Based on the qaulity and uncertianty of data, user can either adjust history operational points or charts to fit hisotry data`"]);
```
olelod, line 83:
```suggestion
          ide4_A_3_1 --> ide4_A_3_2(["`Based on the quality and the uncertainty of the data, the user can either adjust the historical operational points or adjust the charts to fit the historical data`"]);
```

halsyl:
I'm not able to find my old list of comments, but have gone through the text and made new ones, @equinor-schen:

1. It's a bit difficult to see the color of the arrows in the sketch. Maybe you should make them wider?

2. I'm not sure oil export pump should be shown in the sketch, as we rarely model them (and you've listed it under base load in the workflow)

3. Consider removing "Oil produced" in the workflow. We rarely model the oil export pumps, the oil production is only needed for HCEXPORT, meaning that oil production is not really needed for the power/emissions calculations

4. Change from "produced" to "production" as eCalc is not only dealing with historical data (but also prediction)

5. Isn't it a bit strange to list ESPs under other emissions (as they're electrical and need power)? There's not that many ESPs, so I suggest that you just leave it out. The alternative would be to have the separated, as for compressors and water injection.

6. The split between facility input & consumer input is a bit unclear to me. Maybe you should merge them? Max capacity is for instance listed twice for some components.

7. Seems like the only thing to do if exceeding max genset power is to adjust the genset curve. That's something that should be done with care! And there should be a warning like for adjusting compressor/pump charts. Other measures could be performed instead.

8. Seems strange to list validation after checking match in the calibration diagram (that comes after the validation diagram). The first thing should be to check the validation, and then the power/fuel match. I think this will be capture if you just remove validation from the calibration diagram.

9. Consider changing from Sm³/day to Sm³/cd to align with eCalc.

10. Not entirely correct to state that the base load is not dependent on production. It's correct that it's not dependent on variations in production (& injection!), but parts of the base load are dependent on there being prod/inj (i.e. on/off).

11. The explanation of the oil export pumps should probably be rephrased to say something about these pumps normally being modeled as constant load (because there's little variation in actual load for these pumps - and they're small consumers).

olelod (approved):

PR #429, opened 2024-04-10
feat: new PfS columns

frodehk:
ECALC-365

There’s currently two columns with PfS data in Centuries that is not reported in eCalc (LTP file). The columns are required for the RNB reporting, and thus important, but the data are currently not available in eCalc making it impossible to include it in the LTP file.

These data partly requires post-processing of the eCalc results and are often forgotten in the transfer from eCalc to Centuries.

Possibility for additional input to eCalc at generator set level, to get the wanted data in the LTP file:

- [x] Add optional input for genset: cable loss factor (loss from shore) . Used if genset category is POWER-FROM-SHORE
- [x] Calculate Power supply onshore (including cable loss): Power required * (1 + cable loss)
- [x] Add optional time series of maximum PfS (Max usage from shore)
- [x] Report power supply onshore and max usage from shore to LTP-report
- [x] Add tests

https://equinor-ecalc.atlassian.net/browse/ECALC-365?atlOrigin=eyJpIjoiOTg0YmIyNmQ3NWYwNDM4ZjliZWI0MDYwNDBlNDMxMjUiLCJwIjoiaiJ9


src/libecalc/core/consumers/generator_set.py
```diff
@@ -48,6 +50,22 @@ def evaluate(
         fuel_rate = self.evaluate_fuel_rate(power_requirement, variables_map=variables_map)
         power_capacity_margin = self.evaluate_power_capacity_margin(power_requirement, variables_map=variables_map)
 
+        # Check if Power From Shore, if so do necessary calculations.
+        if ConsumerUserDefinedCategoryType.POWER_FROM_SHORE in self.data_transfer_object.user_defined_category.values():
+            # cable_loss = Expression.evaluate(
```
TeeeJay, line 55: Remove commented out code? To be used?

src/libecalc/presentation/exporter/queries.py
```diff
@@ -276,11 +283,15 @@ def query(
                             fuel_consumer_result: GeneratorSetResult = installation_graph.get_energy_result(
                                 fuel_consumer.id
                             )
+                            power_result = fuel_consumer_result.power
+
+                            if self.power_from_shore_output == PowerFromShoreOutputType.POWER_SUPPLY_ONSHORE:
```
TeeeJay, line 288: I assume this is intended, but it is not clear how this works, ie. the same query is used for 3 different scenarios, depending on what is being asked for?

src/tests/libecalc/output/results/test_ltp.py
```diff
@@ -390,3 +393,14 @@ def test_total_oil_loaded_old_method():
...
+
+def test_power_from_shore():
+    """Test power from shore output for LTP export."""
+    [
+        datetime(2027, 1, 1),
+        datetime(2028, 1, 1),
+    ]
+    regularity = 0.2
+
+    ltp_pfs_yaml_factory(regularity=regularity)
```
TeeeJay, line 406: no assert...what is being tested here?

frodehk: The test is not complete yet:-) Will do.

src/libecalc/presentation/exporter/queries.py
frodehk: Yes, instead of having a new query for each extra PfS-column (all three columns are different power-outputs)

src/libecalc/core/result/results.py
```diff
@@ -60,6 +60,8 @@ class GeneratorSetResult(GenericComponentResult):
 
     typ: Literal["genset"] = "genset"
     power_capacity_margin: TimeSeriesStreamDayRate
+    power_supply_onshore: Optional[TimeSeriesStreamDayRate]
+    max_usage_from_shore: Optional[TimeSeriesStreamDayRate]
```
jsolaas, line 64: We should move it to LTP instead of having it in core as we haven't got the time to look into how to implement this the way we want it..in my opinion.

src/tests/libecalc/output/results/test_ltp.py
frodehk: Test is done now

src/libecalc/core/consumers/generator_set.py
frodehk: Fixed

src/libecalc/fixtures/cases/ltp_export/ltp_power_from_shore_yaml.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def ltp_pfs_yaml_factory(
+    regularity: float,
+    cable_loss: ExpressionType,
+    max_usage_from_shore: ExpressionType,
+    load_direct_consumer: float,
+) -> [Asset, VariablesMap]:
+    input_text = f"""
+    START: 2025-01-01
```
jsolaas, line 21: Can we add this yaml case as a fixture in conftest? Specifically in `valid_example_case_yaml_case`, as this will make sure we test that the model is valid with the yaml validation. I think this PR is missing the Yaml class update, similar to the update in dto.

frodehk: Have tried to add fixture in conftest. The Yaml class is updated.

src/tests/libecalc/presentation/yaml/yaml_models/test_pyyaml_yaml_model.py
```diff
@@ -72,3 +72,6 @@ def test_valid_cases(self, valid_example_case_yaml_case):
                 ),
             }
         )
+
+    def test_valid_dto_cases(self, valid_example_case_dto_case):
+        valid_example_case_dto_case.ecalc_model.model_validate(valid_example_case_dto_case.ecalc_model)
```
jsolaas, line 77: The point was to test that the yaml was read and validated correctly, i.e. make sure we have the yaml classes etc.

I think you can remove the fixture created and this test. If we are going to add a test for this case we should set up a YamlCase. Which is basically what is being done in `ltp_pfs_yaml_factory` except it is converted to a DTOCase. For other cases we explicitly write both a YamlCase and DTOCase to make sure the mapping is correct between yaml and dto. If we have a YamlCase it can be added to the dict in `conftest` and it will automatically be tested and made sure it is valid.

jsolaas (approved):

src/tests/libecalc/presentation/yaml/yaml_models/test_pyyaml_yaml_model.py
frodehk: Removed the checks in conftest for now.

PR #431, opened 2024-04-11
chore: include invalid data in pump model plot

frodehk:
ECALC-799

The lower-level results plots currently differ between ‘compressors’ and ‘pumps’. The heading on both power plots state that invalid points are included, however, that is only true for compressors. Pump model plots should also include invalid points.

https://equinor-ecalc.atlassian.net/browse/ECALC-799?atlOrigin=eyJpIjoiYzY1YzExNWJkZTcwNDk2Nzk2MGJmODJmZGVhYTk5OTUiLCJwIjoiaiJ9


src/libecalc/core/models/pump/pump.py
```diff
@@ -305,6 +309,16 @@ def evaluate_rate_ps_pd_density(
...
+        pump_flags = np.where(
+            rates_m3_per_hour > self.pump_chart.maximum_rate, PumpFlag.INVALID_ABOVE_MAXIMUM_PUMP_RATE.value, pump_flags
+        )
+
+        pump_flags = np.where(
+            operational_heads > actual_heads + self._head_margin,
+            PumpFlag.INVALID_REQUIRED_HEAD_ABOVE_ACTUAL_HEAD.value,
+            pump_flags,
+        )
+
```
olelod, line 321: PumpModelResults inherits is_valid from EnergyFunctionResult and is basically is_valid = False when values are nan. I guess you need a new is_valid property on PumpModelResult using the pump_flags you have defined?

frodehk: Agree, that is probably what we need.

PR #438, opened 2024-04-16
fix: detect duplicate keys in yaml file and raise error

olelod:
ECALC-650

If there are duplicate keys defined in the yaml file (for example several VARIABLES with the same name), the last one will be kept and no error given. 

This PR detects duplicate keys and raises and error pointing to where in the yaml file the problem occurs.

### DIFF
--- a/src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
+++ b/src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
@@ -145,6 +145,19 @@ class PyYamlYamlModel(YamlValidator, YamlModel):
                 )
 
         def load(self, yaml_file: ResourceStream):
+            class UniqueKeyLoader(self.__loader):  # type: ignore[name-defined]
+                def construct_mapping(self, node, deep=False):
+                    mapping = set()
+                    for key_node, _ in node.value:
+                        each_key = self.construct_object(key_node, deep=deep)
+                        if each_key in mapping:
+                            raise DataValidationError(
+                                data=None,
+                                message=f"Duplicate key: {each_key!r} is found in {key_node.start_mark.name} on line {key_node.start_mark.line + 1}",
+                            )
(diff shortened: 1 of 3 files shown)


src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
```diff
@@ -145,14 +145,27 @@ def __init__(
                 )
 
         def load(self, yaml_file: ResourceStream):
+            class UniqueKeyLoader(self.__loader):  # type: ignore[name-defined]
+                def construct_mapping(self, node, deep=False):
```
TeeeJay, line 149: how is deep set? Is it by us or by the SafeLoader? When we set it to false by default, does it mean that we only compare keys at a given level, or that it doesnt go further down when searching for duplicate keys?

src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
```diff
@@ -145,14 +145,27 @@ def __init__(
...
+                    mapping = set()
+                    for key_node, _ in node.value:
```
TeeeJay, line 151: is this code exactly as SafeLoader is, except that it checks explicitly for duplicate keys at each level?

olelod: The way I understand it self.__loader is SafeLoader with the !include constructor added. And this also adds checking for duplicate keys to self._loader.

TeeeJay: I see now that you call super(). I didnt see that earlier. So it calls SafeConstructor´s construct_mapping it seems, so then I think it is good

TeeeJay (approved):
add a note in changelog about this, and change the commit message to use fix: instead of chore: imo (as this could have lead to errors?) feel free to disagree :)

PR #449, opened 2024-04-24
fix: error when only venting emitters are specified for an installation

frodehk:
ECALC-1061

Error when venting emitters are defined alone, without a generator set - for a given installation. The reason is that the installation results are empty, and the installation regularity cannot be found for the particular venting emitters.

More robust reading of regularity for venting emitters, independent of installation results: Reading from the asset object directly.

https://equinor-ecalc.atlassian.net/browse/ECALC-1061?atlOrigin=eyJpIjoiNWUwMDMwMjkxMzc2NGEyMmJmMGY5ZTg3NTBlOThkZDQiLCJwIjoiaiJ9

### DIFF
--- a/src/libecalc/application/graph_result.py
+++ b/src/libecalc/application/graph_result.py
@@ -1251,16 +1251,17 @@ class GraphResult:
 
         # When only venting emitters are specified, without a generator set: the installation result is empty.
         # Ensure that the installation regularity is found, even if only venting emitters are defined:
-        regularity_emitters = {
-            installation.id: TimeSeriesFloat(
-                values=TemporalExpression.evaluate(
-                    temporal_expression=TemporalModel(installation.regularity), variables_map=self.variables_map
-                ),
-                unit=Unit.NONE,
-                timesteps=self.variables_map.time_vector,
-            )
-            for installation in asset.installations
-        }
+        if not installation_results:
+            regularities = {
+                installation.id: TimeSeriesFloat(
+                    values=TemporalExpression.evaluate(
+                        temporal_expression=TemporalModel(installation.regularity), variables_map=self.variables_map
+                    ),
+                    unit=Unit.NONE,
+                    timesteps=self.variables_map.time_vector,
+                )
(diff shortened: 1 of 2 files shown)


src/libecalc/application/graph_result.py
```diff
@@ -1249,11 +1249,18 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
...
             sub_components.append(obj)
 
-        regularity_emitters = TimeSeriesFloat(
-            values=[1] * self.variables_map.length,
-            unit=Unit.NONE,
-            timesteps=self.variables_map.time_vector,
-        )
+        # When only venting emitters are specified, without a generator set: the installation result is empty.
+        # Ensure that the installation regularity is found, even if only venting emitters are defined:
+        regularity_emitters = {
```
jsolaas, line 1254: It shouldn't be necessary with a specific regularity for emitters, since it's all the same within an installation. But I guess for the others we use the evaluated regularity from installation result?

jsolaas (commented):
Can we also add a test to make sure this doesn't happen again?

src/libecalc/application/graph_result.py
frodehk: @jsolaas : The challenge is that  installation result is empty, if we have only venting emitters (without fuel consumers or gensets). So it is basically the installation regularity that is read. Yes, it is the evaluated regularity from installation result that is used. Suggestions for other workaround here?

frodehk: `regularity_emitters` is renamed to `regularities`, and is now only set if installation result don´t exist. Ok?

frodehk:
> Can we also add a test to make sure this doesn't happen again?

Have added this check:
https://github.com/equinor/ecalc/blob/0c9ead15c764e128b8126d478b81c556611592dd/src/tests/libecalc/output/results/test_ltp.py#L316

jsolaas (approved):

PR #457, opened 2024-04-26
docs: load examples from file

jsolaas:
Initial work to be able to test and validate the examples we use in docs.

Refs: ECALC-1079


docs/src/components/CodeLinesFromFile.tsx
```diff
@@ -0,0 +1,14 @@
+import CodeBlock, {Props} from '@theme/CodeBlock';
```
TeeeJay, line 1: this theme/package is included by default?

TeeeJay (approved):

docs/src/components/CodeLinesFromFile.tsx
jsolaas: Yes

PR #468, opened 2024-05-07
fix: ensure regularity is evaluated for all installations when only venting emitters

frodehk:
ECALC-1061

- [x] Considered to add tests

Installation results is empty if only venting emitters are specified for one installation. A check was done testing if installation result was empty, then regularity was evaluated for all installations. However, this check is not sufficient, as there can be several installations. One may have only venting emitters, while another can have genset and fuel consumers.

Change test: Check if length of installation_results is less than number of installations.

https://equinor-ecalc.atlassian.net/browse/ECALC-1061?atlOrigin=eyJpIjoiMDMwOGIwNzdkZmRiNGJhZDk5ZDBkMmM2YjVhY2Q2NjIiLCJwIjoiaiJ9

### DIFF
--- a/src/libecalc/application/graph_result.py
+++ b/src/libecalc/application/graph_result.py
@@ -1249,9 +1249,10 @@ class GraphResult:
 
             sub_components.append(obj)
 
-        # When only venting emitters are specified, without a generator set: the installation result is empty.
-        # Ensure that the installation regularity is found, even if only venting emitters are defined:
-        if not installation_results:
+        # When only venting emitters are specified, without a generator set: the installation result
+        # is empty for this installation. Ensure that the installation regularity is found, even if only
+        # venting emitters are defined for one installation:
+        if len(installation_results) < len(asset.installations):
             regularities = {
                 installation.id: TimeSeriesFloat(
                     values=TemporalExpression.evaluate(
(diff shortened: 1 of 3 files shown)


src/libecalc/application/graph_result.py
```diff
@@ -1249,9 +1249,10 @@ def get_asset_result(self) -> libecalc.dto.result.EcalcModelResult:
 
             sub_components.append(obj)
 
-        # When only venting emitters are specified, without a generator set: the installation result is empty.
-        # Ensure that the installation regularity is found, even if only venting emitters are defined:
-        if not installation_results:
+        # When only venting emitters are specified, without a generator set: the installation result
+        # is empty for this installation. Ensure that the installation regularity is found, even if only
+        # venting emitters are defined for one installation:
+        if len(installation_results) < len(asset.installations):
```
TeeeJay, line 1255: `<` or `<=` ?

TeeeJay: should be supplied with a test imo, is that feasible?

frodehk: > `<` or `<=` ?

If I am thinking correctly now: `<` will capture the situations where we have installations with only venting emitters, i.e. installation result is empty for that installation. But, all other cases will have installation results for all installations, or? In any case `<=` will mean that it will always happen, because we will never have more results than installations.

TeeeJay: I see. I would have considered to get regularities from the "model" for all installations in that case, ie change the way we get regularitieties in general, instead of relying on the inst.result. But I undertand and agree that this also works.

TeeeJay (approved):
Remember to use conventional commit when merging/squashing. Also make a patch please :)

PR #473, opened 2024-05-08
chore: make venting emitters emission names case insensitive

frodehk:
ECALC-1148

Emission name, e.g. nmVOC and nmvoc, should be case insensitive for venting emitters to avoid splitting/interpreting the same emission type as different types.

Convert input emission names to lower case, for venting emitters.

https://equinor-ecalc.atlassian.net/browse/ECALC-1148?atlOrigin=eyJpIjoiY2YyMmY5N2ExMjk4NDVhY2E4Y2E3ZDZkMTM2Y2YwNDgiLCJwIjoiaiJ9

### DIFF
--- a/docs/docs/about/references/keywords/EMISSIONS.md
+++ b/docs/docs/about/references/keywords/EMISSIONS.md
@@ -44,6 +44,8 @@ EMISSIONS:
 ~~~~~~~~
 
 ## For venting emitters (type: DIRECT_EMISSION)
+Note that the emission name is case-insensitive.
+
 ### Format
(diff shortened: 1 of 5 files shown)


TeeeJay (approved):
please add a note in changelog about this change/fix. The bug was appearant since 8.12, and was fixed in 8.15 right? So maybe state that? Are there other places where this case-insensitiveness where the user is free to add their own emissions that also must be changed? (we have had this in regualar FUEL for a long time).

frodehk:
As far as I know, other emissions are case-insensitive (emissions in fuel).

PR #487, opened 2024-05-15
fix: use custom type pydantic class to wrap ndarray and avoid validation problem

### LINKED ISSUE #482: Pydantic warnings
**Describe the bug**

Users starting ecalc from command line in the environment provided by Komodo 2024.05 are presented with warnings from Pydantic.

**To Reproduce**

```
(2024.05.01-py38) [havb@be-lx139213:~]$ ecalc
/prog/res/komodo/2024.05.01-py38-rhel8/root/lib64/python3.8/site-packages/pydantic/_internal/_generate_schema.py:382: UserWarning: numpy.ndarray[typing.Any, numpy.dtype[numpy.float64]] is not a Python type (it may be an instance of an object), Pydantic will allow any object with no validation since we cannot even enforce that the input is an instance of the given type. To get rid of this error wrap the type with `pydantic.SkipValidation`.
  warn(
/prog/res/komodo/2024.05.01-py38-rhel8/root/lib64/python3.8/site-packages/pydantic/_internal/_generate_schema.py:382: UserWarning: numpy.ndarray[typing.Any, numpy.dtype[numpy.float64]] is not a Python type (it may be an instance of an object), Pydantic will allow any object with no validation since we cannot even enforce that the input is an instance of the given type. To get rid of this error wrap the type with `pydantic.SkipValidation`.
  warn(
Usage: ecalc [OPTIONS] COMMAND [ARGS]...
Try 'ecalc --help' for help.
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Missing command.                                                                                                     │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

**Expected behavior**

No Pydantic warnings

**Version:**

8.15.1

frodehk:
ECALC-965
fixes #482 

When users are using eCalc directly on CLI, they will be met with a few warnings that will most likely be difficult for them to understand, so we should either handle them or silence them. 

The problem is caused by sending numpy.ndarrays directly into pydantic, giving validation problems. Avoiding the problem using custom type PydanticNDArray to wrap numpy.ndarray, and use that type instead of numpy.ndarray directly.

A more long term solution is to return List and Float from core.

https://equinor-ecalc.atlassian.net/browse/ECALC-965?atlOrigin=eyJpIjoiNjUxNmVlMmY4ZWEyNGU5YzhmN2RhM2MyZTY0NGE2OWYiLCJwIjoiaiJ9

### DIFF
--- a/src/libecalc/core/consumers/legacy_consumer/system/results.py
+++ b/src/libecalc/core/consumers/legacy_consumer/system/results.py
@@ -19,6 +19,7 @@ from libecalc.core.consumers.legacy_consumer.system.operational_setting import (
 )
 from libecalc.core.models.results import CompressorTrainResult, PumpModelResult
 from libecalc.core.result.results import ConsumerModelResult
+from libecalc.core.utils.array_type import PydanticNDArray
 
 
 class ConsumerSystemComponentResult(BaseModel):
@@ -30,7 +31,7 @@ class ConsumerSystemComponentResult(BaseModel):
         return self.consumer_model_result.energy_usage
 
     @property
-    def power(self) -> NDArray[np.float64]:
+    def power(self) -> PydanticNDArray:
         if self.consumer_model_result.power is not None:
             return np.asarray(self.consumer_model_result.power)
         else:
@@ -58,7 +59,7 @@ class ConsumerSystemOperationalSettingResult(BaseModel):
     model_config = ConfigDict(frozen=True)
 
     @property
-    def total_energy_usage(self) -> NDArray[np.float64]:
+    def total_energy_usage(self) -> PydanticNDArray:
         total_energy_usage = np.sum(
             [np.asarray(result.energy_usage) for result in self.consumer_results],
             axis=0,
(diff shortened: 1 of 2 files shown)


src/libecalc/core/consumers/legacy_consumer/system/results.py
```diff
@@ -30,7 +31,7 @@ def energy_usage(self) -> List[Optional[float]]:
         return self.consumer_model_result.energy_usage
 
     @property
-    def power(self) -> NDArray[np.float64]:
+    def power(self) -> PydanticNDArray:
```
TeeeJay, line 34: You suggest to remove np.float64 as annotation? I guess it might have been too much...and not really useful?

frodehk: No, I think np.float64 notation should stand:-)

TeeeJay: I see. I guess this is the same as has been done other places, so should be ok here as well. We should rather spend time fixing it properly as we discussed earlier, just get this fix in place now (wo. float64 then) :)

frodehk: Super, agree. I checked with PydanticNDArray[np.float64], but it failed (BeforeValidator I think).

TeeeJay (approved):
great! please also make a pathc for 8.15 release :) and make sure you use conventional commit syntax as well as referring correctly to jira issue in commit message :)

PR #498, opened 2024-05-28
chore: improve error message for missing header in input csv files

frodehk:
ECALC-773

eCalc correctly points to error in csv file, when missing headers. The name of the file is not given. This can make it difficult /time consuming for the user to debug - if many files.

Improving error message with filename, creating a new Exception.

https://equinor-ecalc.atlassian.net/browse/ECALC-773?atlOrigin=eyJpIjoiNTBiZjFmNmVhOGE3NGExOWE4OTliZjg1ODM4NDRiNjQiLCJwIjoiaiJ9


src/libecalc/infrastructure/file_io.py
```diff
@@ -396,7 +400,7 @@ def _validate_headers(headers: List[str]):
                 "[ _ - # + : . , /] "
             )
         elif re.match(r"^Unnamed: \d+$", header):
-            raise ValueError("CSV input file must include header")
+            raise InvalidResourceHeaderException(f"One or more headers are missing in input-file {name}")
```
TeeeJay, line 403: in my opinion we should probably not have a naution about filename here. It may not always be a filename, it may be a csv dataset retrieved from another place? Therefore, a specific exception as you have here is good, and then catch that higher up, where you can add the additional information about the filename. how does that sound?

TeeeJay: Yes. Basically, if we have an error reading a csv file we would probably have to "stop the line", as in stop execution in order for the user to fix the error to continue? So, yes, as late as possible, higher up the hierarchy, where we are actually trying to read a file called "test.csv", or a data stream...which in theory is also possible...close to the error you have the details of what went wrong (characters on the specific line aso.), and higher up you have the details of the greater task (read a file)..

frodehk: @TeeeJay : The error is now catched differently. Please have a look and see if it is better.

src/libecalc/presentation/yaml/model.py
```diff
@@ -71,7 +71,7 @@ def _create_yaml_configuration(main_yaml_path: Path) -> PyYamlYamlModel:
     def _read_resource(resource_name: Path, *args, read_func: Callable[..., Resource]):
         try:
             return read_func(resource_name, *args)
-        except ValueError as exc:
+        except (InvalidResourceHeaderException, ValueError) as exc:
             logger.error(str(exc))
             raise EcalcError("Failed re read resource", f"Failed to read {resource_name}") from exc
```
TeeeJay, line 76: yes, here since you "trust" InvalidResourceHeaderException (its our exception), and we have a user-friendly message, you can show the message to the user, e.g. add it to the "failed to read message"

src/libecalc/presentation/yaml/model.py
```diff
@@ -71,7 +71,7 @@ def _create_yaml_configuration(main_yaml_path: Path) -> PyYamlYamlModel:
     def _read_resource(resource_name: Path, *args, read_func: Callable[..., Resource]):
         try:
             return read_func(resource_name, *args)
-        except ValueError as exc:
+        except (InvalidResourceHeaderException, ValueError) as exc:
```
TeeeJay, line 74: yes, here since you "trust" InvalidResourceHeaderException (its our exception), and we have a user-friendly message, you can show the message to the user, e.g. add it to the "failed to read message"

src/libecalc/presentation/yaml/model.py
TeeeJay, line 76:
```suggestion
            raise EcalcError("Failed re read resource", f"Failed to read {resource_name}: {str(exc)}") from exc
```

src/tests/libecalc/input/test_file_io.py
```diff
@@ -158,9 +159,9 @@ def test_valid_characters(self, tmp_path_fixture, csv_line: str, is_valid_charac
             )
 
     def test_missing_headers(self, tmp_path_fixture):
-        with pytest.raises(ValueError) as e:
+        with pytest.raises(InvalidResourceHeaderException) as e:
             file_io.read_facility_resource(create_csv_from_line(tmp_path_fixture, "HEADER1        ,,HEADER3"))
-        assert str(e.value) == "CSV input file must include header"
+        assert str(e.value) == "Missing header(s): One or more headers are missing in time series resource"
```
TeeeJay, line 164: if you had added another test reading from a file or resource, and then calling that function in a test, you would get the filename etc in the error message :)

TeeeJay (approved):

src/tests/libecalc/input/test_file_io.py
frodehk: The `create_csv_from_line` actualy gives a filename `csv_file.csv`. It is not captured, as the file name is catched a level above. The only error triggered when using the `file_io.read_facility_resource` isolated like this, is the `InvalidResourceHeaderException` raised in `_validate_headers()`. This does not contain info about file name.

PR #499, opened 2024-05-31
Ecalc 994 patch 8 15 release branch

TeeeJay:

eCalc 8.15.x is in Komodo, and Komodo will have a release next week, that will last until over summer. We should therefore add our dependency changes to that version, and also following (critical) bug fixes.

Just takes dep updates from main to the 8.15 release branch, and reran poetry lock to update lock file correctly.

ECALC-994

### DIFF
--- a/src/libecalc/dto/components.py
+++ b/src/libecalc/dto/components.py
@@ -51,7 +51,7 @@ from libecalc.presentation.yaml.yaml_types.emitters.yaml_venting_emitter import 
 def check_model_energy_usage_type(model_data: Dict[datetime, ConsumerFunction], energy_type: EnergyUsageType):
     for model in model_data.values():
         if model.energy_usage_type != energy_type:
-            raise ValueError(f"Model does not consume {energy_type}")
+            raise ValueError(f"Model does not consume {energy_type.value}")
     return model_data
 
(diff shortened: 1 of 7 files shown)


src/libecalc/dto/components.py
```diff
@@ -51,7 +51,7 @@
 def check_model_energy_usage_type(model_data: Dict[datetime, ConsumerFunction], energy_type: EnergyUsageType):
     for model in model_data.values():
         if model.energy_usage_type != energy_type:
-            raise ValueError(f"Model does not consume {energy_type}")
+            raise ValueError(f"Model does not consume {energy_type.value}")
```
jsolaas, line 54: Do you know when this change happened? It's `str(someEnum)` that becomes `EnumName.name` instead of the string value right? Does that mean a string enum no longer behaves as a string?

jsolaas (approved):

src/libecalc/dto/components.py
TeeeJay: sorry, didnt see comment before I merged. It was introduced when we were supposed to make ecalc compatible with 3.8 -> 3.11. good point. not sure when this occurs ...

PR #495, opened 2024-05-23
chore!: align format of units

frodehk:
BREAKING CHANGE: New unit-format for venting emitters, aligned to what is used in other parts of yaml-file.

ECALC-1059

## Have you remembered _or_ considered?

- [x] To update documentation _(will be done when release is ready)_
- [x] To update manual changelog (`docs/docs/changelog/next.md`) _(will be done when release is ready)_
- [x] To update migration guide (`docs/docs/migration_guides/`) _(will be done when release is ready)_
- [x] To include `BREAKING:` in footer or `!` in header, if breaking
- [x] To add tests
- [x] To use conventional commits syntax (if you squash, make sure that conventional commit is used)
- [x] To include the Jira issue ID somewhere in the commit body (`ECALC-XXXX`)

## Why is this pull request needed?

Units have different format for venting emitters in yaml-file, compared to what is used in other parts of yaml. Can be confusing for users.

## What does this pull request change?

Align unit-format for venting emitters to what is used in other parts of yaml-file. Example: Unit should be now be written as `KILO_PER_DAY`, and not `kg/d`.

## Issues related to this change:
https://equinor-ecalc.atlassian.net/browse/ECALC-1059?atlOrigin=eyJpIjoiYjdkOTA2YjNjOWEwNGJhMWFmODNmZDdkOGU0Y2ViZmQiLCJwIjoiaiJ9

PR #506, opened 2024-05-31
chore: refine unit format alignment

frodehk:
ECALC-1059

Simplify unit definitions in yaml. It is basically a part of the BREAKING CHANGES belonging to the #495.

Write `SM3_PER_DAY` instead of `STANDARD_CUBIC_METERS_PER_DAY`. Write `KG_PER_DAY` instead of `KILO_PER_DAY`.

https://equinor-ecalc.atlassian.net/browse/ECALC-1059?atlOrigin=eyJpIjoiMzA1NzA0MTdiM2NhNDE0YmFiYTBlZmY3N2E1MzMyNjkiLCJwIjoiaiJ9

### DIFF
--- a/src/libecalc/fixtures/cases/all_energy_usage_models/all_models_dto.py
+++ b/src/libecalc/fixtures/cases/all_energy_usage_models/all_models_dto.py
@@ -3,7 +3,6 @@ from datetime import datetime
 import pytest
 
 from libecalc import dto
-from libecalc.common.units import Unit
 from libecalc.common.utils.rates import RateType
 from libecalc.dto.base import (
     ComponentType,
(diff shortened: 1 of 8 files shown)


src/libecalc/fixtures/cases/all_energy_usage_models/all_models_dto.py
```diff
@@ -990,7 +990,7 @@ def methane_venting(regularity) -> YamlVentingEmitter:
             YamlVentingEmission(
                 name="CH4",
                 rate=YamlEmissionRate(
-                    value="FLARE;METHANE_RATE", unit=Unit.KILO_PER_DAY.name, type=RateType.STREAM_DAY
+                    value="FLARE;METHANE_RATE", unit=YamlEmissionRateUnits.KILO_PER_DAY, type=RateType.STREAM_DAY
```
TeeeJay, line 993: If we are to have different enums/strings for different types of equipment, there will be a lot of duplication and it will also not be a "general way" to deal with units in eCalc, which I do not agree with. Instead we should possibly support different ways of "writing" the unit, and that it should be the same all over the place, instead of KG_PR_DAY one place and KILOGRAMS_PER_DAY another place ...lets discuss

TeeeJay (approved):

PR #508, opened 2024-06-03
chore: fix katex equations

jsolaas:


docs/docusaurus.config.ts
```diff
@@ -128,10 +128,10 @@ const config: Config = {
     } satisfies Preset.ThemeConfig,
   stylesheets: [
     {
-      href: 'https://cdn.jsdelivr.net/npm/katex@0.16.10/dist/katex.min.js',
+      href: 'https://cdn.jsdelivr.net/npm/katex@0.16.10/dist/katex.min.css',
```
jsolaas, line 131: Never worked I assume. Not _that_ easy to spot duplicated equations...

TeeeJay: it has worked before...and fixed multiple times...

PR #519, opened 2024-06-07
fix: implement YamlModel specific errors

jsolaas:
YamlError gives additional info about the error in yaml, in addition to ensuring a consistent interface across different yaml readers.

### DIFF
--- a/src/libecalc/presentation/yaml/file_context.py
+++ b/src/libecalc/presentation/yaml/file_context.py
@@ -1,0 +1,52 @@
+from dataclasses import dataclass
+from typing import Optional
+
+from typing_extensions import Self
+
+from libecalc.presentation.yaml.yaml_entities import YamlDict
+
+
+@dataclass
+class FileMark:
+    line_number: int
+    column_number: int
+
+    def __str__(self) -> str:
+        return f" line {self.line_number}, column {self.column_number}"
+
+
+@dataclass
+class FileContext:
+    start: FileMark
+    end: Optional[FileMark] = None
+    name: Optional[str] = None
+
+    def __str__(self) -> str:
+        file_context_string = ""
+        if self.name is not None:
+            file_context_string += f" in '{self.name}',"
+
+        file_context_string += str(self.start)
+
+        return file_context_string
+
+    @classmethod
+    def from_yaml_dict(cls, data: YamlDict) -> Self:
+        if not hasattr(data, "end_mark") or not hasattr(data, "start_mark"):
+            return None
+
+        # This only works with our implementation of pyyaml read
+        # In the future, we can move this logic into PyYamlYamlModel with a better interface in YamlValidator.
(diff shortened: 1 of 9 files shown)


src/libecalc/presentation/yaml/file_context.py
```diff
@@ -0,0 +1,50 @@
...
+    start: FileMark
+    end: Optional[FileMark] = None
+    name: Optional[str] = None
+
+    def __str__(self) -> str:
+        file_context_string = ""
+        if self.name is not None:
+            file_context_string += f" in '{self.name}',"
+
+        file_context_string += str(self.start)
```
TeeeJay, line 27: not include end, if set?

src/tests/libecalc/presentation/yaml/yaml_models/test_yaml_model_errors.py
```diff
@@ -0,0 +1,39 @@
...
+from libecalc.presentation.yaml.yaml_entities import ResourceStream
+from libecalc.presentation.yaml.yaml_models.exceptions import YamlError
+from libecalc.presentation.yaml.yaml_models.yaml_model import YamlModel, YamlModelType
+
+invalid_models = [
+    (
+        """
+`INSTALLATIONS: []
+""",
+        "invalid_token_start",
```
TeeeJay, line 13: should this message be presented to users? Should be consider making it more understandable? (if so)

src/libecalc/presentation/yaml/file_context.py
```diff
@@ -0,0 +1,50 @@
...
+class FileMark:
+    line_number: int
+    column_number: int
+
+    def __str__(self) -> str:
+        return f" line {self.line_number}, column {self.column_number}"
+
+
+@dataclass
+class FileContext:
```
TeeeJay, line 17: can this also be used for other resources than yaml, such as csv? Relevant?

src/tests/libecalc/presentation/yaml/yaml_models/test_yaml_model_errors.py
jsolaas: That's not the expected value in the test, it is just a description/name of the model. This test only makes sure the error raised is YamlError. I haven't done much with the error messages, so we will show what pyyaml/ruamel reports. That may not be understandable to users.

src/libecalc/presentation/yaml/file_context.py
jsolaas: Not sure, seems a bit confusing to me. We might have to be more specific in the errors here, as sometimes start may mean "the error is here" and other times it means the error is between start and end

src/libecalc/presentation/yaml/file_context.py
jsolaas: Sure, that would make sense to me. I'm hoping we can improve on this when looking into validation (of csv resources)

PR #532, opened 2024-06-18
chore: make sure that pressure calculation tolerances are the same at…

olelod:
… different places in the code

Pressure calculation tolerance in the iteration to find the outlet pressure of a compressor stage had a tolerance of 0.01, while other tolerances inside the code were set to 0.001. This led to edge cases where there should be a solution but seemingly the target pressure of a compressor stage was not met (calculated pressure with too large relative difference from the target pressure).

Removing the constant OUTLET_PRESSURE_CONVERGENCE_TOLERANCE and using the constant PRESSURE_CALCULATION_TOLERANCE everywhere. 

Results have changed in tests because the calculation is more accurate (adding iterations in the calculation of the outlet pressure of a compressor stage). 

Refs.
ECALC-1533


frodehk (approved):

TeeeJay:
@olelod ok to merge - and also to create a patch for 8.20 release as well - to be patched this week

PR #533, opened 2024-06-19
chore: make sure all time series are resampled

frodehk:
ECALC-1003

Currently only results in classes that inherits from TabularTimeSeries are resampled. This means that currently for example time series within stage_results (CompressorStageResult) and inlet/outlet_stream_conditions (CompressorStreamCondition) are not resampled. 

All time series should be resampled to avoid inconsistent results when date frequency is mot data defined.

- [ ] Ensure the remaining classes inherit from `TabularTimeSeries` 
- [ ] Update units to allow for resampling of `KILO_PER_HOUR` and `ACTUAL_VOLUMETRIC_M3_PER_HOUR`
- [ ] Ensure "per hour" is accounted for before resampling

https://equinor-ecalc.atlassian.net/browse/ECALC-1003?atlOrigin=eyJpIjoiZjNmZDE5OGE1ZTM3NGZmZDk3NjYzZGJiOTM3ZDRkZmUiLCJwIjoiaiJ9

### DIFF
--- a/src/libecalc/common/utils/rates.py
+++ b/src/libecalc/common/utils/rates.py
@@ -1038,6 +1038,11 @@ class TimeSeriesRate(TimeSeries[float]):
         if freq is Frequency.NONE:
             return self.model_copy()
 
+        # Resampling everything requires resampling of e.g. stage results, inlet/outlet stream conditions.
+        # These include rates per hour, not only per day. Need to check this and convert to per day:
+        if self.unit in [Unit.KILO_PER_HOUR, Unit.ACTUAL_VOLUMETRIC_M3_PER_HOUR]:
+            self.values = [v * 24 for v in self.values]
+
         # make resampled calendar day volumes via cumulative calendar day volumes
         calendar_day_volumes = (
             TimeSeriesVolumesCumulative(
--- a/src/libecalc/dto/result/tabular_time_series.py
+++ b/src/libecalc/dto/result/tabular_time_series.py
@@ -5,6 +5,7 @@ from typing import List, Optional
 import pandas as pd
 from typing_extensions import Self
 
+import libecalc.dto as dto
 from libecalc.common.time_utils import Frequency, resample_time_steps
 from libecalc.common.units import Unit
 from libecalc.common.utils.rates import (
@@ -18,7 +19,7 @@ from libecalc.dto.result.base import EcalcResultBaseModel
 
 
 class TabularTimeSeries(ABC, EcalcResultBaseModel):
-    name: str
+    name: Optional[str] = None
     timesteps: List[datetime]
 
     def to_dataframe(
@@ -110,6 +111,12 @@ class TabularTimeSeries(ABC, EcalcResultBaseModel):
                 else:
                     # NOTE: Operational settings are not resampled. Should add support?
                     pass
+
+            # Resampling compressor stream results must be done separately:
+            elif isinstance(values, dto.result.results.CompressorStreamConditionResult):
+                for stream_attribute, stream_values in values:
+                    if isinstance(stream_values, TimeSeries):
+                        values.__setattr__(stream_attribute, stream_values.resample(freq=freq))
             else:
                 # NOTE: turbine_result is not resampled. Should add support?
                 pass
--- a/src/libecalc/application/graph_result.py
+++ b/src/libecalc/application/graph_result.py
@@ -634,14 +634,12 @@ class GraphResult:
                                     else [math.nan] * len(model.timesteps),
                                     unit=Unit.NONE,
                                 ),
-                                density_kg_per_m3=TimeSeriesRate(
+                                density_kg_per_m3=TimeSeriesFloat(
                                     timesteps=model.timesteps,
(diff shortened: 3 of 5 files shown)


src/libecalc/dto/result/tabular_time_series.py
```diff
@@ -18,7 +19,7 @@
 
 
 class TabularTimeSeries(ABC, EcalcResultBaseModel):
-    name: str
+    name: Optional[str] = None
```
frodehk, line 22: Will setting name to optional create problems elsewhere?

src/libecalc/common/utils/rates.py
```diff
@@ -1038,6 +1038,11 @@ def resample(
         if freq is Frequency.NONE:
             return self.model_copy()
 
+        # Resampling everything requires resampling of e.g. stage results, inlet/outlet stream conditions.
+        # These include rates per hour, not only per day. Need to check this and convert to per day:
+        if self.unit in [Unit.KILO_PER_HOUR, Unit.ACTUAL_VOLUMETRIC_M3_PER_HOUR]:
+            self.values = [v * 24 for v in self.values]
```
jsolaas, line 1044: This should probably be handled by using the `to_unit` function in TimeSeries.

Can we do the same below for calendar_day_volumes (and stream_day_columes)? `self.to_calendar_day().to_volumes().cumulative().resample().to_volumes()`

Do we have any tests for these functions?

src/libecalc/dto/result/tabular_time_series.py
jsolaas, line 22: Given these changes, TabularTimeSeries shouldn't handle both `to_dataframe` and `resample`. Since name is something we need in `to_dataframe`. This change should include some refactoring to separate them. 

Instead of adding to this, can we move resample into its own class/function? So instead of calling result.resample(), we use Resampler.resample(result). At least that doesn't make to_dataframe more confusing.

src/libecalc/dto/result/tabular_time_series.py
```diff
@@ -110,6 +111,12 @@ def resample(self, freq: Frequency) -> Self:
                 else:
                     # NOTE: Operational settings are not resampled. Should add support?
                     pass
+
+            # Resampling compressor stream results must be done separately:
```
jsolaas, line 115: Why? Since it can't inherit from TabularTimeSeries because it is missing timesteps?

src/libecalc/dto/result/tabular_time_series.py
jsolaas: It might, `to_dataframe` assumes name is set when creating the column names.

src/libecalc/dto/result/tabular_time_series.py
frodehk: I guess my comment is a bit misleading. The comment can be replaced by: "Resampling of compressor stream results happens here". Did it clarify or was it something more you pointed at?

src/libecalc/dto/result/tabular_time_series.py
frodehk: And put the resampler-function in EcalcBaseModel (dto.base)? I guess it is easier to have the resampler as a function in a class that all inherit from, compared to have a separate Resampler-class?

src/libecalc/common/utils/rates.py
frodehk: Not many tests as I am aware of. We have one in test_rates: `test_compute_stream_day_rate` (`Rates.to_stream_day()`).

src/libecalc/dto/result/tabular_time_series.py
frodehk: Changed, it is no longer optional.

src/libecalc/common/utils/rates.py
frodehk: Changed, unit check not necessary.

src/libecalc/dto/result/tabular_time_series.py
frodehk: Changed, name not optional.

src/libecalc/dto/result/tabular_time_series.py
```diff
@@ -110,6 +111,12 @@ def resample(self, freq: Frequency) -> Self:
                 else:
                     # NOTE: Operational settings are not resampled. Should add support?
                     pass
+
+            # Resampling compressor stream results:
+            elif isinstance(values, dto.result.results.CompressorStreamConditionResult):
```
olelod, line 116: Can't you just say `if isinstance(values, TimeSeries) or isinstance(values, TabularTimeSeries):` on line 101 and drop this?

frodehk: Yes of course, nice catch:-)

src/libecalc/dto/result/tabular_time_series.py
frodehk: Removed, added extra check further up.

jsolaas (approved):

PR #541, opened 2024-06-27
chore: improve error message for extra pfs columns

frodehk:
ECALC-1342

- [x] I have added tests 

Error message if illegal category is used for extra power from shore columns (`CABLE_LOSS` and `MAX_USAGE_FROM_SHORE`) is the same, even if only one of them is used in the model. Can be confusing for user.

- [x] Improve error message to point at the problematic column (or both columns if they are in the model)
- [x] Add and update tests

https://equinor-ecalc.atlassian.net/browse/ECALC-1342?atlOrigin=eyJpIjoiMzljZmE4MjlkZWZjNGFmM2EwYmIxNDU4ZGQyNTdkOWUiLCJwIjoiaiJ9


jsolaas (approved):
It is possible to create reusable validators, they seem to be the same? We could also probably create a validator that could run separately for CABLE_LOSS and MAX_USAGE_FROM_SHORE, which would create two errors in the case where both are specified without the correct category.

PR #547, opened 2024-07-02
fix: error message for compressor model in turbine

jsolaas:
Include the turbine name in the error message.

Refs: ECALC-1199


markusrf (approved):

jsolaas:
Not patching this

PR #555, opened 2024-07-25
feat: implement adjustment factors for genset

frodehk:
ECALC-1369

Adjustment factor and adjustment constant are not implemented for genset. It is already implemented as a part of facility inputs.

Make sure fuel consumption is scaled according to the adjustment factor and constant given in facility inputs.

https://equinor-ecalc.atlassian.net/browse/ECALC-1369?atlOrigin=eyJpIjoiM2MxNDJhZjNhZDkyNDUzMGExODg3NjJiNzYzMzg3MmQiLCJwIjoiaiJ9

### DIFF
--- a/src/libecalc/core/models/generator.py
+++ b/src/libecalc/core/models/generator.py
@@ -10,10 +10,16 @@ class GeneratorModelSampled:
         self,
         data_transfer_object: dto.GeneratorSetSampled,
     ):
+        fuel_values = data_transfer_object.fuel_values
+        if data_transfer_object.energy_usage_adjustment_factor is not None:
+            fuel_values = list(np.array(fuel_values) * data_transfer_object.energy_usage_adjustment_factor)
+        if data_transfer_object.energy_usage_adjustment_constant is not None:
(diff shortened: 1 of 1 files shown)


src/libecalc/core/models/generator.py
```diff
@@ -10,10 +10,16 @@ def __init__(
         self,
         data_transfer_object: dto.GeneratorSetSampled,
     ):
+        fuel_values = data_transfer_object.fuel_values
```
TeeeJay, line 13: We are probably doing it the same way other places, but I think that this "manipulation" of the data should be done earlier, when reading the data. Why? Because we alter the data(-file), and not the model itself. If the data file is used other places/for other things, it should still be applied. In theory a file can be used more than one place. Please let me know what you think.

TeeeJay: Although, the way it has been implemented, sending the adjustment factor and constant to the genset, it makes sense to put here...

TeeeJay (approved):

PR #556, opened 2024-07-30
chore: update dependencies

markusrf:


poetry.lock
```diff
@@ -2594,109 +2611,119 @@ files = [
 
 [[package]]
 name = "pydantic"
-version = "2.7.4"
+version = "2.8.2"
```
TeeeJay, line 2614: minor bump pydantic. should take note in case sth breaks with validation etc

TeeeJay (approved):

PR #557, opened 2024-07-31
fix: read requested pressures in compressor system

frodehk:
ECALC-1230

eCalc fails when extracting requested pressures from compressor system, if the date for the temporal model is not included in the timesteps for the input data.

Update method for extracting requested pressures.

https://equinor-ecalc.atlassian.net/browse/ECALC-1230?atlOrigin=eyJpIjoiNTFiNjQ5ZTI2N2ZiNDY5YmJjZGZiM2M3YjM0OGM3YzYiLCJwIjoiaiJ9

### DIFF
--- a/src/libecalc/common/time_utils.py
+++ b/src/libecalc/common/time_utils.py
@@ -70,7 +70,7 @@ class Period:
 
     def get_timesteps(self, timesteps: List[datetime]) -> List[datetime]:
         """
-        Finds all timesteps from global time that are within the period.
+        Get all given timesteps that are within this period.
         Returns empty list if no timesteps are outside period.
         """
         timesteps = [timestep for timestep in timesteps if self.__contains__(timestep)]
--- a/src/libecalc/common/time_utils.py
+++ b/src/libecalc/common/time_utils.py
@@ -74,11 +74,8 @@ class Period:
         Returns empty list if no timesteps are outside period.
         """
         timesteps = [timestep for timestep in timesteps if self.__contains__(timestep)]
-        timesteps_in_period = [
-            timestep for timestep, included in zip(timesteps, timesteps_included_in_period) if included
-        ]
 
-        return timesteps_in_period
+        return timesteps
 
 
 @dataclass
--- a/src/libecalc/common/time_utils.py
+++ b/src/libecalc/common/time_utils.py
@@ -68,6 +68,18 @@ class Period:
                 f"Period: {self.start}:{self.end} - timesteps: {timesteps}"
             ) from e
 
+    def get_timesteps(self, timesteps: List[datetime]) -> List[datetime]:
+        """
+        Finds all timesteps from global time that are within the period.
+        Returns empty list if no timesteps are outside period.
+        """
+        timesteps_included_in_period = [self.__contains__(timestep) for timestep in timesteps]
+        timesteps_in_period = [
+            timestep for timestep, included in zip(timesteps, timesteps_included_in_period) if included
+        ]
+
+        return timesteps_in_period
+
 
 @dataclass
 class Periods:
--- a/src/libecalc/common/time_utils.py
+++ b/src/libecalc/common/time_utils.py
@@ -73,7 +73,7 @@ class Period:
         Finds all timesteps from global time that are within the period.
         Returns empty list if no timesteps are outside period.
         """
-        timesteps_included_in_period = [self.__contains__(timestep) for timestep in timesteps]
+        timesteps = [timestep for timestep in timesteps if self.__contains__(timestep)]
         timesteps_in_period = [
             timestep for timestep, included in zip(timesteps, timesteps_included_in_period) if included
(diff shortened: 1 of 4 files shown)


src/tests/libecalc/dto/results/model/test_requested_pressures_compressors.py
```diff
@@ -116,3 +121,36 @@ def test_requested_pressures_compressor_system_temporal_model(result: EcalcModel
...
+    model_timesteps = [
+        datetime(2018, 1, 1),
+        datetime(2018, 6, 23, 17, 20, 50),
+        datetime(2018, 7, 5, 12, 20, 50),
+        datetime(2022, 1, 1),
+    ]
+
+    operational_settings_used.timesteps = model_timesteps
+
+    # Ensure that method do not fail if model timesteps do not correspond exactly to dates in temporal model
```
TeeeJay, line 149: also check that the evaluated/fetched pressure values are those one expect to get at the different timesteps?

src/libecalc/presentation/json_result/mapper.py
```diff
@@ -79,49 +78,43 @@ def get_requested_compressor_pressures(
...
-                    operational_setting_used_id = get_operational_setting_used_id(
-                        timestep=timestep, operational_settings_used=operational_settings_used
-                    )
 
-                    operational_setting = model_subset.operational_settings[operational_setting_used_id]
+    for period, model in TemporalModel(energy_usage_model).items():
+        if isinstance(model, CompressorSystemConsumerFunction):
+            # Loop timesteps in temporal model, to find correct operational settings used:
+            for timestep in model_timesteps:
+                if timestep >= period.start and timestep <= period.end:
```
TeeeJay, line 86: why is this test needed? Can we get timesteps "outside" the temporal model here? What if that happens? Is that an error (and should be handled) or is it normal and "proceed as normal"?

frodehk: Yes, `model_timesteps` are all timesteps for the component. The test is done in order to ensure that we are within the temporal model. The point is to grab the correct operational setting for the actual timestep.

TeeeJay (changes requested):

src/libecalc/presentation/json_result/mapper.py
TeeeJay: do we have a method like `period.get_timesteps(include_start=True, include_end=True)` that we can use instead to iterate? to avoid iterating _all_ timesteps from global start to global end for all temporal models?

frodehk: A new `get_timesteps` method is now included in `period`.

src/tests/libecalc/dto/results/model/test_requested_pressures_compressors.py
frodehk: Updated to check the pressures as well.

src/libecalc/common/time_utils.py
```diff
@@ -68,6 +68,18 @@ def get_timestep_indices(self, timesteps: List[datetime]) -> Tuple[int, int]:
                 f"Period: {self.start}:{self.end} - timesteps: {timesteps}"
             ) from e
 
+    def get_timesteps(self, timesteps: List[datetime]) -> List[datetime]:
+        """
+        Finds all timesteps from global time that are within the period.
```
TeeeJay, line 73:
```suggestion
        Get all given timesteps  that are within this period.
```

src/libecalc/common/time_utils.py
```diff
@@ -68,6 +68,18 @@ def get_timestep_indices(self, timesteps: List[datetime]) -> Tuple[int, int]:
...
+        Returns empty list if no timesteps are outside period.
+        """
+        timesteps_included_in_period = [self.__contains__(timestep) for timestep in timesteps]
```
TeeeJay, line 76:
```suggestion
        timesteps = [timestep for timestep in timesteps if self.__contains__(timestep)]
```
?

src/libecalc/common/time_utils.py
```diff
@@ -68,6 +68,18 @@ def get_timestep_indices(self, timesteps: List[datetime]) -> Tuple[int, int]:
                 f"Period: {self.start}:{self.end} - timesteps: {timesteps}"
             ) from e
 
+    def get_timesteps(self, timesteps: List[datetime]) -> List[datetime]:
```
TeeeJay, line 71: unit test for this?

TeeeJay (approved):

TeeeJay (approved):
nice, but add a unit test for the new period method?

src/libecalc/common/time_utils.py
frodehk: Unit tests are added

frodehk:
> nice, but add a unit test for the new period method?

Done

PR #567, opened 2024-08-15
Ecalc 1379

olelod:

### DIFF
--- a/src/libecalc/core/consumers/legacy_consumer/component.py
+++ b/src/libecalc/core/consumers/legacy_consumer/component.py
@@ -245,6 +245,19 @@ class Consumer(BaseConsumer):
             temporal_expression=TemporalModel(self._consumer_dto.regularity),
             variables_map=variables_map,
         )
+        # figure out if regularity has been extrapolated and override if it is extrapolated to 0
+        regularity_extrapolated = TemporalExpression.extrapolated(
+            temporal_expression=TemporalModel(self._consumer_dto.regularity),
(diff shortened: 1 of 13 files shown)


src/libecalc/core/consumers/legacy_consumer/component.py
```diff
@@ -245,6 +245,19 @@ def evaluate(
             temporal_expression=TemporalModel(self._consumer_dto.regularity),
             variables_map=variables_map,
         )
+        # figure out if regularity has been extrapolated and override if it is extrapolated to 0
+        regularity_extrapolated = TemporalExpression.extrapolated(
+            temporal_expression=TemporalModel(self._consumer_dto.regularity),
+            variables_map=variables_map,
+        )
+        regularity = [
+            1
```
TeeeJay, line 254: will this also apply to timesteps at the end of the global time vector, if e.g. the input file with regularity ends at 2024, but the global time vector goes to 2026? Often, it is (or was) set explicitly to 0 in the last timestep (for different type of data, regularity, produiction etc), as an explicit end to make sure that extrapolation worked for the few remainign timesteps at the end - and in that case regularity should be 0, and not 1...right?

TeeeJay: I guess this is correct for missign timesteps in the beginning and in the middle of the global time vector, but not at the end. or perhaps those in the middle should be interpolated (as you do I think?)

olelod:
Not relevant anymore

PR #568, opened 2024-08-15
feat: add adjustment factor for models

frodehk:
ECALC-1361

`POWER_ADJUSTMENT_FACTOR` is not yet implemented for models. There are examples where users are applying the `POWERLOSSFACTOR` for calibration of equipment/models, in lack of other alternatives (the power loss factor is not intended for equipment calibration). In some cases this can give problems, e.g. for directly driven compressors (they don´t have any power loss). 

- [x] Implement POWER_ADJUSTMENT_FACTOR for models
- [x] Add tests

https://equinor-ecalc.atlassian.net/browse/ECALC-1361?atlOrigin=eyJpIjoiZTVhYWFkZWE2ZDEwNDg4ZmE1MGQyMTI4MmRjOGI4NzciLCJwIjoiaiJ9


src/libecalc/core/models/compressor/base.py
```diff
@@ -130,6 +130,7 @@ def evaluate_turbine_based_on_compressor_model_result(
             load_adjusted = np.where(
                 np.asarray(compressor_energy_function_result.power) > 0,
                 np.asarray(compressor_energy_function_result.power)
+                * self.data_transfer_object.energy_usage_adjustment_factor
```
TeeeJay, line 133: I think the skip of using factor here was intended, at least it seems like. There was some confusion between 2 different but very similar "adjustment factors" at some point - verify that this is not the case any more. Perhaps check COMPRESSOR_TABULAR or TABULAR?

frodehk: @olelod do you remember if the adjustment factor was intentionally skipped here? And why?

olelod: The ADJUSTMENT (CONSTANT and FACTOR) for FACILILITY_INPUTS are mapped to energy_usage_adjustment_constant and energy_usage_adjustment_factor in the dto. The POWER_ADJUSTMENT_CONSTANT is mapped to energy_usage_adjustment_constant but energy_usage_adjustment_factor is defaulted to 1. I can not remember if that was intentional - and I can not see the reason why it should be like that.

TeeeJay: ok, is it possible to include this in a test to have it covered - and see effect before/after? Also, add this explicitly in changelog to make users aware of this bug - and how this will change results to be correct - also for UAT to have a test to test before/after to verify correctness and expected change :)

src/libecalc/presentation/yaml/yaml_types/models/yaml_compressor_trains.py
```diff
@@ -62,6 +62,11 @@ class YamlSingleSpeedCompressorTrain(YamlCompressorTrainBase):
         description="Constant to adjust power usage in MW",
         title="POWER_ADJUSTMENT_CONSTANT",
     )
+    power_adjustment_factor: float = Field(
```
TeeeJay, line 65: inform in docs about this new feature, and when  (8.22) it was introduced in keyword section for POWER_ADJUSTMENT_FACTOR (and _CONSTANT?)

TeeeJay (changes requested):
please add tests for found bug (and possibly isolate it in separate commit/PR), and for new functionality + update docs and changelog :)

src/libecalc/core/models/compressor/base.py
TeeeJay: now, I am thinking that perhaps the reason it was not set was because it was hardcoded to 1? And in that case it doesnt need to be informed, unless users expected it to work?

src/libecalc/core/models/turbine.py
```diff
@@ -40,6 +40,7 @@ def max_power(self) -> Optional[float]:
         )
 
     def evaluate(self, load: NDArray[np.float64], fuel_lower_heating_value: float = 0) -> TurbineResult:
+        # Why do we divide by the adjustment factor here, and not pure linear adjustment?
```
TeeeJay, line 43: probably incorrect? I guess one has thought opposite of what is correct - similarily to power loss factor perhaps? but it wasnt implemented before right, so it didnt have any effect of dividing by 1?

frodehk: That is true. It was defaulted to 1

frodehk: @olelod do you recall if this is intentional or not? (dividing instead of linear adjustment for turbine)

frodehk: From what I understand, there is a need to adjust/calibrate at turbine level: It is a difference between polytropic efficiency in the compressor, and loss in the turbine. Hence, it is my understanding that it should be possible to adjust them separately.

src/libecalc/core/models/turbine.py
```diff
@@ -34,12 +34,17 @@ def __init__(
...
+        # Note: the scaling is opposite to the turbine load adjustment
         return (
             self._maximum_load * self.data_transfer_object.energy_usage_adjustment_factor
             - self.data_transfer_object.energy_usage_adjustment_constant
         )
 
     def evaluate(self, load: NDArray[np.float64], fuel_lower_heating_value: float = 0) -> TurbineResult:
+        # Calibration of turbine load:
+        # Linear adjustment a´*x + b´, but where a´=1/a and b´=b/a.
+        # This is probably to ensu
```
olelod, line 47: Probably some text missing here

src/tests/libecalc/core/models/compressor_modelling/test_variable_speed_compressor_train_common_shaft.py
```diff
@@ -362,3 +362,24 @@ def test_variable_speed_compressor_train_vs_unisim_methane(variable_speed_compre
...
+
+    compressor_train.data_transfer_object.energy_usage_adjustment_factor = adjustment_factor
+    compressor_train.data_transfer_object.energy_usage_adjustment_constant = adjustment_constant
+
+    result_adjusted = compressor_train.evaluate_rate_ps_pd(
+        rate=np.asarray([7000]),
+        suction_pressure=np.asarray([30]),
+        discharge_pressure=np.asarray([100.0]),
+    )
+    assert result_adjusted.power[0] == result.power[0] * 1.5 + adjustment_constant
```
olelod, line 385:
```suggestion
    assert result_adjusted.power[0] == result.power[0] * adjustment_factor + adjustment_constant
```

src/libecalc/core/models/turbine.py
frodehk, line 47:
```suggestion
```

frodehk:
Closing due to rebase-issues. Continued here: https://github.com/equinor/ecalc/pull/611

PR #576, opened 2024-08-26
ci: cache pre-commit

jsolaas:


.github/workflows/pre-ci.yml
```diff
@@ -18,9 +18,13 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "3.8"
+          cache: "pip"
 
       - name: Install pre-commit
-        run: pip install pre-commit==3.2.2
-
+        run: python -m pip install -r pre-commit-requirements.txt
```
TeeeJay, line 24: why not pip directly but through module? required?

.github/workflows/pre-ci.yml
```diff
@@ -18,9 +18,13 @@ jobs:
...
+      - uses: actions/cache@v4
+        with:
+          path: ~/.cache/pre-commit/
+          key: pre-commit-3|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
```
TeeeJay, line 28: is the cache shared across actions and workflows - so it needs to be unique? that is why you add a -3- to the key? risk of using incorrect cache?

TeeeJay (approved):

.github/workflows/pre-ci.yml
jsolaas: Isn't it safer? You are guaranteed that you are using the correct pip/module

.github/workflows/pre-ci.yml
jsolaas: Shared yes, added a 3 because of pre-commit version. Not sure if it's necessary.

PR #581, opened 2024-08-28
refactor: move target pressure and choking from compressor stage to compressor train

olelod:

Target pressures and pressure before/after choking has been confusing when it is related to the compressor stages. Moving towards defining the fluid stream conditions where we need to know them instead.

This now includes the following:

- Adding inlet/outlet stream conditions to the compressor train (calculations and results). Upstream choking will now happen between the train inlet and the inlet of the first compressor stage. Downstream choking will happen between the outlet of the last compressor stage and the outlet of the compressor train. All recirculation happens within a compressor stage (or over all compressor stages for COMMON_ASV).
- Moving the target pressure calculations to the inlet/outlet of the compressor train, instead of on the inlet/outlet of the compressor stages (and adding status check for target intermediate pressure, when relevant).
- Removing all references to before/after choking, since the fluid stream is now defined everywhere. At a later stage we may also model the chokes explicitly in the code. 
     

ECALC-1426


src/libecalc/core/models/compressor/results.py
```diff
@@ -373,14 +359,120 @@ def from_result_list_to_dto(
...
+        return inlet_stream_condition_train, outlet_stream_condition_train, compressor_stage_result
 
     model_config = ConfigDict(extra="forbid")
 
     @property
     def failure_status(self):
         if not all(r.is_valid for r in self.stage_results):
-            for i, stage in enumerate(self.stage_results):
+            for stage in self.stage_results:
                 if not stage.is_valid:
```
olelod, line 476: After target pressure status is moved to train - stage.within_capacity and stage.is_valid is the same, so fix this if-statement...

src/tests/libecalc/integration/snapshots/test_all_consumer_with_time_slots_models/test_all_consumer_with_time_slots_models_results/all_consumer_with_time_slots_models_v3.json
```diff
@@ -654,8 +671,8 @@
                                 ]
                             },
                             "is_valid": [
-                                false,
-                                false
+                                true,
```
olelod, line 674: The compressor stage operates within its capacity so this individual compressor now has is_valid true. The comparison to target pressures has moved to the compressor train - is_valid is still false on the compressor train.

olelod:
Closing this, since the content has been split into several other pull requests

PR #589, opened 2024-09-04
refactor: move target pressure status from stages to compressor train

olelod:
Move target pressures to be defined only on the compressor train, instead of also on the individual compressor stages. The target pressure status check is then also moved to the compressor train result, not on the individual compressor stage results.

Preparation to add inlet/outlet streams to compressor train result, so we can remove pressure before/after choking from the compressor stage results.

Refs.
ECALC-1582


src/tests/libecalc/integration/snapshots/test_all_consumer_with_time_slots_models/test_all_consumer_with_time_slots_models_results/all_consumer_with_time_slots_models_v3.json
```diff
@@ -654,8 +654,8 @@
                                 ]
                             },
                             "is_valid": [
-                                false,
```
olelod, line 657: is_valid changes for the compressor stage results because the compressor stages operate within capacity, hence the compressor stage result is valid. The compressor train is still not able to meet the required discharge pressure, so is_valid on the compressor train is still false.

jsolaas (approved):

TeeeJay:
@olelod worth a note in changelog I think, since it is not a pure refactor but also a "tiny fix", wrt incorrect flags previously for stages?

olelod:
> @olelod worth a note in changelog I think, since it is not a pure refactor but also a "tiny fix", wrt incorrect flags previously for stages?

Good point. Did not think of that. Will add one.

PR #592, opened 2024-09-04
refactor: return compressor stage result when only stage is calculate…

olelod:
…d for simplified train

We are creating a compressor train result with a single stage when we are really only calculating one stage.


TeeeJay (approved):
I assume that this was not done in order to handle the case with trains in a generic way, even if it was only one stage? @jsolaas any input?

jsolaas (approved):
This seems to be an internal change while calculating? `CompressorTrainSimplified` will still return a train result, right?

PR #599, opened 2024-09-10
refactor: avoid using dtos in core

jsolaas:
Core should not import other stuff except common.


TeeeJay (approved):
Great! Is that all -or just the beginning?

jsolaas:
> Great! Is that all -or just the beginning?

beginning

PR #605, opened 2024-09-12
refactor: introduce resource protocol

jsolaas:
Make YamlModel more flexible by not expecting a specific resource class from ResourceService


src/ecalc_cli/infrastructure/file_resource_service.py
```diff
@@ -0,0 +1,43 @@
...
+from libecalc.presentation.yaml.yaml_entities import MemoryResource
+from libecalc.presentation.yaml.yaml_models.yaml_model import YamlValidator
+
+
+class FileResourceService(ResourceService):
+    def __init__(self, working_directory: Path):
+        self._working_directory = working_directory
+
+    @staticmethod
+    def _read_resource(resource_name: Path, *args, read_func: Callable[..., MemoryResource]):
```
TeeeJay, line 18:
```suggestion
    def _read_resource(resource_name: Path, *args, read_func: Callable[..., MemoryResource]) -> MemoryResource:
```

src/ecalc_cli/infrastructure/file_resource_service.py
```diff
@@ -0,0 +1,43 @@
...
+from libecalc.common.errors.exceptions import EcalcError, HeaderNotFound
+from libecalc.common.logger import logger
+from libecalc.infrastructure.file_io import read_facility_resource, read_timeseries_resource
+from libecalc.presentation.yaml.resource import Resource
+from libecalc.presentation.yaml.resource_service import ResourceService
+from libecalc.presentation.yaml.yaml_entities import MemoryResource
+from libecalc.presentation.yaml.yaml_models.yaml_model import YamlValidator
+
+
+class FileResourceService(ResourceService):
```
TeeeJay, line 13: What if we were to provide a Resource that is not loaded into memory - but that reads from the file on the fly, how would we name the service responsible for that?

src/libecalc/fixtures/case_utils.py
```diff
@@ -3,10 +3,10 @@
 
 from libecalc.fixtures import YamlCase
 from libecalc.infrastructure.file_io import read_resource_from_filepath
-from libecalc.presentation.yaml.yaml_entities import Resource
+from libecalc.presentation.yaml.yaml_entities import MemoryResource
 
 
-def _read_resources(directory: Path, resource_names: List[str]) -> Dict[str, Resource]:
+def _read_resources(directory: Path, resource_names: List[str]) -> Dict[str, MemoryResource]:
```
TeeeJay, line 9: replace this with file_resource_service?

src/libecalc/presentation/yaml/mappers/variables_mapper/time_series_collection_mapper.py
```diff
@@ -53,14 +54,22 @@ def _parse_date(date_input: Union[int, str]) -> datetime:
...
-    return time_vector, columns
+
+def parse_time_series_from_resource(resource: Resource):
+    time_series_resource_headers = resource.get_headers()
+    if EcalcYamlKeywords.date in time_series_resource_headers:
+        # Find the column named "DATE" and use that as time vector
+        time_vector = resource.get_column(EcalcYamlKeywords.date)
+        headers = [header for header in time_series_resource_headers if header != EcalcYamlKeywords.date]
+    else:
+        # Legacy: support random names for time vector as long as it is the first column
```
TeeeJay, line 68: log a warning in order to catch this and possibly inform users?

src/libecalc/presentation/yaml/mappers/facility_input.py
```diff
@@ -45,6 +46,13 @@ def _get_adjustment_factor(data: Dict) -> float:
     return data.get(EcalcYamlKeywords.facility_adjustment, {}).get(EcalcYamlKeywords.facility_adjustment_factor) or 1
 
 
+def _get_column_or_none(resource: Resource, header: str) -> Optional[List[Union[float, int, str]]]:
+    try:
+        return resource.get_column(header)
+    except InvalidResource:
```
TeeeJay, line 52: no logging? when will this occur?

TeeeJay (approved):

src/ecalc_cli/infrastructure/file_resource_service.py
jsolaas: StreamResourceService :eyes: 

Not sure about the naming. `MemoryResource` was something I decided to use after I created this one. Maybe `MemoryResourceService` works better. Not sure what to call the different implementations. Might be easier to name them when we have more.

src/libecalc/fixtures/case_types.py
```diff
@@ -4,12 +4,12 @@
...
 
 from libecalc import dto
-from libecalc.presentation.yaml.yaml_entities import Resource
+from libecalc.presentation.yaml.yaml_entities import MemoryResource
 
 
 @dataclass
 class YamlCase:
-    resources: Dict[str, Resource]
+    resources: Dict[str, MemoryResource]
```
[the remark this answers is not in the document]
jsolaas: I haven't looked at the YamlCase, this change is because of the rename. I'm assuming we could manage with Resource, but somewhere in the tests headers and data properties are probably being used.

src/libecalc/presentation/yaml/mappers/variables_mapper/time_series_collection_mapper.py
jsolaas: Should we? A resource with `Dates` as header instead of `Date` for example.

src/libecalc/fixtures/case_utils.py
jsolaas: That would require a configuration, i.e. the yaml file parsed. We could look into using YamlModel here, with FileResourceService and FileConfigurationService. I would have to look into what YamlCase is used for in our tests.

src/tests/libecalc/input/mappers/variables_mapper/test_timeseries.py
```diff
@@ -152,73 +153,90 @@ def test_valid_time_series_unsorted(self):
...
             ["DATE", "OIL_PROD", "BVBV"],
             [["01.01.2017"], [5016]],
-            "The number of columns provided do not match for header and data: data: 1, headers: 2",
-        ),
-        # headers, data mismatch (-1)
-        (
-            ["DATE", "OIL_PROD"],
-            [["01.01.2017"], [5016], [232]],
-            "The number of columns provided do not match for header and data: data: 2, headers: 1",
+            snapshot("""\
```
jsolaas, line 156: Removed test. Current behavior uses headers as what exists, then look up columns based on that. If we are going to catch this we should validate the resource. 

If this is a case where the user forgot to add the header, it should come up when the column is referenced.

PR #612, opened 2024-09-19
refactor: time series collection, resource handling

jsolaas:

Previously, time_series_collection.py and
time_series_collection_mapper.py did a lot of stuff. In addition to dealing with the resource data and validating that, the yaml data was also validated. As this was bundled it was difficult to reuse some of the logic. Each separate step should now be available by using the correct class.

Create domain objects for TimeSeriesResource, TimeSeriesCollection++. This should make the behavior more clear, and provide more flexibility in the future.


src/libecalc/presentation/yaml/domain/time_series_collection.py
```diff
@@ -0,0 +1,78 @@
...
+    def get_time_vector(self) -> List[datetime]:
+        return self._resource.get_time_vector()
+
+    def get_time_series_references(self) -> List[str]:
+        return self._resource.get_headers()
+
+    def get_time_series(self, reference_id: str) -> TimeSeries:
+        try:
+            return TimeSeries(
+                reference_id=f"{self.name};{reference_id}",
```
TeeeJay, line 51: a bit ambigious when both "name+column_name" and column_name is referenced to as reference_id, in different contexts ... can we call it column_name/id here? and that a reference id is "timeseriescollectionname + column name", to get things straight?

src/libecalc/presentation/yaml/domain/time_series_collections.py
```diff
@@ -0,0 +1,52 @@
...
+from libecalc.presentation.yaml.domain.time_series import TimeSeries
+from libecalc.presentation.yaml.domain.time_series_collection import TimeSeriesCollection
+from libecalc.presentation.yaml.domain.time_series_exceptions import TimeSeriesNotFound
+from libecalc.presentation.yaml.domain.time_series_provider import TimeSeriesProvider
+from libecalc.presentation.yaml.domain.time_series_resource import TimeSeriesResource
+from libecalc.presentation.yaml.resource import Resource
+from libecalc.presentation.yaml.yaml_types.time_series.yaml_time_series import YamlTimeSeriesCollection
+
+
+class TimeSeriesCollections(TimeSeriesProvider):
```
TeeeJay, line 13: does it make sense that both a timeseries and a timeseries collection has the same interface? it makes e.g. reference_id a bit ambigious of what it is - that can lead to errors? also, quite many other metods are needed in each of the classes that doesnt match the interface/mixin/protocol ...

TeeeJay (commented):
very nice job - much better code than it was :)

src/libecalc/presentation/yaml/domain/time_series_collection.py
jsolaas: collection_id and time_series_id? We started using reference_id for resources in the api, in that context it's the name of the resource. 

I agree it's confusing, the way I see it it's just an id, and an id is expected to have meaning only within the context. I.e. time_series.id and collection.id. But it's always useful to then name that id more specifically in other contexts, where you only have the id (time_series_id and collection_id, etc).

src/libecalc/presentation/yaml/domain/time_series_collections.py
jsolaas: It does not make sense, but it _could_ make sense that all domain classes should have an id for example (reference_id in this case). 

The TimeSeriesProvider is only used for TimeSeriesCollections and TimeSeriesCollection. Since both just provide time_series.

So timeseries and timeseries collection should not have the same interface. 

TimeSeriesProvider has potential IMO. As we maybe could separate ResourceService into TimeSeriesProvider and FacilityResourceService (or something). 

Provider vs Service is not something I have thought much about btw.

src/libecalc/common/errors/exceptions.py
```diff
@@ -98,3 +103,20 @@ def __init__(self, header: str):
...
             "Missing column", f"Column matching header '{header}' is missing.", error_type=EcalcErrorType.CLIENT_ERROR
         )
+
+
+class InvalidColumn(InvalidResource):
+    def __init__(self, header: str, message: str, row: int = None):
+        self.header = header
+        self.row = row
+        super().__init__(
+            "Invalid column",
```
TeeeJay, line 113: possible to indicate which column that is invalid?

src/libecalc/presentation/yaml/domain/time_series_resource.py
```diff
@@ -0,0 +1,143 @@
...
+from typing_extensions import Self
+
+from libecalc.common.errors.exceptions import InvalidColumn, InvalidHeader, InvalidResource, NoColumns
+from libecalc.common.string.string_utils import get_duplicates
+from libecalc.presentation.yaml.mappers.variables_mapper.time_series_collection_mapper import parse_time_vector
+from libecalc.presentation.yaml.resource import Resource
+from libecalc.presentation.yaml.yaml_keywords import EcalcYamlKeywords
+
+
+class InvalidTimeSeriesResource(InvalidResource):
```
TeeeJay, line 16: should we use the convention for expceptions to end in Exception or Error? I think we should

TeeeJay (approved):

src/libecalc/common/errors/exceptions.py
jsolaas: It's handled in the message currently. Not sure what to make of the title in these errors vs message.

PR #619, opened 2024-09-23
refactor: introduce periods in variables map

olelod:
Introduce concept of periods in VariablesMap. To make all rates have a start date and an end date. This means that:

- There is a global time vector in VariablesMap providing all the start/end dates for the periods
- VariablesMap.periods provides a list of periods
- time_vector in VariablesMap will contain all but the last time step in the global time vector and can be considered as the start dates for all the periods
- If there is no specific end date given in the YAML file using the END keyword, the last date found in the resource files will be used as the end date (this will typically be the last line in one of the resource files where INFLUENCE_TIME_VECTOR is set to True.
- the last line of values in one resource file will be kept for the period starting that time until the next date found in another file - even if extrapolation is false.  

Refs
ECALC-1585


src/libecalc/common/variables.py
```diff
@@ -30,30 +32,96 @@ class VariablesMap(BaseModel):
...
+        # if it is not empty []
+        if info.data["global_time_vector"] and any(
+            len(v[keys]) != len(info.data["global_time_vector"]) - 1 for keys, values in v.items()
+        ):
+            raise ProgrammingError(
+                "Time series: The number of time steps should be one more than the number of values for each variable."
+                "Values are for periods defined by consecutive start and end dates."
+                "Most likely a bug, report to eCalc Dev Team."
+            )
+        return v
```
jsolaas, line 50:
```suggestion
    @model_validator(mode="after")
    def check_length_of_time_vector_vs_variables(self):
        # time_vector should contain one more item than the number of values in each variable
        # if it is not empty []
        if self.global_time_vector and any(
            len(variable) != len(self.global_time_vector) - 1 for variable in self.variables.values()
        ):
            raise ProgrammingError(
                "Time series: The number of time steps should be one more than the number of values for each variable."
                "Values are for periods defined by consecutive start and end dates."
                "Most likely a bug, report to eCalc Dev Team."
            )
        return self
```

src/libecalc/presentation/exporter/queries.py
```diff
@@ -96,28 +97,30 @@ def query(
...
+                    if Period.intersects(period, installation_graph.variables_map.period):
+                        if self.consumer_categories is None or category in self.consumer_categories:
+                            period = Period.intersection(period, installation_graph.variables_map.period)
+                            fuel_consumer_result = installation_graph.get_energy_result(fuel_consumer.id)
+                            fuel_volumes = (
+                                TimeSeriesRate.from_timeseries_stream_day_rate(
+                                    fuel_consumer_result.energy_usage, regularity=regularity
+                                )
+                                .for_period(period)
+                                .to_volumes_for_period(period)
```
jsolaas, line 109: Not sure why we need both `for_period` and `to_volumes_for_period`. Either `to_volumes_for_period` or `for_period(period).to_volumes()` makes sense to me.

src/libecalc/common/time_utils.py
```diff
@@ -53,6 +53,21 @@ def intersects(first: Period, second: Period) -> bool:
...
+        """
+        Args:
+            first:
+            second:
+
+        Returns:
+
+        """
+        if not Period.intersects(first, second):
+            return Period()
```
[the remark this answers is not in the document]
TeeeJay: In particular here, as Period() would indicate from datetime.min to datetime.max due to defaults. I agree that None makes sense, unless there is an attempt to avoid "null", to be able to also "use the None-Period in further logic", but I cannot come up with what would make sense to have as start and end in that case.

src/libecalc/common/time_utils.py
```diff
@@ -53,6 +53,21 @@ def intersects(first: Period, second: Period) -> bool:
...
+
+        return Period(max(first.start, second.start), min(first.end, second.end))
```
TeeeJay, line 69: nifty :) I have seen "algorithms" for periods/intervals and intersection logic that stretches over "pages" before :)

src/libecalc/common/time_utils.py
```diff
@@ -122,6 +137,11 @@ def get_period(self, time: datetime) -> Period:
 
         raise ProgrammingError(f"Period for date '{time}' not found in periods")
 
+    @property
+    def period(self) -> Period:
+        """Get the period which is the union of all the sub periods"""
+        return Period(start=self.periods[0].start, end=self.periods[-1].end)
```
TeeeJay, line 143: do we know that periods is at least of length 1? I can see that previous methods make same assumption...

src/libecalc/common/utils/rates.py
```diff
@@ -943,8 +943,15 @@ def merge(self, other: TimeSeries) -> TimeSeriesRate:
         )
 
     def for_period(self, period: Period) -> Self:
+        if period.end < self.timesteps[0] or period.start > self.timesteps[-1]:
+            return self.__class__(
```
TeeeJay, line 947: I like that you create a "null class" instead of None, to make it possible to use it, as it is still valid - albeit empty...

TeeeJay:
thanks for great description and nice ground work. remember unit tests for the basic period, periods etc functionality for different (edge) cases.

olelod:
Included in https://github.com/equinor/ecalc/pull/636

PR #623, opened 2024-09-24
chore: create variables map protocol for use in core

frodehk:
ECALC-1664

Introduce protocol with get-methods for VariablesMap, to reduce core-dependency and increase flexibility.

Implement VariablesMapService (Protocol) in variables.

https://equinor-ecalc.atlassian.net/browse/ECALC-1664?atlOrigin=eyJpIjoiNjg4YTVjNjZiZGNiNDE2M2JmNmJjNDcyZTg5MTJjNjciLCJwIjoiaiJ9


src/libecalc/common/variables.py
```diff
@@ -65,3 +66,45 @@ def get_subset_for_timestep(self, current_timestep: datetime) -> VariablesMap:
...
+    def get_time_vector(self) -> [List[datetime]]: ...
+
+    @abc.abstractmethod
+    def get_variables(self) -> Dict[str, List[Annotated[float, Field(allow_inf_nan=False)]]]: ...
+
+    @abc.abstractmethod
+    def get_period(self) -> Period: ...
+
+    @abc.abstractmethod
+    def get_subset(self, start_index: int, end_index: int) -> VariablesMap: ...
```
jsolaas, line 101:
```suggestion
    def get_subset(self, start_index: int, end_index: int) -> VariablesMapService: ...
```

src/libecalc/common/variables.py
```diff
@@ -65,3 +66,45 @@ def get_subset_for_timestep(self, current_timestep: datetime) -> VariablesMap:
...
+
+    @abc.abstractmethod
+    def get_subset_from_period(self) -> VariablesMap: ...
```
jsolaas, line 104:
```suggestion
    def get_subset_from_period(self) -> VariablesMapService: ...
```

src/libecalc/common/variables.py
```diff
@@ -65,3 +66,45 @@ def get_subset_for_timestep(self, current_timestep: datetime) -> VariablesMap:
...
+
+    @abc.abstractmethod
+    def get_subset_for_timestep(self) -> VariablesMap: ...
```
jsolaas, line 107:
```suggestion
    def get_subset_for_timestep(self) -> VariablesMapService: ...
```

src/libecalc/common/variables.py
```diff
@@ -65,3 +66,45 @@ def get_subset_for_timestep(self, current_timestep: datetime) -> VariablesMap:
...
+
+    @abc.abstractmethod
+    def get_variables_map(self) -> VariablesMap: ...
```
jsolaas, line 110:
```suggestion
    def get_variables_map(self) -> VariablesMapService: ...
```

src/libecalc/common/variables.py
```diff
@@ -65,3 +66,45 @@ def get_subset_for_timestep(self, current_timestep: datetime) -> VariablesMap:
...
+
+
+T = TypeVar(
+    "T",
+    bound=Union[List[datetime], Dict[str, List[Annotated[float, Field(allow_inf_nan=False)]]], Period, VariablesMap],
+    covariant=True,
+)
+
+
+class VariablesMapService(Protocol[T]):
```
jsolaas, line 90:
```suggestion
class VariablesMapService(Protocol):
```

src/libecalc/common/variables.py
```diff
@@ -65,3 +66,45 @@ def get_subset_for_timestep(self, current_timestep: datetime) -> VariablesMap:
...
+    @abc.abstractmethod
+    def get_time_vector(self) -> [List[datetime]]: ...
+
+    @abc.abstractmethod
+    def get_variables(self) -> Dict[str, List[Annotated[float, Field(allow_inf_nan=False)]]]: ...
```
jsolaas, line 95:
```suggestion
    def get_variables(self) -> Dict[str, List[float]]: ...
```

I don't think the annotations does anything here?

src/libecalc/common/variables.py
```diff
@@ -65,3 +66,45 @@ def get_subset_for_timestep(self, current_timestep: datetime) -> VariablesMap:
...
+
+    def get_variables_map(self):
+        return self
+
+
+T = TypeVar(
+    "T",
+    bound=Union[List[datetime], Dict[str, List[Annotated[float, Field(allow_inf_nan=False)]]], Period, VariablesMap],
+    covariant=True,
+)
```
jsolaas, line 87: I don't think this does anything. Using a TypeVar would require _some_ input to variables map that decided what the type is, and that the `T` was used as a return type in the methods. Unless I misunderstood something.

jsolaas (commented):
Looks like a good start! I think we should look into redefining the interface also, which would require more work.

If for example we made an evaluate function in `VariablesMapService` we could make evaluating expressions the responsibility of the VariablesService. That way we wouldn't have to expose as much of the internals. 

```
def evaluate(expression_or_variable_or_temporal_expression) -> List[float]: ...
```

An alternative is to get rid of the whole VariablesMap also, and instead let the expressions be able to evaluate itself, I think....

TeeeJay:
> Looks like a good start! I think we should look into redefining the interface also, which would require more work.
> 
> If for example we made an evaluate function in `VariablesMapService` we could make evaluating expressions the responsibility of the VariablesService. That way we wouldn't have to expose as much of the internals.
> 
> ```
> def evaluate(expression_or_variable_or_temporal_expression) -> List[float]: ...
> ```
> 
> An alternative is to get rid of the whole VariablesMap also, and instead let the expressions be able to evaluate itself, I think....

Agree, very good start indeed. I guess the example would also require to send in the relevant timesteps/vector, as that information is only where you need to evaluate the expression etc atm?

We can potentially also rename the "VariablesMap/Service" to something that better fits its responsibility in that case, such as `Evaluator` since it can evaluate "a lot of things" to (list of) floats ... ?

jsolaas (approved):

PR #637, opened 2024-10-03
fix: remove emission intensities

olelod:
ECALC-1699


docs/docs/changelog/next.md
```diff
@@ -1,41 +1,47 @@
...
+  constant and a factor. It can be used to calibrate equipment.
+
+## Fixes
+
+- If a compressor train is not able to reach the requested inlet or outlet pressure, both the compressor stages and
+  the compressor train used to be reported as invalid. This is now changed. If the compressor stages are operating within
+  their capacity, they will now be reported as valid even if the compressor train as a whole is not able to reach
+  the requested inlet or outlet pressures.
+
+- Emission intensities are no longer calculated and reported in the results, due to discrepancies between how it was
```
olelod, line 35: I removed some newline characters, and everything appears as new. Only this paragraph is really changed.

jsolaas (approved):

docs/docs/changelog/next.md
TeeeJay: Change to (Equinor neutral/agnostic, or something similar to ):

 Emission intensities are no longer calculated and reported in the results, due to the following discrepancies:
1) CO2 intensity is defined as the total emission divided by the total HC export. eCalc only calculates the future data, so the correct CO2 intensity should be found by adding the historical to these future data. The hydrocarbon export rate is still reported to help the user calculate the CO2 intensity outside of eCalc.
2) Methane intensity was incorrectly calculated based on HC export. The common definition is the methane emission divided by the natural gas throughput of the wanted equipment- or installation level. This is safest done outside of eCalc to ensure proper definition.

PR #638, opened 2024-10-03
Refactor/yaml domain components

jsolaas:

This pull request is needed to be able to remove dtos. Dtos should be removed to improve flexibility by defining interfaces (abc/protocols) and depend on those interfaces instead of specific classes. 

The responsibilities of different classes should also become more clear. I.e. Yaml-classes should make sure the yaml is valid, these classes should be able to depend on valid yaml classes, and further validate the combination of yaml and resources. 

ReferenceService should also help in defining the different references we can have. Interfaces should be implemented here also, and classes implementing them, instead of pydantic.


src/libecalc/dto/component_graph.py
```diff
@@ -1,15 +1,98 @@
...
+    @abc.abstractmethod
+    def get_components(self) -> List[Component]: ...
+
+    @abc.abstractmethod
+    def evaluate(self) -> None: ...
+
+
+# TODO: Refactor ComponentGraph to EnergyGraph, protocol and composition instead of inherit from graph.
+#       EnergyGraph should know about ProcessGraph.
+
```
jsolaas, line 92: This should be discussed. This is the interfaces we implement for different components. Currently asset and installation don't implement an interface, so the ComponentGraph is a bit inconsistent. We should figure out which types of components we have, or implement an alternative interface.

src/libecalc/presentation/yaml/domain/components/electricity_consumer_component.py
```diff
@@ -0,0 +1,95 @@
...
+    def get_node_info(self) -> NodeInfo:
+        return NodeInfo(
+            id=self.id,
+            name=self.name,
+            component_level=ComponentLevel.CONSUMER,
+            component_type=self._dto_consumer.component_type,
+        )
+
+    def get_power_requirement(self, period: Period = None) -> TimeSeriesStreamDayRate:
+        consumer_result = self._core_consumer.evaluate(self._expression_evaluator)
```
jsolaas, line 82: I think we should aim to remove both dto and core consumer here, and rather treat the models as core.

src/libecalc/presentation/yaml/domain/components/electricity_consumer_component.py
```diff
@@ -0,0 +1,95 @@
...
+        self._ecalc_model_result = consumer_result
+        return consumer_result.component_result.power
+
+    def get_ecalc_model_result(self) -> EcalcModelResult:
```
jsolaas, line 86: I also think we should also aim to remove EcalcModelResult, instead provide specific results such as Streams, fuel usage, power usage, emissions. Per equipment results should also be considered. Instead of having EcalcModelResult for FuelConsumer that contains both fuel and power usage, we could have two sets of results, one that represents the turbine, and another that represents the compressor. Turbine naturally has fuel usage, compressor has power usage. If the fuel consumer doesn't provide power usage, there's only a turbine result.

We should think about this when defining the different types of components we have. each component might contain several components in the `result` context, but not in the evaluation context (Similar to ProcessGraph/ConsumerSystem).

src/libecalc/presentation/yaml/domain/components/venting_emitter_component.py
```diff
@@ -0,0 +1,111 @@
...
+        self._expression_evaluator = expression_evaluator
+        self._yaml_venting_emitter = yaml_venting_emitter
+        self._regularity = regularity
+
+    @property
+    def id(self):
+        return generate_id(self._yaml_venting_emitter.name)
+
+    def get_node_info(self) -> NodeInfo:
+        return NodeInfo(
```
jsolaas, line 38: A better name? Could also include category? What else?

src/libecalc/presentation/yaml/domain/reference_service.py
```diff
@@ -0,0 +1,24 @@
...
+class ReferenceService(Protocol):
+    def get_fuel_reference(self, reference: str) -> FuelType: ...
+
+    def get_generator_set_model(self, reference: str) -> GeneratorSetSampled: ...
+
+    def get_compressor_model(self, reference: str) -> CompressorModel: ...
+
+    def get_pump_model(self, reference: str) -> PumpModel: ...
+
+    def get_tabulated_model(self, reference: str) -> TabulatedData: ...
```
jsolaas, line 24: Should look into creating abstractions here, abc or protocol. Maybe split CompressorModel, as I think it is difficult to define a common interface for all the different models.

Creating classes wrapping dto and core, similar to this PR. I think...

src/libecalc/presentation/yaml/yaml_entities.py
```diff
@@ -30,11 +31,50 @@ def get_column(self, header: str) -> List[Union[float, int, str]]:
...
+
+    def get_generator_set_model(self, reference: str) -> GeneratorSetSampled:
+        model = self._get_model_reference(reference, "generator set model")
+        if not isinstance(model, GeneratorSetSampled):
+            raise InvalidReferenceException("generator set model", reference)
+        return model
+
+    def get_compressor_model(self, reference: str) -> CompressorModel:
+        model = self._get_model_reference(reference, "compressor model")
+        if not isinstance(model, get_args(CompressorModel)):
```
jsolaas, line 62: get_args since it's a union..

src/libecalc/presentation/yaml/yaml_types/components/yaml_installation.py
```diff
@@ -39,7 +39,7 @@ class YamlInstallation(YamlBase):
     )
     category: InstallationUserDefinedCategoryType = CategoryField(None)
     hydrocarbon_export: YamlTemporalModel[YamlExpressionType] = Field(
-        None,
+        0,
```
jsolaas, line 42: Setting defaults in yaml makes it more clear and easier to use the property.

src/libecalc/dto/component_graph.py
TeeeJay: good point. At the same time asset and installation is currently only "containers", and not a part of a network per se. However wrt. power networks and inter-networks they should play a role. I think it makes sense to focus on what is "below installation" here, as the installation and asset currently just "aggregates" the data ...

jsolaas:
Some changes implemented separately

PR #647, opened 2024-10-14
refactor: exporter queries

jsolaas:
The exporter knew too much about the specifics. This commit implements an interface to get specific sets of results, which basically is the query. What's left in the queries is alignment of results and aggregation of attributes.


src/libecalc/presentation/exporter/appliers.py
```diff
@@ -2,32 +2,34 @@
...
         chain modifiers, if needed.
         :param data:
         :return:
         """
-        if self.__modifier is not None:
-            return self._modify(self.__modifier._modify(data))
+        if self.__before_modifier is not None:
+            return self._modify(
+                self.__before_modifier.modify(data)
+            )  # TODO: Remove comment -> Call public modify to allow for infinite chaining, riiiiiiight?
```
jsolaas, line 32: Right?

src/libecalc/presentation/exporter/configs/configs.py
```diff
@@ -623,7 +621,7 @@ def aggregator(frequency: Frequency) -> Aggregator:
                     name="gasTurbineCompressorConsumption",
                     title="Total Power Consumed By Gas-Turbine Driven Compressors",
                     unit=Unit.GIGA_WATT_HOURS,
-                    query=FuelConsumerPowerConsumptionQuery(
+                    query=PowerConsumptionQuery(
```
jsolaas, line 624: Both fuel and el now use the same query

src/libecalc/presentation/exporter/configs/configs.py
```diff
@@ -642,15 +640,15 @@ def aggregator(frequency: Frequency) -> Aggregator:
...
                         producer_categories=["POWER-FROM-SHORE"],
                     ),
                 ),
                 Applier(
                     name="powerSupplyOnshore",
                     title="Power Supply Onshore",
                     unit=Unit.GIGA_WATT_HOURS,
-                    query=PowerSupplyOnshoreQuery(
+                    query=ElectricityGeneratedQuery(
```
jsolaas, line 651: Electricity generated instead of a specific power from shore query

src/libecalc/presentation/exporter/configs/configs.py
```diff
@@ -642,15 +640,15 @@ def aggregator(frequency: Frequency) -> Aggregator:
                     name="fromShoreConsumption",
                     title="Total Electricity Consumed From Power-From-Shore",
                     unit=Unit.GIGA_WATT_HOURS,
-                    query=ElectricityGeneratedQuery(
+                    query=PowerConsumptionQuery(
```
jsolaas, line 643: This should have been consumption. Since default `power` from generator set does not calculate `cable_loss` the generated power == consumed power.

src/libecalc/presentation/exporter/domain/exportable.py
jsolaas, line 1: Any name suggestions?

src/libecalc/presentation/exporter/infrastructure.py
jsolaas, line 1: Not sure where to put this.

src/tests/ecalc_cli/snapshots/test_app/test_new_ltp_export_properly/test.INST_A.ltp.tsv.json
```diff
@@ -463,6 +463,22 @@
...
+    "5":"47.09376",
+    "6":"47.09376",
+    "7":"47.09376",
+    "8":"43.28755",
+    "9":"47.09376",
+    "10":"47.09376",
+    "11":"47.09376",
+    "12":"43.28755",
+    "13":"87.6"
+  },
```
jsolaas, line 481: Here's the change

src/libecalc/presentation/exporter/infrastructure.py
```diff
@@ -0,0 +1,405 @@
...
+    def get_electricity_production(self, unit: Unit) -> AttributeSet:
+        attributes = []
+        for fuel_consumer in self._installation_dto.fuel_consumers:
+            if not isinstance(fuel_consumer, GeneratorSet):
+                continue
+
+            fuel_consumer_result = self._installation_graph.get_energy_result(fuel_consumer.id)
+            assert isinstance(fuel_consumer_result, GeneratorSetResult)
+            consumer_category = TemporalModel(fuel_consumer.user_defined_category)
+            for period, category in consumer_category.items():
```
jsolaas, line 59:
```suggestion
            for period, category in consumer_category.items():
            
                if category == "POWER-FROM-SHORE" and fuel_consumer.cable_loss is None:
                    # Edge case; we don't want to assume cable loss = 0 if power from shore. So we skip reporting produced electricity if cable loss isn't specified
                    continue
```

Adding a check here would solve the extra power supply column, but that is something we want to avoid IMO

src/libecalc/presentation/exporter/configs/configs.py
TeeeJay: good point, I think it is worth adding that comment to the query itself, to avoid misunderstanding or change in name later etc

TeeeJay: or, problably best to add here

src/libecalc/presentation/exporter/infrastructure.py
TeeeJay: I agree that we should not handle edge cases like that, it will never end ...

src/libecalc/presentation/exporter/infrastructure.py
```diff
@@ -0,0 +1,405 @@
...
+                if fuel_consumer.cable_loss is not None:
+                    # TODO: Move this calculation into generator set
```
TeeeJay, line 61: agree

src/libecalc/presentation/exporter/domain/exportable.py
```diff
@@ -0,0 +1,74 @@
...
+    INSTALLATION = "INSTALLATION"
+
+
+class ConsumptionType(str, Enum):
+    FUEL = "FUEL"
+    ELECTRICITY = "ELECTRICITY"
+
+
+@dataclass
+class AttributeMeta:
```
TeeeJay, line 20: not sure if I agree on the naming here, a bit confusing for me (for now at least) - introducing another keyword ... but maybe I just need to get used to it ...

src/libecalc/presentation/exporter/infrastructure.py
```diff
@@ -0,0 +1,405 @@
...
+    ExportableType,
+)
+from libecalc.presentation.yaml.yaml_types.emitters.yaml_venting_emitter import (
+    YamlDirectTypeEmitter,
+    YamlOilTypeEmitter,
+    YamlVentingType,
+)
+
+
+class TimeSeriesAttribute(Attribute):
```
TeeeJay, line 32: I like this idea, and it makes sense to divide query into this imo. I cant help but thinking that this could be useful outside LTP/STP as well, perhaps if it was  a bit more generic/less directly related to LTP/STP? I dont see all the details, but a "usefuel" layer imo

TeeeJay (approved):
nice

TeeeJay (approved):

src/libecalc/presentation/exporter/domain/exportable.py
jsolaas: `Attribute` or `Meta`? I'm open to proposals..it was named before I knew what to put in there. Currently it's just the necessary categories used to filter attributes in the queries.

src/libecalc/presentation/exporter/infrastructure.py
jsolaas: Not sure I understand. I agree that an interface for TimeSeriesAttributes makes sense outside exporter. Currently we are using TimeSeries classes directly (in common, some places we use pydantic functionality, other places we don't). 

I also think the Exportable interface makes sense in an energy usage context, not only for export. But it is a bit specific. Instead of `get_power_consumption` returning an AttributeSet, it might make more sense to have `get_power_consumers` where each power_consumer can answer `get_power_consumption -> Attribute/AttributeSet`. Since we might be interested in other stuff also.

src/libecalc/presentation/exporter/domain/exportable.py
TeeeJay: Both. If I had better names I would have mentioned it ;)

src/libecalc/presentation/exporter/infrastructure.py
TeeeJay: I agree that it is too specific the way it is now, but the "notion" of splitting into "energy producing/consuming infrastructure" - as discussed.

PR #654, opened 2024-10-21
Ecalc 1531 numpy2 ready

kjbrak:

This pull request is needed because the project should be ready for numpy 2 upgrade in the future for its own sake, and in case the Komodo environment will upgrade to numpy 2 eventually.

Note: Since the deprecation warnings are found by running the tests, the refactoring is already tested for.

Refactoring some code related to numpy 2 deprecation warnings.

Ecalc 1531

### DIFF
--- a/src/libecalc/core/models/compressor/train/simplified_train.py
+++ b/src/libecalc/core/models/compressor/train/simplified_train.py
@@ -15,14 +15,9 @@ from libecalc.core.models.compressor.results import (
 )
 from libecalc.core.models.compressor.train.base import CompressorTrainModel
 from libecalc.core.models.compressor.train.chart import VariableSpeedCompressorChart
-from libecalc.core.models.compressor.train.chart.chart_creator import (
-    CompressorChartCreator,
-)
+from libecalc.core.models.compressor.train.chart.chart_creator import CompressorChartCreator
 from libecalc.core.models.compressor.train.fluid import FluidStream
-from libecalc.core.models.compressor.train.stage import (
-    CompressorTrainStage,
-    UndefinedCompressorStage,
-)
+from libecalc.core.models.compressor.train.stage import CompressorTrainStage, UndefinedCompressorStage
 from libecalc.core.models.compressor.train.utils.enthalpy_calculations import (
     calculate_enthalpy_change_head_iteration,
     calculate_polytropic_head_campbell,
@@ -517,12 +512,24 @@ class CompressorTrainSimplifiedKnownStages(CompressorTrainSimplified):
                     extrapolate_heads_below_minimum=False,
                 )
             )
-            polytropic_efficiency = float(
-                compressor_chart.efficiency_as_function_of_rate_and_head(
+
+            try:
+                efficiency_array: NDArray[np.float64] = compressor_chart.efficiency_as_function_of_rate_and_head(
                     rates=np.atleast_1d(maximum_actual_volume_rate),
                     heads=np.atleast_1d(polytropic_head),
                 )
-            )
+
+                if efficiency_array.size == 0:
+                    raise ValueError("Efficiency array is empty, cannot proceed with calculation.")
+
+                polytropic_efficiency = efficiency_array[0]
+
+            except IndexError as e:
+                raise IndexError("Failed to extract efficiency: array index out of range.") from e
+
+            except Exception as e:
+                raise RuntimeError("An unexpected error occurred during efficiency extraction.") from e
+
             polytropic_head = calculate_polytropic_head_campbell(
                 polytropic_efficiency=polytropic_efficiency,
                 kappa=kappa,
--- a/src/libecalc/core/models/compressor/train/simplified_train.py
+++ b/src/libecalc/core/models/compressor/train/simplified_train.py
@@ -506,29 +506,18 @@ class CompressorTrainSimplifiedKnownStages(CompressorTrainSimplified):
         expected_diff = 1e-3
         max_iterations = 20
         while not converged and i < max_iterations:
-            maximum_actual_volume_rate = float(
+            maximum_actual_volume_rate: float = float(
                 maximum_rate_function(
                     heads=polytropic_head,
                     extrapolate_heads_below_minimum=False,
                 )
             )
 
-            try:
-                efficiency_array: NDArray[np.float64] = compressor_chart.efficiency_as_function_of_rate_and_head(
-                    rates=np.atleast_1d(maximum_actual_volume_rate),
-                    heads=np.atleast_1d(polytropic_head),
-                )
-
-                if efficiency_array.size == 0:
-                    raise ValueError("Efficiency array is empty, cannot proceed with calculation.")
-
-                polytropic_efficiency = efficiency_array[0]
-
-            except IndexError as e:
-                raise IndexError("Failed to extract efficiency: array index out of range.") from e
-
-            except Exception as e:
-                raise RuntimeError("An unexpected error occurred during efficiency extraction.") from e
+            efficiency_array: NDArray[np.float64] = compressor_chart.efficiency_as_function_of_rate_and_head(
+                rates=np.atleast_1d(maximum_actual_volume_rate),
+                heads=np.atleast_1d(polytropic_head),
+            )
+            polytropic_efficiency = efficiency_array[0]
 
             polytropic_head = calculate_polytropic_head_campbell(
                 polytropic_efficiency=polytropic_efficiency,
--- a/src/libecalc/presentation/flow_diagram/fde_models.py
+++ b/src/libecalc/presentation/flow_diagram/fde_models.py
@@ -76,4 +76,4 @@ class Node(BaseModel):
     model_config = ConfigDict(frozen=True)
 
 
-FlowDiagram.update_forward_refs()
+FlowDiagram.model_rebuild()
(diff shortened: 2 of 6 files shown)


src/libecalc/core/models/compressor/train/simplified_train.py
```diff
@@ -517,12 +512,24 @@ def calculate_maximum_rate_given_outlet_pressure_single_calculation_point(
...
+                efficiency_array: NDArray[np.float64] = compressor_chart.efficiency_as_function_of_rate_and_head(
                     rates=np.atleast_1d(maximum_actual_volume_rate),
                     heads=np.atleast_1d(polytropic_head),
                 )
-            )
+
+                if efficiency_array.size == 0:
+                    raise ValueError("Efficiency array is empty, cannot proceed with calculation.")
+
+                polytropic_efficiency = efficiency_array[0]
```
TeeeJay, line 525: should it therefore always be of length 1? Is it used for different cases, so in this case it is special and only an array of length 1 is expected?

TeeeJay (commented):
ok, from me, but perhaps Ole Petter is better to check this, he is better with numpy than me :)

TeeeJay:
@olelod when you have time :)

src/tests/libecalc/presentation/yaml/test_venting_emitter.py
```diff
@@ -134,9 +129,22 @@ def test_venting_emitter_oil_volume(variables_map):
...
+        if regularity_array.size == 0:
+            raise ValueError("Regularity array is empty, cannot proceed with test.")
+
+        regularity_evaluated = float(regularity_array[0])
+
+    except IndexError as e:
+        raise IndexError("Failed to evaluate regularity: array index out of range.") from e
+
+    except Exception as e:
+        raise RuntimeError("An unexpected error occurred during regularity evaluation in the test.") from e
```
jsolaas, line 146: Since this is a test we should probably just make it clear what we are expecting.

expected_regularity = 1.0

src/tests/libecalc/presentation/yaml/test_venting_emitter.py
```diff
@@ -134,9 +129,22 @@ def test_venting_emitter_oil_volume(variables_map):
...
+
     expected_result = [oil_value * regularity_evaluated * emission_factor / 1000 for oil_value in oil_values()]
```
jsolaas, line 148: expected_result = [oil_value * expected_regularity * emission_factor / 1000 for oil_value in oil_values()]

src/libecalc/presentation/flow_diagram/fde_models.py
```diff
@@ -76,4 +76,4 @@ class Node(BaseModel):
     model_config = ConfigDict(frozen=True)
 
 
-FlowDiagram.update_forward_refs()
+FlowDiagram.model_rebuild()
```
jsolaas, line 79: I think this can be removed instead of replaced.

src/libecalc/core/models/compressor/train/simplified_train.py
```diff
@@ -517,12 +512,24 @@ def calculate_maximum_rate_given_outlet_pressure_single_calculation_point(
...
+
+            except IndexError as e:
+                raise IndexError("Failed to extract efficiency: array index out of range.") from e
+
+            except Exception as e:
+                raise RuntimeError("An unexpected error occurred during efficiency extraction.") from e
```
jsolaas, line 531: This isn't necessary IMO. Just let it fail on index lookup instead. If this can actually happen we should either make sure it doesn't or provide some guidance to the user by raising an EcalcError with info.

src/tests/libecalc/presentation/yaml/test_venting_emitter.py
kjbrak: Har endret en del på det. Se om det ser greit ut med bruk av assert osv.

src/tests/libecalc/presentation/yaml/test_venting_emitter.py
kjbrak: I have made some changes. Take a look if seems ok with the use of assert etc.

src/libecalc/presentation/flow_diagram/fde_models.py
kjbrak: Why is that? I dont understand.
https://docs.pydantic.dev/latest/concepts/models/#rebuilding-model-schema 
_Pydantic tries to determine when this is necessary automatically and error if it wasn't done, but you may want to call [model_rebuild()](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_rebuild) proactively when dealing with recursive models or generics._

jsolaas: Just because it seemed to work without it. I didn't check the actual schema. I'm guessing it's fine since all the classes will be defined when using the classes.

But seeing as they mention calling it proactively we might as well do that.

src/tests/libecalc/presentation/yaml/test_venting_emitter.py
jsolaas: It's ok, the test will fail if we break venting emitters. But I still think we could make it simpler by just using expected_regularity in expected_result. 

As far as I can see the try except basically ends up testing expressions.

src/tests/libecalc/presentation/yaml/test_venting_emitter.py
```diff
@@ -134,9 +130,18 @@ def test_venting_emitter_oil_volume(variables_map):
...
+        )
+
+        regularity_evaluated = float(regularity_array[0])
+
+    except IndexError as e:
+        raise IndexError("Failed to evaluate regularity: array index out of range.") from e
+
+    assert regularity_evaluated == regularity_expected
+
     expected_result = [oil_value * regularity_evaluated * emission_factor / 1000 for oil_value in oil_values()]
```
jsolaas, line 145:
```suggestion
    expected_result = [oil_value * expected_regularity * emission_factor / 1000 for oil_value in oil_values()]
```

kjbrak: Yes, to not evaluate the regularity at all. Probably the best yes. And the evaluation etc. is tested separately. Ok, will change it.

src/libecalc/core/models/compressor/train/simplified_train.py
olelod: For all other compressor model `_evaluate_rate_ps_pd` has a for loop, so only floats are passed around from there. Maybe we should refactor simplified train to do the same...?

src/libecalc/core/models/compressor/train/simplified_train.py
kjbrak: I simplified error handling. Now its basically like before except indexing the array element in a try-catch and raising IndexError if it doesnt work, but it should work since all relevant values further up are floats. Take a look and see if its ok.

src/libecalc/core/models/compressor/train/simplified_train.py
kjbrak: I simplified error handling and I guess @jsolaas will give feedback on it.

TeeeJay: Not entirely sure what @jsolaas thought, but I also think that try...catch here is not needed as we do not add any more context to the exception, and we do not massage the exception to be sent directly to the user. As Jostein suggests, if we understand why this occurs here (exception), we can add that semantics and change to EcalcError instead where we now for sure that the exception has been handled, and it is safe to show e.g. the message to the user. The user should then probably understand what is the problem...

If we cannot say why the exception occurs, and we are unable to add more semantics/context to it, it is better to leave it (not try ... except)

kjbrak: Ok. Then it will actually be like the way I did it the first time last week before I overwrote the commit and started adding unnecessary complexity. 

I will remove the try catch since there should be no way the array len >1 as long as polytropic_head is a float and the function efficiency_as_function_of_rate_and_head is working correctly.

kjbrak: If @jsolaas and you others are ok with the latest change then its ready to be merged. Let me know it there is still some changes needed or if you see a need for the EcalcError to be used here.

jsolaas (approved):
Looks good!

kjbrak:
There were conflicts during merge. 
I rebased locally and pushed it again. 
* I had some rebase conflicts that I think should be ok now. 
* Had to add a small cleaning commit at the end to fix it. 
* I will let it sit for a day before merging the PR so if anyone wants to take an extra look to make sure the resolution of the conflicts was ok and everything looks fine then that would be nice.

src/tests/libecalc/presentation/yaml/test_venting_emitter.py
```diff
@@ -136,9 +132,18 @@ def test_venting_emitter_oil_volume(variables_map):
...
```
jsolaas, line 147:
```suggestion
    expected_result = [oil_value * regularity_expected * emission_factor / 1000 for oil_value in oil_values()]
```

We can still do this?

kjbrak: Not sure how I overlooked this. It seems one commit got lost in the rebase and I didnt catch it! Thank you.

PR #666, opened 2024-10-23
chore: remove unused code

jsolaas:
When looping consumer results we won't get installations and assets,
as we are calculating them after based on the results of sub components.


olelod (approved):
:bulb:

PR #671, opened 2024-10-25
Run STP snapshots in docker container on x86 arch

TeeeJay:

- [x] I have remembered to update manual changelog (`docs/drafts/next.draft.md`)

To run STP/LTP tests in Docker since there are discrepancies between ARM and x86 for Python 3.11. Always runnign this in x86 in container avoids this problem. We are ok with teh differences, they are negligible.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -1,0 +1,34 @@
+
+# libecalc test directory
+
+This directory contains a convenience lightweight Dockerfile to run tests for libecalc. The original reason
+behind this need is to run snapshot tests in an x86_64 environment, since developer may run on ARM64 architecture.
+Some of the snapshot tests are architecture dependent, and running them on an ARM64 machine will result in
+different values than for x86_64. This is due to Delanay triangulation algorithm used in some of the
+compressor calculations.
+
+## Running tests
+
+Since running tests in a container is only required for some of the tests, the default setup is already
+configured in *docker-compose.yml* file. To run tests, simply run (from test/ dir):
+
+```bash
+docker compose run test
+```
+
+or (since we currently only have one service, and default is to run the docker-snapshot test)
+
+```bash
+docker compose up
+```
+
+This will build the container and run the tests marked 'dockersnapshot'. If the snapshot has changed, it will fail with exit code 1, and
+update snapshots. Rerun the test again, with the command above, to make sure it passes.
+
+## Alternative test
+
+If you want to run the dockersnapshot tests and update snapshots without docker compose, and ad-hoc, you can run the following command, from libecalc root:
+
+```bash
+docker run --rm -v .:/project/libecalc -w /project/libecalc -t $(docker build -q . -f Dockerfile --target build) poetry run pytest -m dockersnapshot --snapshot-update
+```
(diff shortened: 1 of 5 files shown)


kjbrak:
* Would it be possible to make use of the existing Dockerfile and create a shell script that runs it with a custom command such as CMD ["poetry", "run", "pytest"] ?

* Could it be better to make custom pytest markers named e.g. docker_snapshot and have the docker-compose file have a command to run only the tests with this marker so its more generic?
command: ["poetry", "run", "pytest", "-m", "docker_snapshot"--snapshot-update"]

TeeeJay:
> * Would it be possible to make use of the existing Dockerfile and create a shell script (e.g. run_tests_in_docker.sh) that runs it with a custom command  CMD ["poetry", "run", "pytest"] ?

that was my initial plan. I can do that! :)

>     * Could it be better to make custom pytest markers named e.g. docker_snapshot and have the docker-compose file have a command to run only the tests with this marker so its more generic?
>       command: ["poetry", "run", "pytest", "-m", "docker_snapshot", "--snapshot-update"]

I just realized that I forgot that I wanted to leave that test out when not running in docker, so yes, then a marker is the way to go. Its probably fast enough to use the marker instead of directly referencing the test as you suggest as well, will check.

kjbrak:
> > * Would it be possible to make use of the existing Dockerfile and create a shell script (e.g. run_tests_in_docker.sh) that runs it with a custom command  CMD ["poetry", "run", "pytest"] ?
> 
> that was my initial plan. I can do that! :)

FYI. This command is running test for me with the existing Dockerfile:
`docker run --rm -v .:/project/libecalc -w /project/libecalc -t $(docker build -q . -f Dockerfile --target build) poetry run pytest -m snapshot`

test/README.md
```diff
@@ -0,0 +1,28 @@
...
+different values than for x86_64. This is due to Delanay triangulation algorithm used in some of the
+compressor calculations.
+
+## Running tests
+
+Since running tests in a container is only required for some of the tests, the default setup is already
+configured in *docker-compose.yml* file. To run tests, simply run:
+
+```bash
+docker compose up
```
TeeeJay, line 16:
```suggestion
docker compose run test
```

instead, to make it clearer that we just run the one service? Also, I support your suggestion to add a .sh file/script instead, which is probably more intuitive.

TeeeJay:
> > > * Would it be possible to make use of the existing Dockerfile and create a shell script (e.g. run_tests_in_docker.sh) that runs it with a custom command  CMD ["poetry", "run", "pytest"] ?
> > 
> > 
> > that was my initial plan. I can do that! :)
> 
> FYI. This command is running test for me with the existing Dockerfile: `docker run --rm -v .:/project/libecalc -w /project/libecalc -t $(docker build -q . -f Dockerfile --target build) poetry run pytest -m snapshot`

should probably also negate the default, to not run tests marked with docker-snapshot as default locally ...

test/README.md
```diff
@@ -0,0 +1,34 @@
+
+# libecalc test directory
+
+This directory contains a convenience lightweight Dockerfile to run tests for libecalc. The original reason
```
markusrf, line 4: No Dockerfile anymore

markusrf (approved):

PR #682, opened 2024-11-01
Check factory for testdata

frodehk:

Check factory method for generating test data


frodehk:
Will not be implemented now

PR #687, opened 2024-11-06
test: allow both validation and not in yaml builders

jsolaas:
We need a way to construct invalid yaml when testing error messages.

Moved the build method to the parent Builder class.


src/tests/libecalc/presentation/exporter/test_ltp.py
```diff
@@ -399,29 +403,45 @@ def test_only_venting_emitters_no_fuelconsumers():
     assert emissions_ch4 == emissions_ch4_asset
 
 
-def test_no_emitters_or_fuelconsumers():
+class EmptyResourceService(ResourceService):
+    def get_resources(self, configuration: YamlValidator) -> dict[str, Resource]:
+        return {}
+
+
+def test_no_emitters_or_fuelconsumers(
```
jsolaas, line 411: This test should probably be moved somewhere else

frodehk: Agree, maybe in https://github.com/equinor/ecalc/blob/main/src/tests/libecalc/presentation/yaml/test_venting_emitter.py

frodehk (approved):
Nice. Good approach. It will facilitate more consistent testing.

PR #688, opened 2024-11-07
chore/move test directory

jsolaas:
- **chore: move tests outside src directory**


src/libecalc/testing/yaml_builder.py
```diff
@@ -115,32 +111,6 @@ def with_consumption_rate_type(self, consumption_rate_type: ConsumptionRateType)
...
-
-    def with_time_series(self, time_series):
-        self.time_series = time_series
-        return self
-
-    def with_facility_inputs(self, facility_inputs) -> Self:
-        self.facility_inputs = facility_inputs
-        return self
-
-
```
jsolaas, line 143: Not used (yet).

Had to make some changes in yaml_builder since mypy now checks it. I guess we don't run mypy on tests..

frodehk (approved):

PR #700, opened 2024-11-14
refactor: create energy interfaces

jsolaas:
Introduce and use energy related interfaces in EnergyCalculator in
order to not require dtos in energy calculator.


src/libecalc/application/energy_calculator.py
```diff
@@ -41,39 +44,82 @@
...
+                    consumes=energy_component.consumes,
                     energy_usage_model=TemporalModel(
                         {
                             period: EnergyModelMapper.from_dto_to_domain(model)
-                            for period, model in component_dto.energy_usage_model.items()
+                            for period, model in energy_component.energy_usage_model.items()
                         }
                     ),
                 )
-                consumer_results[component_dto.id] = consumer.evaluate(expression_evaluator=variables_map)
```
jsolaas, line 72: Next -> move the contents of each if into the corresponding dto in order to not depend on dto in energy calculator. Need to pass expression_evaluator as a parameter

frodehk (approved):
Nice!

PR #701, opened 2024-11-15
refactor/remove dto imports fluid turbine

jsolaas:
- **refactor: move EosModel into fluid**
- **refactor: move fluid.py into common**
- **refactor: remove dto import in core turbine**


src/libecalc/core/models/compressor/results.py
```diff
@@ -5,6 +5,8 @@
 import numpy as np
 from pydantic import BaseModel, ConfigDict
 
+from libecalc.common.fluid import FluidComposition
+from libecalc.common.fluid import FluidStream as FluidStreamDTO
```
olelod, line 9: Why still import it as FluidStreamDTO? Just let it be FluidStream?

PR #704, opened 2024-11-18
chore: remove unused methods

TeeeJay:
unused? avoid using because it is not intuitive? complicates things?

Refs: ECALC-1830

### DIFF
--- a/src/ecalc_cli/io/output.py
+++ b/src/ecalc_cli/io/output.py
@@ -151,9 +151,12 @@ def export_tsv(
     prognosis_filter = config.filter(frequency=frequency)
     result = prognosis_filter.filter(ExportableGraphResult(results), resampled_periods)
 
-    row_based_data: dict[str, list[str]] = CSVFormatter(
-        separation_character="\t", index_formatters=PeriodFormatterConfig.get_row_index_formatters()
-    ).format_groups(result)
+    row_based_data: dict[str, list[str]] = {
+        key: CSVFormatter(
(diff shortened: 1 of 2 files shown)


src/ecalc_cli/io/output.py
```diff
@@ -151,9 +151,12 @@ def export_tsv(
...
     result = prognosis_filter.filter(ExportableGraphResult(results), resampled_periods)
 
-    row_based_data: dict[str, list[str]] = CSVFormatter(
-        separation_character="\t", index_formatters=PeriodFormatterConfig.get_row_index_formatters()
-    ).format_groups(result)
+    row_based_data: dict[str, list[str]] = {
+        key: CSVFormatter(
+            separation_character="\t", index_formatters=PeriodFormatterConfig.get_row_index_formatters()
+        ).format(val)
+        for key, val in result.items()
```
jsolaas, line 158:
```suggestion
        for key, val in result.groups
```

Maybe we should move the loop in groups in here instead, since FilteredResult don't need to know about Formattable. Unnecessary coupling

jsolaas (approved):

src/ecalc_cli/io/output.py
jsolaas: I'm struggling with the directory structure we currently have. Should Formattable be placed in domain, FormattableGroupedQuery in infrastructure etc.? Or should we have a similar structure within presentation.exporter? What is "important" enough to be considered domain. When I think about it I just want to separate interfaces from implementations, so formattable could be in the domain, but it isn't what I think of as the core ecalc domain.....Should Formatter be a completely separate module?

PR #707, opened 2024-11-20
refactor: implement emitter interface

jsolaas:
Move emission calculation into Emitters


src/libecalc/application/energy_calculator.py
```diff
@@ -230,34 +220,14 @@ def evaluate_emissions(self) -> dict[str, dict[str, EmissionResult]]:
...
 
-                for emission_name, emission_rate in emission_rates.items():
-                    emission_result = EmissionResult(
-                        name=emission_name,
-                        periods=self._expression_evaluator.get_periods(),
-                        rate=emission_rate,
-                    )
-                    venting_emitter_results[emission_name] = emission_result
-                emission_results[energy_component.id] = venting_emitter_results
+                if emission_result is not None:
```
TeeeJay, line 230: why can it be None?

src/libecalc/dto/components.py
```diff
@@ -167,11 +173,26 @@ class FuelConsumer(BaseConsumer):
...
+    def evaluate_emissions(
+        self,
+        energy_context: ComponentEnergyContext,
+        energy_model: EnergyModel,
+        expression_evaluator: ExpressionEvaluator,
+    ) -> Optional[dict[str, EmissionResult]]:
+        fuel_model = FuelModel(self.fuel)
+        fuel_usage = energy_context.get_fuel_usage()
+
+        assert fuel_usage is not None
```
TeeeJay, line 189: I would like us to have a discussion around assertions. I have always considered them bad in production code, but perhaps it is ok? I want to shed some light on that :)

src/libecalc/dto/components.py
```diff
@@ -283,6 +304,23 @@ class ConsumerSystem(BaseConsumer):
     stream_conditions_priorities: Priorities[SystemStreamConditions]
     consumers: Union[list[CompressorComponent], list[PumpComponent]]
 
+    def evaluate_emissions(
+        self,
+        energy_context: ComponentEnergyContext,
+        energy_model: EnergyModel,
+        expression_evaluator: ExpressionEvaluator,
+    ) -> Optional[dict[str, EmissionResult]]:
+        if self.consumes == ConsumptionType.FUEL and self.fuel is not None:
```
TeeeJay, line 313: if self.is_fuelconsumer():

TeeeJay: if fuel is none, isnt that an error?

TeeeJay (approved):
nice simplification

src/libecalc/application/energy_calculator.py
jsolaas: Because of consumer system v2 which can be both fuel and el, so a class that might be an emitter.

src/libecalc/dto/components.py
jsolaas: Sure! Let's discuss

PR #709, opened 2024-11-22
feat: allow cross overs both in and out of the same compressor train …

olelod:
…in a compressor system

This can be done if it is made sure that the compressor trains in the compressor system are processed in the right order. You can only pass rates to a compressor train that has not been processed yet. Hence, the indexes in the list specifying the cross overs must be increasing (or zero, meaning no cross over).

A case with a compressor system with three compressor trains (A, B, C). Send everything to A first, if it is too much pass the exceeding rate on to train B, if it is too much for train B also, pass the remaining rate on to train C.


src/libecalc/presentation/yaml/yaml_types/components/legacy/energy_usage_model/yaml_energy_usage_model_consumer_system.py
```diff
@@ -65,6 +65,24 @@ class YamlCompressorSystemOperationalSetting(YamlBase):
         description="Set suction pressure equal for all consumers in a consumer system operational setting. \n\n$ECALC_DOCS_KEYWORDS_URL/SUCTION_PRESSURE",
     )
 
+    @model_validator(mode="after")
+    def ensure_increasing_cross_over(self):
+        if self.crossover is None:
+            return self
+        for index, crossover in enumerate(self.crossover):
+            if crossover >= max(self.crossover[: index + 1]) or crossover == 0:
```
TeeeJay, line 73:
```suggestion
            if self.crossovers_exhausted(index, crossover) or self.has_no_crossovers():
```
or similar? more fluent and easier to read? too use more of those types of fluent typing/syntactic sugar?

src/libecalc/presentation/yaml/yaml_types/components/legacy/energy_usage_model/yaml_energy_usage_model_consumer_system.py
```diff
@@ -65,6 +65,36 @@ class YamlCompressorSystemOperationalSetting(YamlBase):
...
+            )  # compressor_train_index is 0-indexed, crossover_to is 1-indexed
+            no_crossover = crossover_to == 0
+            if crossover_to_compressor_train_index > compressor_train_index or no_crossover:
+                pass  # passing excess rate to a comp. train not yet evaluated or no crossover defined for this comp. train
+            else:
+                raise ValueError(
+                    f"Crossover: {self.crossover}\n"
+                    "The compressor trains in the compressor system are not defined in the correct order, according to "
+                    "the way the crossovers are set up. eCalc can now try to pass excess rates to compressor trains in "
+                    "the system that has already been evaluated. \n\n"
```
jsolaas, line 84: Is this correct? Or should it be not evaluated?

src/libecalc/presentation/yaml/yaml_types/components/legacy/energy_usage_model/yaml_energy_usage_model_consumer_system.py
```diff
@@ -65,6 +65,36 @@ class YamlCompressorSystemOperationalSetting(YamlBase):
...
+                    "To avoid loops and to avoid passing rates to compressor trains that have already been "
+                    "processed, the index of the crossovers should be either 0 (no crossover) or larger than the index "
+                    "of the current compressor train (passing excess rate to a compressor train not yet evaluated). \n\n"
+                    "CROSSOVER: [2, 3, 0] is valid, but CROSSOVER: [2, 1, 0] is not. \n"
```
jsolaas, line 88: v2 does the sorting for the user, just sayin'

src/libecalc/presentation/yaml/yaml_types/components/legacy/energy_usage_model/yaml_energy_usage_model_consumer_system.py
olelod: The way it works now if you have crossover: [0, 1, 0], you first evaluate the first consumer (here it has no crossover). The next consumers can pass excess rate to the first one, potentially exceeding the maximum rate there, and the operational setting will be chosen even if it is invalid.

jsolaas: The sentence `eCalc can now try to pass excess rates to compressor trains the system that has already been evaluated` confused me. Is the intention to inform the user why the current crossover setting isn't allowed? So `If we were to allow this crossover setup, eCalc™ could try to pass excess rates to compressor trains in the system that has already been evaluated`?

jsolaas (approved):

src/libecalc/presentation/yaml/yaml_types/components/legacy/energy_usage_model/yaml_energy_usage_model_consumer_system.py
olelod: Poorly phrased, I guess... Yes, that was happening in one model. The first train is within capacity, the second one outside capacity, so rate was passed to the first one (which then with the additional rate was outside capacity).

PR #714, opened 2024-11-26
refactor: use builders in venting emitter tests

frodehk:
246

Yaml builders should replace the dtos (data transfer objects) in the current tests for venting emitters.

- [x] Replace dtos with builders in tests for venting emitters
- [x] Delete old yaml "factory" for venting emitters
- [x] Delete utilities.py in ltp_export
- [x] Move common functions for ltp from LptTestHelper class up to conftest as a ltp_functions fixture

https://github.com/orgs/equinor/projects/869/views/5?pane=issue&itemId=88720060&issue=equinor%7Cecalc-internal%7C246


tests/libecalc/presentation/yaml/test_venting_emitter.py
```diff
@@ -1,18 +1,15 @@
+from copy import deepcopy
```
jsolaas, line 1: Based on the usage of ltp_functions in this test, it seems that this also tests ltp. So we could move the tests to exporter/test_ltp or create a new file in exporter for test_venting_emitter_ltp. That way ltp functions can be a common fixture within exporter, instead of top-level conftest

jsolaas (approved):

tests/libecalc/presentation/yaml/test_venting_emitter.py
frodehk: Some emitter tests moved to ltp, and fixture is moved back to LtpTestHelper

jsolaas (approved):

PR #722, opened 2024-12-03
chore: new builders for compressors multiple streams and pressures

frodehk:

This is just a draft PR to save some work, that may be needed in time.


frodehk:
Check need for this after refactoring.

PR #723, opened 2024-12-03
create turbine builder and update turbine factory

frodehk:

Part of replacing dtos with builders in tests.

- [ ] Create yaml builder for turbine
- [ ] Update core turbine_factory to use builder.


tests/libecalc/core/conftest.py
```diff
@@ -285,7 +287,20 @@ def turbine_dto() -> dto.Turbine:
 
 
 @pytest.fixture
-def turbine_factory(turbine_dto):
+def yaml_turbine() -> YamlTurbine:
+    return (
+        YamlTurbineBuilder()
+        .with_name("compressor_train_turbine")
+        .with_turbine_loads([0, 2.352, 4.589, 6.853, 9.125, 11.399, 13.673, 15.947, 18.223, 20.496, 22.767])
```
TeeeJay, line 294: should loads and efficiencies be "merged"  or does it make sense to provide them independently, to be able to use different setups? also, in this case as far as I can see, you use the same parameters as with the `with_test_data` method, use that instead ?

TeeeJay (approved):

PR #728, opened 2024-12-09
fix/ensure unique timeseries

jsolaas:
- **fix: ensure unique time_series names**
- **chore: minor fixes in builders**
- **fix: ensure unique names across facility_inputs and models**
- **fix: ensure emission names are unique**


TeeeJay (approved):
nice :) In all of these cases, a duplicate would have led to the last one is taken account, or is the models/facility_inputs an exception that wouldnt have led to error?

jsolaas:
> nice :) In all of these cases, a duplicate would have led to the last one is taken account, or is the models/facility_inputs an exception that wouldnt have led to error?

We had some checks in dtos and in create_references which would have caught a lot of these cases, but not always when validating (only run). If we didn't have those I'm guessing we create a dict in most cases, which would then only consider the last (that's the case for facility/models at least).

PR #730, opened 2024-12-10
fix: don't raise error when missing time series on token collection

jsolaas:
When collecting the tokens from time series collection, we don't want to
raise an error since that would limit the information we get from
validation. Instead we want to collect the tokens we can, then validate
the model based on that context. The errors will still give information
about a missing resource, but also additional information, and file
context.

### DIFF
--- a/src/libecalc/presentation/yaml/domain/time_series_collections.py
+++ b/src/libecalc/presentation/yaml/domain/time_series_collections.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+from typing import Self
 
 from libecalc.presentation.yaml.domain.time_series import TimeSeries
 from libecalc.presentation.yaml.domain.time_series_collection import TimeSeriesCollection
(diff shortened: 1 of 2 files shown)


src/libecalc/presentation/yaml/domain/time_series_collections.py
jsolaas, line 1: Weird diff. I moved init logic to create, with an additional `raise_on_error` flag.

TeeeJay (approved):

src/libecalc/presentation/yaml/model.py
```diff
@@ -122,7 +122,13 @@ def end(self) -> Optional[datetime]:
         return self._configuration.end
 
     def _get_time_series_collections(self) -> TimeSeriesCollections:
-        return TimeSeriesCollections(time_series=self._configuration.time_series, resources=self.resources)
+        time_series_collections, err = TimeSeriesCollections.create(
+            time_series=self._configuration.time_series,
+            resources=self.resources,
+            raise_on_error=True,
+        )
+        assert len(err) == 0
```
jsolaas, line 130: raise_on_error should make this empty

PR #732, opened 2024-12-11
docs: remove api reference

frodehk:

API reference in docs is currently not in use. Should be removed until it needed.

Remove `Generate API reference` in workflow and remove link to api reference in docs.


TeeeJay (approved):
also delete the make-api-reference.py ?

PR #734, opened 2024-12-12
chore: upgrade dependencies

markusrf:


TeeeJay (changes requested):
change pydantic dep to <2.10 due to change of literal of enum to string, instead of constr str with value

TeeeJay (approved):

PR #737, opened 2024-12-13
refactor: move to energy components

frodehk:

Refactor parts of core to allow for improved and more flexible validation, removing dtos, avoid circular imports, remove/merge duplicate classes. The dto layer should be kept, but pydantic validation should be removed.

- [x] Move energy components (e.g. FuelConsumer, ElectricityConsumer) from components.py (dto) to domain/infrastructure/energy_components. 
- [x] Remove components.py


TeeeJay (approved):

TeeeJay (commented):
Nice, but see if you next time is able to keep git history, e.g. that most of the code is detected as a move, and not delete and create. E.g. move first, stage the changes so that git sees that, and then make changes/refactor, if possible :)

PR #738, opened 2024-12-13
Ecalc 1687 docs tests coupling

kjbrak:

Organization of tests and coupling between docs and tests to avoid redundancy of example test data and to assure example cases in docs are tested.

- Using only one version of advanced and simple example in libecalc example folder. Making sure to use the correct version and deleting the other one.
- Modifying the mdx files to account for changes in yaml files etc.
- Modifying the yaml files to make tests pass if needed.
- Adding config plugin in docusaurus for example alias to make things more neat and smoother to update in the future.
- Include Drogon in libecalc examples and integrate it into the testing framework. Add some basic tests such as validate etc.

ECALC-1687


docs/docusaurus.config.ts
```diff
@@ -147,6 +148,16 @@ const config: Config = {
       }),
     ],
   ],
+  plugins: [
+    [
+        'docusaurus-plugin-module-alias',
+        {
+            alias: {
+                '@examples': resolve(__dirname, '../src/libecalc/examples/'),
```
jsolaas, line 156: Very nice!

jsolaas (approved):
Looks good! Now it should be easy to add a test that makes sure our examples does not have invalid results!

I didn't look at the yaml changes or snapshot changes :eyes:

frodehk (approved):
Nice work!

frodehk:
> nice. @frodehk what you think? you are more into this

The yaml changes are in line with the last update of the advanced model, as far as I can see. Including an update with CONTROL_MARGIN for required components.

docs/docusaurus.config.ts
kjbrak: FYI. I first tried to implement the webpack alias as a small custom plugin (in its own folder/files) to avoid 3rd party dependency. But that did not work. I then tried this plugin and it worked as intended.

kjbrak: I just updated the snapshot for that one test since it failed after the changes. I am not sure how to validate if it is correct or not?

PR #741, opened 2024-12-17
fix: delta profile when 2 result sets have different periods

TeeeJay:
due to opposite check of what it was supposed to, ie. check that
we only include periods that are given in the common period vector,
but instead there was a check that the separate result had a
period that was not in the common period vector.

Resolves equinor/ecalc-internal#298

### DIFF
--- a/src/libecalc/presentation/simple_result/simple.py
+++ b/src/libecalc/presentation/simple_result/simple.py
@@ -4,7 +4,6 @@ from pydantic import BaseModel, ConfigDict, model_validator
 
 from libecalc.common.component_info.component_level import ComponentLevel
 from libecalc.common.component_type import ComponentType
-from libecalc.common.errors.exceptions import ProgrammingError
 from libecalc.common.logger import logger
 from libecalc.common.string.string_utils import to_camel_case
 from libecalc.common.time_utils import Period, Periods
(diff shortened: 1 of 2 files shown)


krisolba1 (approved):

src/libecalc/presentation/simple_result/simple.py
```diff
@@ -159,10 +164,9 @@ def fit_to_periods(
                         # Assume index exist if emission exist
                         emission.rate.append(component.emissions[emission.name].rate[period_index])
             else:
-                # This is a developer error, we should provide the correct period.
-                raise ProgrammingError(
-                    f"Provided periods includes period not found in component {component.id}. "
-                    f"Extraneous period: {period}. This should not happen, contact support."
+                # We do not trim extraneous periods in beginning and end for a component. We only try to fit to the common global period.
```
jsolaas, line 167: I don't understand this comment. "We do not trim extraneous periods in beginning and end for a component", does that refer to this code? Or to normalize_components?

jsolaas (approved):

src/libecalc/presentation/simple_result/simple.py
TeeeJay: ok, perhaps misleading. referring to that the components we get as input still may have "incompatible" start and end dates, but it is actually this code that trims/crops those...

PR #763, opened 2025-01-14
chore: update validation error to allow string or none

frodehk:

Update validation error to allow string or None. Avoid TypeScript errors when trying to assign to assign string | null to location.

### DIFF
--- a/src/libecalc/presentation/yaml/validation_errors.py
+++ b/src/libecalc/presentation/yaml/validation_errors.py
@@ -35,44 +35,6 @@ def _get_position_in_file_message(mark: Mark) -> str:
     return message + "\n"
 
 
-class ValidationError(Exception):
-    def __init__(self, message: Optional[str] = None, location: Optional[str] = None):
-        self._location = location
-        self._message = message
-        super().__init__(message)
-
-    @property
-    def location(self) -> Optional[str]:
-        return self._location
-
-    @property
-    def message(self) -> Optional[str]:
-        return self._message
-
-
-def dict_node_representer(dumper: Dumper, data):
-    return dumper.represent_dict(dict(data))
-
-
-yaml.add_representer(YamlDict, dict_node_representer)
-
-
-def list_node_representer(dumper: Dumper, data):
-    return dumper.represent_list(list(data))
-
-
-yaml.add_representer(YamlList, list_node_representer)
-
-
-class DumpFlowStyle(enum.Enum):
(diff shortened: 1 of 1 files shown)


src/libecalc/presentation/yaml/validation_errors.py
```diff
@@ -36,7 +36,18 @@ def _get_position_in_file_message(mark: Mark) -> str:
 
 
 class ValidationError(Exception):
-    pass
+    def __init__(self, message: Optional[str] = None, location: Optional[str] = None):
+        self._location = location
+        self._message = message
+        super().__init__(message)
```
TeeeJay, line 42: super().__init__ should in general be called first in init method

src/libecalc/presentation/yaml/validation_errors.py
```diff
@@ -141,7 +152,7 @@ def __init__(
         dump_flow_style: Optional[DumpFlowStyle] = None,
     ):
         super().__init__(message)
-        self.message = message
+        self._message = message
```
TeeeJay, line 155: this is not necessary, as the ValidationError will deal with the message parameter
```suggestion
```

src/libecalc/presentation/yaml/validation_errors.py
```diff
@@ -141,7 +152,7 @@ def __init__(
         dump_flow_style: Optional[DumpFlowStyle] = None,
     ):
         super().__init__(message)
```
TeeeJay, line 154:
```suggestion
        super().__init__(message=message)
```

TeeeJay (approved):
see suggested changes

src/libecalc/presentation/yaml/validation_errors.py
frodehk: The issue is that the Exception class does not accept keyword arguments. Hence message must be passed directly in super().__init__()

src/libecalc/presentation/yaml/validation_errors.py
```diff
@@ -36,7 +36,18 @@ def _get_position_in_file_message(mark: Mark) -> str:
 
 
 class ValidationError(Exception):
-    pass
+    def __init__(self, message: Optional[str] = None, location: Optional[str] = None):
+        self._location = location
```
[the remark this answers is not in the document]
frodehk: Updated to type Location

PR #773, opened 2025-01-23
chore: emission intensity and results

frodehk:

Draft: started with the need to reintroduce emission intensity. Realized need for refactoring of results. Hence, some refactoring was tested out.


frodehk:
This is taken care of in other PRs.

PR #775, opened 2025-01-27
fix(cli): include timesteps in json

jsolaas:
Make it easier to use the start times of periods by including it in the
json alongside periods.

Refs:
https://github.com/equinor/ecalc-internal/issues/444


TeeeJay:
Thats it? :p Adding it like this, I think we add it as a temporary solution, we should also consider if end_times needed to be added, without having to rename the current structure? To not have to break again?

Also, since this is a workaround - the json schema will not show change, right? (The static json schema I added earlier) ...

TeeeJay (approved):

jsolaas:
> Thats it? :p Adding it like this, I think we add it as a temporary solution, we should also consider if end_times needed to be added, without having to rename the current structure? To not have to break again?

Adding end times would just require adding it to the dict in `with_timesteps`. What do you consider `break` here? Changing the json for cli? The serialization context added here makes it so that only the cli will see this change.

> Also, since this is a workaround - the json schema will not show change, right? (The static json schema I added earlier) ... `generate_json_schema.py`.

The json schema will not show change. The json-schema we use is for input, but the json-schema for results won't change either.

PR #776, opened 2025-01-27
chore: move expression evaluator to components

frodehk:

Make components able to evaluate expressions. It is a part of the preparation for refactoring of e.g. results.

Sending expression evaluator to all components.


src/libecalc/fixtures/cases/consumer_with_time_slots_models.py
jsolaas, line 1: Does this setup make sense still? Move to builders?

src/libecalc/presentation/yaml/yaml_types/emitters/yaml_venting_emitter.py
```diff
@@ -121,6 +121,13 @@ def id(self) -> str:
         description="The emissions for the emitter of type DIRECT_EMISSION",
     )
 
+    expression_evaluator: Optional[ExpressionEvaluator] = None
+
+    emission_results: Optional[Optional[dict[str, EmissionResult]]] = None
```
jsolaas, line 126: We should do something else here I think, same as the other types where we have a Yaml class and a domain. To avoid dealing with pydantic.

src/libecalc/application/energy/emitter.py
```diff
@@ -21,5 +25,7 @@ def evaluate_emissions(
         self,
         energy_context: ComponentEnergyContext,
         energy_model: EnergyModel,
-        expression_evaluator: ExpressionEvaluator,
     ) -> Optional[dict[str, EmissionResult]]: ...
+
+    def set_expression_evaluator(self, expression_evaluator: ExpressionEvaluator):
+        self.expression_evaluator = expression_evaluator
```
jsolaas, line 31: Do we need a setter?

src/libecalc/application/energy/emitter.py
```diff
@@ -12,6 +14,8 @@ class Emitter(abc.ABC):
     Something that emits something.
     """
 
+    expression_evaluator: Optional[ExpressionEvaluator] = Field(default=None, exclude=True)
```
jsolaas, line 17: It doesn't make sense to me that we require expression_evaluator for an emitter. I think of it as an implementation detail. Hopefully after this the application- and domain-layer don't need to know that expression_evaluator exists at all.

src/libecalc/common/variables.py
```diff
@@ -176,3 +177,7 @@ def get_subset_for_period(self, period: Period) -> ExpressionEvaluator: ...
     def evaluate(
         self, expression: Union[Expression, TemporalModel, dict[Period, Expression]]
     ) -> NDArray[np.float64]: ...
+
+    @classmethod
+    def __get_pydantic_core_schema__(cls, source, handler) -> CoreSchema:
+        return core_schema.any_schema()
```
jsolaas, line 183: Hopefully we can avoid this by changing venting emitter a bit

TeeeJay:
> Looks good. I think we should make some changes to make sure expression_evaluator and pydantic isn't something we have to deal with in domain/application.
> 
> I think this change is a good idea as long as expression evaluator isn't meant to be changed. I don't like the idea of having a setter for expression_evaluator, instead I think we should create a new consumer with another expression_evaluator if that is needed.
> 
> The domain classes will then be a "locked" representation of the yaml model. The yaml model describes a single expression_evaluator, so the need to change it in the domain classes seems unnecessary to me. (Probably useful for testing, but we should just use a factory instead).

I noticed this. Not entirely sure how it works, but is it always the same instance that is being used everywhere? In that case we can create a singleton, that is accessible everywhere, instead of injecting it explicitly. To e.g consider it as a service; "ExpressionEvaluatorService"?

src/libecalc/fixtures/cases/consumer_with_time_slots_models.py
frodehk: I think builders makes sense. Maybe separate PR?

src/libecalc/presentation/yaml/yaml_types/emitters/yaml_venting_emitter.py
frodehk: A new VentingEmitter is implemented in domain. With subclasses DirectVentingEmitter and OilVentingEmitter.

src/libecalc/application/energy/emitter.py
frodehk: It is removed now

src/libecalc/application/energy/emitter.py
frodehk: It is removed now

src/libecalc/common/variables.py
frodehk: It is not needed now as we have updated venting emitter.

src/libecalc/domain/infrastructure/energy_components/electricity_consumer/electricity_consumer.py
```diff
@@ -78,9 +80,10 @@ def get_component_process_type(self) -> ComponentType:
     def get_name(self) -> str:
         return self.name
 
-    def evaluate_energy_usage(
-        self, expression_evaluator: ExpressionEvaluator, context: ComponentEnergyContext
-    ) -> dict[str, EcalcModelResult]:
+    def set_consumer_results(self, consumer_results: dict[str, EcalcModelResult]):
+        self.consumer_results = consumer_results
```
jsolaas, line 84: Remove `set_consumer_result`?

src/libecalc/domain/infrastructure/emitters/venting_emitter.py
```diff
@@ -0,0 +1,167 @@
...
+        self.emitter_type = YamlVentingType.OIL_VOLUME
+
+    def get_emissions(self) -> dict[str, TimeSeriesStreamDayRate]:
+        oil_rates = self.expression_evaluator.evaluate(
+            expression=Expression.setup_from_expression(value=self.volume.rate.value)
+        )
+        if self.volume.rate.type == RateType.CALENDAR_DAY:
+            oil_rates = Rates.to_stream_day(
+                calendar_day_rates=np.asarray(oil_rates), regularity=self.regularity_evaluated
+            ).tolist()
```
jsolaas, line 135: Use get_oil_rates here?

src/libecalc/domain/infrastructure/emitters/venting_emitter.py
```diff
@@ -0,0 +1,167 @@
...
+    @property
+    def regularity_evaluated(self):
+        return self.expression_evaluator.evaluate(expression=TemporalModel(self._regularity)).tolist()
+
+    def evaluate_emissions(
+        self,
+        energy_context: ComponentEnergyContext,
+        energy_model: EnergyModel,
+    ) -> Optional[dict[str, EmissionResult]]:
+        self._regularity = energy_model.get_regularity(self.id)
```
jsolaas, line 56: Pass regularity into __init__ instead? 

If I remember correctly `get_regularity` for EcalcModel was kind of a workaround since we didn't have the flexibility in these classes as we have now. I don't think we have to pass energy_model into evaluate_emissions any more.

frodehk: Yes, agree.

src/libecalc/domain/infrastructure/emitters/venting_emitter.py
frodehk: Done

src/libecalc/domain/infrastructure/energy_components/consumer_system/consumer_system_dto.py
```diff
@@ -48,6 +51,13 @@
 from libecalc.dto.types import ConsumerUserDefinedCategoryType
 from libecalc.dto.utils.validators import validate_temporal_model
 from libecalc.expression import Expression
+from libecalc.presentation.yaml.domain.reference_service import ReferenceService
+from libecalc.presentation.yaml.yaml_types.components.system.priorities import YamlPriorities
+from libecalc.presentation.yaml.yaml_types.components.system.yaml_system_component_conditions import (
+    YamlSystemComponentConditions,
+)
```
jsolaas, line 58: This import is the wrong way, they should probably be redefined here, i.e. domain and yaml should not use the same objects. I get why it is done, but I think we need a mapping from one to the other.

src/libecalc/domain/infrastructure/energy_components/consumer_system/consumer_system_dto.py
```diff
@@ -68,11 +81,17 @@ def __init__(
...
-        self.consumers = consumers
+        self.yaml_component_conditions = yaml_component_conditions
+        self.yaml_consumers = yaml_consumers
+        self.yaml_priorities = yaml_priorities
+        self.expression_evaluator = expression_evaluator
+        self.references = references
+        self.target_period = target_period
         self.fuel = self.validate_fuel_exist(name=self.name, fuel=fuel, consumes=consumes)
         self.component_type = component_type
+        self.consumers = self.calculate_consumers()
```
jsolaas, line 92: This is also a mapping, right? It should probably be done before creating this class.

src/libecalc/presentation/yaml/mappers/component_mapper.py
```diff
@@ -125,12 +165,19 @@ def from_yaml_to_dto(
...
+                    consumes=consumes,
+                    yaml_component_conditions=data.component_conditions,
+                    yaml_priorities=data.stream_conditions_priorities,
+                    yaml_consumers=data.consumers,
+                    expression_evaluator=expression_evaluator,
                     references=self.__references,
                     target_period=self._target_period,
                     fuel=fuel,
+                    component_type=ComponentType.CONSUMER_SYSTEM_V2,
                 )
```
jsolaas, line 181: Moving `to_dto` functionality here (into a separate method), then only passing domain objects to consumer system is probably what we need to do?

src/libecalc/domain/infrastructure/energy_components/consumer_system/consumer_system_dto.py
```diff
@@ -48,6 +51,13 @@
 from libecalc.dto.types import ConsumerUserDefinedCategoryType
 from libecalc.dto.utils.validators import validate_temporal_model
 from libecalc.expression import Expression
+from libecalc.presentation.yaml.domain.reference_service import ReferenceService
+from libecalc.presentation.yaml.yaml_types.components.system.priorities import YamlPriorities
+from libecalc.presentation.yaml.yaml_types.components.system.yaml_system_component_conditions import (
+    YamlSystemComponentConditions,
+)
+
+TYamlConsumer = ForwardRef("TYamlConsumer")
```
jsolaas, line 60: Hopefully we don't need this if we just pass domain classes to consumer system.

src/libecalc/presentation/yaml/yaml_types/components/system/priorities.py
jsolaas, line 1: Are the changes here, and in `type_aliases.py`, `yaml_consumer_system.py` because of circular imports?

src/libecalc/presentation/yaml/yaml_types/components/yaml_generator_set.py
```diff
@@ -57,7 +57,7 @@ class YamlGeneratorSet(YamlBase):
         Annotated[
             Union[
                 YamlElectricityConsumer,
-                YamlConsumerSystem,
+                "YamlConsumerSystem",
```
jsolaas, line 60: Also circular import?

frodehk: Initially the assumption was that it was circular import. But it seems now that it is just some remains after debugging, i.e. it wasn´t circular import after all. The change is reverted.

src/libecalc/presentation/yaml/yaml_types/components/system/priorities.py
frodehk: In yaml_consumer_system.py the main changes are due to circular import. Most is removing functionality to dto. The type_aliases.py is created to ensure consistency, as TYamlConsumer is used both in yaml_consumer_system.py and consumer_system_dto.py. The Way TYamlConsumer is defined, i.e. moving the annotation and discriminator to the place where it is used, instead of the initial definition, is due to mypy. Splitting out priorities is assumed to be related to circular imports. But it is possible that it is not necessary, as ForwardRef seems to be the main factor to solve it. Will check.

src/libecalc/domain/infrastructure/energy_components/consumer_system/consumer_system_dto.py
frodehk: This is fixed. New to_dto methods are implemented to handle it.

src/libecalc/domain/infrastructure/energy_components/consumer_system/consumer_system_dto.py
frodehk: This is fixed by new to_dto method now.

src/libecalc/presentation/yaml/mappers/component_mapper.py
frodehk: Yes, it is done now.

src/libecalc/domain/infrastructure/energy_components/consumer_system/consumer_system_dto.py
frodehk: It is removed now

src/libecalc/presentation/yaml/yaml_types/components/system/priorities.py
frodehk: After recent changes/updates, YamlPriorities can most likely be moved back to yaml_consumer_system.py. However, we have a separate `to_do` method in yaml_priorities.py; @jsolaas is it ok to have YamlPriorities in a separate module after all?

src/libecalc/common/priority_optimizer.py
```diff
@@ -5,7 +5,7 @@
 from functools import reduce
 from typing import Generic, TypeVar
 
-from libecalc.common.priorities import PriorityID
+from libecalc.domain.infrastructure.energy_components.base.component_dto import PriorityID
```
jsolaas, line 8: Redefine or move PriorityID to here, common should not import from domain.

src/libecalc/domain/infrastructure/emitters/venting_emitter.py
```diff
@@ -0,0 +1,160 @@
+from typing import Optional
+
+import numpy as np
+
+from libecalc.application.energy.component_energy_context import ComponentEnergyContext
+from libecalc.application.energy.emitter import Emitter
+from libecalc.application.energy.energy_component import EnergyComponent
+from libecalc.application.energy.energy_model import EnergyModel
```
jsolaas, line 8: Maybe move these into domain in a separate PR?

src/libecalc/domain/infrastructure/emitters/venting_emitter.py
```diff
@@ -0,0 +1,160 @@
...
+from libecalc.common.temporal_model import TemporalModel
+from libecalc.common.time_utils import Period
+from libecalc.common.units import Unit
+from libecalc.common.utils.rates import Rates, RateType, TimeSeriesFloat, TimeSeriesStreamDayRate
+from libecalc.common.variables import ExpressionEvaluator
+from libecalc.core.result.emission import EmissionResult
+from libecalc.dto.types import ConsumerUserDefinedCategoryType
+from libecalc.dto.utils.validators import convert_expression
+from libecalc.expression import Expression
+from libecalc.presentation.yaml.yaml_types.emitters.yaml_venting_emitter import (
```
jsolaas, line 20: We should look into this one also, as we don't want to import from presentation here

src/libecalc/domain/infrastructure/energy_components/consumer_system/consumer_system_dto.py
```diff
@@ -69,10 +73,13 @@ def __init__(
         self.consumes = consumes
         validate_temporal_model(self.regularity)
         self.component_conditions = component_conditions
-        self.stream_conditions_priorities = stream_conditions_priorities
-        self.consumers = consumers
+        self.stream_conditions_priorities = priorities
+        self.expression_evaluator = expression_evaluator
+        self.references = references
```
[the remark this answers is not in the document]
frodehk: Used in individual consumers, but not needed for system. Fixed.

src/libecalc/domain/infrastructure/emitters/venting_emitter.py
frodehk: Fixed: implemented new domain objects for venting emitter related content.

src/libecalc/presentation/yaml/parse_input.py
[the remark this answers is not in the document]
frodehk: Yes, it is fixed now, rebase.

PR #785, opened 2025-02-11
refactor: store results in fuel and electricity components

frodehk:

This PR is part of a larger refactoring of results. It will not implement any functional changes, but is part of a preparation with the objective to allow individual components to be queried directly for their results, simplifying the retrieval process.

- [x] Modify the evaluate_energy_usage method in the component classes to store results within the respective components for: `FuelConsumer`, `ElectricityConsumer`, `GeneratorSet` and `ConsumerSystem` in the domain layer
- [x] Modify the evaluate_emissions method in the component classes to store results within the respective components.

https://github.com/equinor/ecalc-internal/issues/505

### DIFF
--- a/src/libecalc/domain/infrastructure/emitters/venting_emitter.py
+++ b/src/libecalc/domain/infrastructure/emitters/venting_emitter.py
@@ -140,9 +140,6 @@ class VentingEmitter(Emitter, EnergyComponent):
     def is_container(self) -> bool:
         return False
 
-    def get_emission_results(self) -> Optional[dict[str, EmissionResult]]:
-        return self.emission_results
-
 
 class DirectVentingEmitter(VentingEmitter):
     def __init__(self, emissions: list[VentingEmission], **kwargs):
--- a/src/libecalc/domain/infrastructure/emitters/venting_emitter.py
+++ b/src/libecalc/domain/infrastructure/emitters/venting_emitter.py
@@ -72,6 +72,7 @@ class VentingEmitter(Emitter, EnergyComponent):
         self.user_defined_category = user_defined_category
         self.emitter_type = emitter_type
         self._regularity = regularity
+        self.emission_results: Optional[dict[str, EmissionResult]] = None
 
     @property
     def id(self) -> str:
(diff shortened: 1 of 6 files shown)


src/libecalc/domain/infrastructure/emitters/venting_emitter.py
```diff
@@ -96,7 +97,8 @@ def evaluate_emissions(
                 rate=emission_rate,
             )
             venting_emitter_results[emission_name] = emission_result
-        return venting_emitter_results
+        self.emission_results = venting_emitter_results
+        return self.emission_results
```
TeeeJay, line 101: I reckon that returning the full result here is temporary, since other code depends on it, but that it will later be removed? Or is it here for flexibility?

I "always thought" it would be convenient with bare minimum input and output here. E.g:

`VentingEmitter.simulate() -> Emissions` (perhaps this should be in an emission domain?)
`Generator.simulate(power_demand) -> fuel_(gas)_demand`
etc

FWIW :)

src/libecalc/domain/infrastructure/emitters/venting_emitter.py
```diff
@@ -138,6 +140,9 @@ def is_provider(self) -> bool:
     def is_container(self) -> bool:
         return False
 
+    def get_emission_results(self) -> Optional[dict[str, EmissionResult]]:
```
TeeeJay, line 143: I would also want this more "robust", ie a flag indicating whether it has been calculated/simulated or not, and raise an Exception if that is the case.

Must also keep in mind that when we store state in objects, there is always a risk that something is being "mutated", ie. suddenly it is set, when we expect it to be empty, or other way around, reset when we expect it to be there due to "reuse" of objects. Both for us, and possible other users of the library (in different ways...).

That aside, I support the idea of keeping the "big nasty results" and "avoid aggregating results too early", by e.g. keeping it like this, easily accessible.

If we create immutable objects, we avoid this problem, if that is even relevant/possible. Not sure how that would fit in... But I believe that the "results" we have is "just" "things we didnt know before we simulated", that they are just the "x and y and z's in an equition" that we didnt know as a part of the calculation - and that they often fit nicely in as a part of the "objects" state, they complete "the object state". And some parts of the "result" are just "events" to notify e.g. about validity (was the equation solvable?) etc.

Just braindumping without really giving any concrete advice or comment. I do think this is in the right direction though, as you mention - it may/will lead to reason about smaller objects, and only what we want instead of the whole ecalc model all the time ...

jsolaas (approved):

PR #788, opened 2025-02-14
feat: re-introduce co2 intensity in results

frodehk:
equinor/ecalc-internal#303

co2 intensities should be a part of the eCalc results again. 

- [x] Re-implement co2 intensities: cumulative intensities and yearly intensities 
- [x] Ensure correct resampling of intensities: Use resampled emissions and hydrocarbon export
- [x] Intensities are reported both in csv and json results
- [x] Create tests
- [x] Update snapshots

https://github.com/equinor/ecalc-internal/issues/303

### DIFF
--- a/src/ecalc_cli/commands/run.py
+++ b/src/ecalc_cli/commands/run.py
@@ -139,14 +139,15 @@ def run(
         consumer_results=consumer_results,
         variables_map=model.variables,
         emission_results=emission_results,
-        output_frequency=frequency,
     )
 
     run_info.end = datetime.now()
 
     output_prefix: Path = output_folder / name_prefix
 
-    results_dto = get_asset_result(results_core)
+    include_emission_intensity = frequency == libecalc.common.time_utils.Frequency.YEAR
+
+    results_dto = get_asset_result(results_core, include_emission_intensity)
 
     if (
         frequency != libecalc.common.time_utils.Frequency.NONE
--- a/examples/simple_yaml_model.ipynb
+++ b/examples/simple_yaml_model.ipynb
@@ -89,7 +89,6 @@
     "    consumer_results=consumer_results,\n",
     "    variables_map=yaml_model.variables,\n",
     "    emission_results=emission_results,\n",
-    "    output_frequency=yaml_model._output_frequency,\n",
     ")"
    ]
   },
--- a/src/libecalc/application/graph_result.py
+++ b/src/libecalc/application/graph_result.py
@@ -1,6 +1,5 @@
 from pydantic import BaseModel
 
-from ecalc_cli.types import Frequency
 from libecalc.common.variables import VariablesMap
 from libecalc.core.result import ComponentResult, EcalcModelResult
 from libecalc.core.result.emission import EmissionResult
@@ -20,13 +19,11 @@ class GraphResult:
         consumer_results: dict[str, EcalcModelResult],
         emission_results: dict[str, dict[str, EmissionResult]],
         variables_map: VariablesMap,
-        output_frequency: Frequency,
     ):
         self.graph = graph
         self.consumer_results = consumer_results
         self.emission_results = emission_results
         self.variables_map = variables_map
-        self.output_frequency = output_frequency
 
     def get_subgraph(self, node_id: str) -> "GraphResult":
         subgraph = self.graph.get_node(node_id).get_graph()
@@ -44,7 +41,6 @@ class GraphResult:
                 for component_id, emissions in self.emission_results.items()
                 if component_id in subgraph
             },
-            output_frequency=self.output_frequency,
         )
 
     def get_energy_result(self, component_id: str) -> ComponentResult:
--- a/src/libecalc/presentation/json_result/mapper.py
+++ b/src/libecalc/presentation/json_result/mapper.py
@@ -15,7 +15,7 @@ from libecalc.common.decorators.feature_flags import Feature
 from libecalc.common.errors.exceptions import ProgrammingError
 from libecalc.common.math.numbers import Numbers
 from libecalc.common.temporal_model import TemporalModel
-from libecalc.common.time_utils import Frequency, Period, Periods
+from libecalc.common.time_utils import Period, Periods
 from libecalc.common.units import Unit
 from libecalc.common.utils.rates import (
     RateType,
@@ -396,7 +396,9 @@ def _evaluate_installations(
     return installation_results
 
 
-def get_asset_result(graph_result: GraphResult) -> libecalc.presentation.json_result.result.results.EcalcModelResult:
+def get_asset_result(
+    graph_result: GraphResult, include_emission_intensity: bool
+) -> libecalc.presentation.json_result.result.results.EcalcModelResult:
     asset_id = graph_result.graph.root
     asset = graph_result.graph.get_node(asset_id)
 
@@ -1463,7 +1465,7 @@ def get_asset_result(graph_result: GraphResult) -> libecalc.presentation.json_re
             hydrocarbon_export_rate=asset_hydrocarbon_export_rate_core,
             emissions=asset_aggregated_emissions,
         )
-        if installation_results and graph_result.output_frequency == Frequency.YEAR
+        if installation_results and include_emission_intensity
         else [],
     )
 
--- a/tests/ecalc_cli/test_app.py
+++ b/tests/ecalc_cli/test_app.py
@@ -11,6 +11,7 @@ import yaml
 from typer.testing import CliRunner
 
 from ecalc_cli import main
+from libecalc.common.time_utils import Frequency
 from libecalc.common.run_info import RunInfo
 from libecalc.dto.utils.validators import COMPONENT_NAME_ALLOWED_CHARS
 from libecalc.presentation.yaml.model_validation_exception import ModelValidationException
--- a/tests/libecalc/dto/results/model/test_requested_pressures_compressors.py
+++ b/tests/libecalc/dto/results/model/test_requested_pressures_compressors.py
@@ -4,7 +4,7 @@ import pytest
 
 from libecalc.application.energy_calculator import EnergyCalculator
 from libecalc.application.graph_result import EcalcModelResult, GraphResult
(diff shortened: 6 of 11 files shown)


src/libecalc/common/utils/emission_intensity.py
```diff
@@ -0,0 +1,49 @@
+from libecalc.common.units import Unit
+from libecalc.common.utils.rates import (
+    TimeSeriesIntensity,
+    TimeSeriesVolumesCumulative,
+)
+
+
+class EmissionIntensity:
```
TeeeJay, line 8: I would like to have this in the "emission domain" ...I think :)

src/libecalc/common/utils/emission_intensity.py
```diff
@@ -0,0 +1,62 @@
...
+        emission_cumulative: TimeSeriesVolumesCumulative,
+        hydrocarbon_export_cumulative: TimeSeriesVolumesCumulative,
+        unit: Unit = Unit.KG_SM3,
+    ):
+        self.emission_cumulative = emission_cumulative
+        self.hydrocarbon_export_cumulative = hydrocarbon_export_cumulative
+        self.unit = unit
+        self.periods = emission_cumulative.periods
+
+    def calculate_intensity_periods(self):
```
TeeeJay, line 33:
```suggestion
    def calculate_intensity_periods(self) -> TimeSeriesIntensity:
```

src/libecalc/common/utils/emission_intensity.py
```diff
@@ -0,0 +1,62 @@
...
+        periods (list): The periods over which the data is calculated. Used only when calculating intensity for each period within the data range.
+    """
+
+    def __init__(
+        self,
+        emission_cumulative: TimeSeriesVolumesCumulative,
+        hydrocarbon_export_cumulative: TimeSeriesVolumesCumulative,
+        unit: Unit = Unit.KG_SM3,
+    ):
+        self.emission_cumulative = emission_cumulative
```
TeeeJay, line 28: Should be a check that the 2 timeseries are of same length?

src/libecalc/common/utils/emission_intensity.py
```diff
@@ -0,0 +1,62 @@
...
+        hydrocarbon_export_cumulative (TimeSeriesVolumesCumulative): The cumulative hydrocarbon export data.
+        unit (Unit): The unit of measurement for the intensity.
+        periods (list): The periods over which the data is calculated. Used only when calculating intensity for each period within the data range.
+    """
+
+    def __init__(
+        self,
+        emission_cumulative: TimeSeriesVolumesCumulative,
+        hydrocarbon_export_cumulative: TimeSeriesVolumesCumulative,
+        unit: Unit = Unit.KG_SM3,
```
TeeeJay, line 26: should unit be set here? It is the requested unit, and should therefore be verified that it is what we end up with?

src/libecalc/common/utils/rates.py
```diff
@@ -665,6 +665,66 @@ def to_rate(self, regularity: Optional[list[float]] = None) -> TimeSeriesRate:
...
+
+        if self.unit == Unit.KILO and other.unit == Unit.STANDARD_CUBIC_METER:
+            unit = Unit.KG_SM3
+        else:
+            raise ProgrammingError(
+                f"Unable to divide unit '{self.unit}' by unit '{other.unit}'. Please add unit conversion."
+            )
+        return TimeSeriesCalendarDayRate(
+            periods=self.periods,
+            values=np.divide(
```
TeeeJay, line 680: Here we could also check length, to get a nice error instead of numpy error, if they are of uneven lengths?

TeeeJay (approved):
LGTM, I guess you mostly copied and refactored the previous implementation? I do not have complete overview if calculations are done in the right place or done correctly...sb else to approve that :)

src/libecalc/common/utils/rates.py
frodehk: Done now

src/libecalc/common/utils/emission_intensity.py
frodehk: Done now

src/libecalc/common/utils/emission_intensity.py
frodehk: Done

src/libecalc/common/utils/emission_intensity.py
frodehk: I see that we now have domain.energy with emitter, energy_components, energy_model. I guess we should have a separate domain.emission and move emitter and emission_intensity here?

src/libecalc/common/utils/emission_intensity.py
frodehk: Removed unit and added check for compatible units in this context

src/libecalc/common/utils/emission_intensity.py
TeeeJay: I would at least like to discuss. Perhaps there are reasons and arguments not to, but I thought maybe it would make sense...but it may easily not as well (to tightly coupled with fuel f ex) ...

frodehk: I think I agree that it should be in domain. Under a separate emission level. We can discuss. @jsolaas, what do you think?

frodehk:
> LGTM, I guess you mostly copied and refactored the previous implementation? I do not have complete overview if calculations are done in the right place or done correctly...sb else to approve that :)

Much of it is previous implementation. However, calculations of periodical intensities are different. We are now using resampled emissions and hydrocarbon exports to recalculate intensity for resampled periods.

TeeeJay:
> > LGTM, I guess you mostly copied and refactored the previous implementation? I do not have complete overview if calculations are done in the right place or done correctly...sb else to approve that :)
> 
> Much of it is previous implementation. However, calculations of periodical intensities are different. We are now using resampled emissions and hydrocarbon exports to recalculate intensity for resampled periods.

Ok, @olelod to verify/review then :)

src/libecalc/presentation/json_result/result/emission.py
```diff
@@ -62,3 +63,11 @@ def from_emission_core_result(cls, emission_result: EmissionCoreResult, regulari
             periods=emission_result.periods,
             rate=TimeSeriesRate.from_timeseries_stream_day_rate(emission_result.rate, regularity),
         )
+
+
```
frodehk, line 67: Should we be more precise in the naming? E.g. intensity_cumulative and intensity_by_period?

frodehk: If I understand correctly, you will move the "result" part to domain, as a part of the EmissionIntensity class (domain.emission.emission_intensity)?

TeeeJay: I see now that I confuse you :) I do not have the overview, but I think the current `EmissionIntensity` is perhaps a `EmissionIntensityService` , or the `Emission`class has a `Emission.calculate_intensity(hc_export: float)` **OR** either the Asset or the Installation has a `installation.calculate_emission_intensity(emission_type)->EmissionIntensity` or sth along those lines :) Just trying to think and trigger about the "where does it come from, what does it belong to etc" in a domain perspective :)

olelod:
> > > LGTM, I guess you mostly copied and refactored the previous implementation? I do not have complete overview if calculations are done in the right place or done correctly...sb else to approve that :)
> > 
> > 
> > Much of it is previous implementation. However, calculations of periodical intensities are different. We are now using resampled emissions and hydrocarbon exports to recalculate intensity for resampled periods.
> 
> Ok, @olelod to verify/review then :)

Before we had this intensity by 'yearly buckets', which was a bit weird. Now, you get the intensity for each year if you run with yearly frequency. It makes more sense to get period intensities related to the periods you are interested in the results on. Later, when we get around to resampling results, you can monthly/yearly intensities by post processing the results.

olelod (approved):

src/libecalc/domain/emission/emission_intensity.py
```diff
@@ -0,0 +1,75 @@
...
+        self.periods = emission_cumulative.periods
+
+    def calculate_for_periods(self) -> TimeSeriesIntensity:
+        """
+        Calculate the emission intensity for each period over the entire data range.
+        """
+        emission_volumes = self.emission_cumulative.to_volumes()
+        hydrocarbon_export_volumes = self.hydrocarbon_export_cumulative.to_volumes()
+
+        intensity = emission_volumes / hydrocarbon_export_volumes
```
olelod, line 53: We may run into trouble here if the hydrocarbon_export_volumes are zero.

src/libecalc/domain/emission/emission_intensity.py
```diff
@@ -0,0 +1,75 @@
...
+            unit=self.unit,
+            emissions=emission_volumes,
+            hc_export=hydrocarbon_export_volumes,
+        )
+
+    def calculate_cumulative(self) -> TimeSeriesIntensity:
+        """
+        Calculate the cumulative emission intensity over the entire data range.
+        """
+        intensity = self.emission_cumulative / self.hydrocarbon_export_cumulative
```
olelod, line 67: And here

frodehk: I think it will be null or None.

src/libecalc/domain/emission/emission_intensity.py
frodehk: I guess it will be null/None, that 0 export volumes for a small period will be a gap in the graph in web. But should probably be tested.

src/ecalc_cli/commands/run.py
```diff
@@ -147,7 +148,9 @@ def run(
 
     results_dto = get_asset_result(results_core)
 
-    if frequency != Frequency.NONE:
+    if (
+        frequency != libecalc.common.time_utils.Frequency.NONE
+    ):  # Not sure why this had to be changed from Frequency.NONE to libecalc.common.time_utils.Frequency.NONE
```
TeeeJay, line 153: We have at least 2 frequency enums ... probably some mess.

PR #790, opened 2025-02-17
perf(libecalc): Increase performance of datetime-parsing

Aleksander-Karlsson:

### DIFF
--- a/src/libecalc/presentation/yaml/mappers/variables_mapper/time_series_collection_mapper.py
+++ b/src/libecalc/presentation/yaml/mappers/variables_mapper/time_series_collection_mapper.py
@@ -2,37 +2,77 @@ import re
 from datetime import datetime
 from typing import Union
 
-import pandas
+import pandas as pd
 
 from libecalc.common.errors.exceptions import InvalidResourceException
 from libecalc.presentation.yaml.resource import Resource
+from libecalc.presentation.yaml.validation_errors import ValidationError
 from libecalc.presentation.yaml.yaml_keywords import EcalcYamlKeywords
 
 
-def _parse_date(date_input: Union[int, str]) -> datetime:
+def parse_time_vector(date_input: list[int | str]) -> list[datetime]:
+    """Parse entire timeseries in a single format.
+
+    Args:
+        date_input: Dates in unknown format.
+
+    Returns:
+        Consistent dates.
+
+    Raises:
+        ValidationError:
+            If dates do not match any of the given patterns.
+            If dates are in an inconsistent format.
     """
-    Parse timeseries input:
+    date_patterns = {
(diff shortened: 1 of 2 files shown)


src/libecalc/presentation/yaml/mappers/variables_mapper/time_series_collection_mapper.py
```diff
@@ -1,38 +1,76 @@
...
+        Consistent dates.
+
+    Raises:
+        ValidationError:
+            If dates do not match any of the given patterns.
+            If dates are in an inconsistent format.
+    """
+    date_patterns = {
+        # Only year supplied (YYYY e.g. 1996).
+        "YEAR_ONLY": r"\d{4}",
```
TeeeJay, line 27: We have traditionally in eCalc had support for 3 different date formats, so in theory other formats has not been supported. See `run.py` and `datetime/utils.py` under `date_format_option`. This was set on CLI (cant remember if this was in and/or output actually ...`

We "always" wanted to only accept ISO8601 YYYY-MM-DD HH:MM:SS but we had to accept DD.MM.YYYY as that was the most common format in csv files.

We do have several places where we parse data from csv, or at least we had. Might be a problem as well. Jostein should know...

I think we open up for more challenges now to support more dateformats than we had, but if we at some point always preprocess .csv data into json structures f ex, than it might be handy to support many raw import formats ... Since we now use it directly, I think we should stick to just a few ...

I would like to hear what others think

ketwik: Agree to limit the support of the date formats to only the 3 formats already supported and not open for more variations at this point in time as Gary Neville would say.

ketwik: Before releasing this change I think we also need to add validation on saving resource files that enforces same format in the same file. 
At the same time check if we have resource files with multiple date formats and if so migrate these.

src/libecalc/presentation/yaml/mappers/variables_mapper/time_series_collection_mapper.py
```diff
@@ -1,49 +0,0 @@
-import re
```
jsolaas, line 1: Is this an empty file now?

jsolaas (approved):

TeeeJay (approved):

src/libecalc/presentation/yaml/mappers/variables_mapper/time_series_collection_mapper.py
Aleksander-Karlsson: Good catch, thought I'd deleted it but I'll fix!

PR #793, opened 2025-02-19
feat: manually control java process for neqsim

jsolaas:
When starting a new java-process we also need to clear the NeqsimFluid cache, since the py4j/java-connection is stored within the cached objects (JavaObject from py4j).

### DIFF
--- a/src/ecalc_neqsim_wrapper/java_service.py
+++ b/src/ecalc_neqsim_wrapper/java_service.py
@@ -1,27 +1,76 @@
+import logging
 import os
 from os import path
+from typing import Optional
 
 from py4j.java_gateway import JavaGateway
 
-gateway = None
-local_os_name = os.name
-colon = ":"
-if local_os_name == "nt":
-    colon = ";"
+_logger = logging.getLogger(__name__)
 
 
-def create_classpath(jars):
+# Java process started explicitly, should only be used 'on-demand', not on import
+_neqsim_service: Optional["NeqsimService"] = None
+
+
+_local_os_name = os.name
+_colon = ":"
+if _local_os_name == "nt":
+    _colon = ";"
+
+
+def _create_classpath(jars):
     """Create path to NeqSim .jar file"""
     resources_dir = path.dirname(__file__) + "/lib"
-    return colon.join([path.join(resources_dir, jar) for jar in jars])
+    return _colon.join([path.join(resources_dir, jar) for jar in jars])
 
 
-def start_server():
-    """Start JVM for NeqSim Wrapper"""
-    global gateway
+def _start_server(maximum_memory: str = "4G") -> JavaGateway:
+    """
+    Start JVM for NeqSim Wrapper
+    Returns: (int, Popen) port, process
+
+    """
     jars = ["NeqSim.jar"]
-    classpath = create_classpath(jars)
-    import logging
+    classpath = _create_classpath(jars)
 
     logging.getLogger("py4j").setLevel(logging.ERROR)
-    return JavaGateway.launch_gateway(classpath=classpath, die_on_exit=True, javaopts=["-Xmx4G"])
+    return JavaGateway.launch_gateway(classpath=classpath, die_on_exit=False, javaopts=[f"-Xmx{maximum_memory}"])
+
+
+class NeqsimService:
+    def __init__(self, maximum_memory: str = "4G"):
+        global _neqsim_service
+        self._gateway = _start_server(maximum_memory=maximum_memory)
+        _logger.info(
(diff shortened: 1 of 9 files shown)


src/ecalc_neqsim_wrapper/java_service.py
```diff
@@ -1,27 +1,80 @@
...
+
+class NeqsimService:
+    def __init__(self, gateway: JavaGateway):
+        self.gateway = gateway
+
+    def get_neqsim_module(self):
+        return self.gateway.jvm.neqsim
+
+
+@contextmanager
```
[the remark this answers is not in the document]
jsolaas: I dunno, I don't think any of them are recommended or both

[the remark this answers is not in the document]
jsolaas: I don't think one is preferred over the other is what I'm trying to say, depends I guess

src/ecalc_neqsim_wrapper/thermo.py
```diff
@@ -27,8 +25,13 @@
 
 NEQSIM_MIXING_RULE = 2
 
-ThermodynamicSystem = neqsim.thermo.system.SystemEos
-ThermodynamicOperations = neqsim.thermodynamicoperations.ThermodynamicOperations
+
+class ThermodynamicSystem(Protocol):
```
TeeeJay, line 29: planning to add anything to these 2 protocols? I dont understand the need for them atm

src/ecalc_neqsim_wrapper/thermo.py
```diff
@@ -152,31 +154,60 @@ def create_thermo_system(
...
+            eos_model_type=eos_model,
             temperature_kelvin=temperature_kelvin,
             pressure_bara=pressure_bara,
             mixing_rule=mixing_rule,
         )
 
-        return NeqsimFluid(thermodynamic_system=thermodynamic_system, use_gerg=use_gerg)
+        # Since we are caching the java objects, they will contain the connection info to the java process (py4j).
+        # That connection info might be outdated; clear the cache if that is the case.
+        if thermodynamic_system._gateway_client.port != get_neqsim()._gateway_client.port:
```
TeeeJay, line 168: is this a current problem? not possible to reset in the beginning of the context(manager)? or is it wrt. possible future where there are > 1 thread or async process running? In that case, could/should the port be a part of the parameter list? Or, does it matter - if 2 processes have the same parameters, wouldnt the LRU cache be the same and ok to reuse?

src/ecalc_neqsim_wrapper/java_service.py
```diff
@@ -1,27 +1,80 @@
...
 
-gateway = None
-local_os_name = os.name
-colon = ":"
-if local_os_name == "nt":
-    colon = ";"
+_logger = logging.getLogger(__name__)
 
+# Java process started explicitly, should only be used 'on-demand', not on import
+_java_gateway: ContextVar[JavaGateway] = ContextVar("The java gateway")
```
TeeeJay, line 13: yes, probably not needed atm because of 1 process/worker/thread, and most likely it will always be like that? It would just add extra complexity/lack of understanding in keeping it? Unless it adds extra semantics by indicating that it unique to a context (manager), even though it is not needded since there is always one (even in tests)?

src/ecalc_neqsim_wrapper/thermo.py
jsolaas: I just had to replace the usage of neqsim (java connection) in python types. What should we do? Create protocols showing the functions we use in the java objects? Encapsulate the java objects? Not sure what's best, so I added empty protocols instead of removing the types.

src/ecalc_neqsim_wrapper/thermo.py
jsolaas: Current problem. Tests fail since we start a new java-process with random port for each test, while the cache is kept across tests in the same process.

src/ecalc_neqsim_wrapper/thermo.py
TeeeJay: ok, probably "facade" pattern then, adding methods the way we want us to use them, and then for the neqsim-implementation, call the java-methods...

src/ecalc_neqsim_wrapper/thermo.py
TeeeJay: hmm.  I thought the parameters defined the cache, and if they are the same, then the cache could be reused, even "across different tests". Or is there (I guess so) other parameters that also matters?

TeeeJay (approved):

PR #794, opened 2025-02-19
chore: remove create-jira-release

jsolaas:
Also add concurrency to on-push-any-branch. This will cancel workflows
when we push a new change to branches.


.github/workflows/on-push-any-branch.yml
```diff
@@ -8,6 +8,10 @@ on:
     tags-ignore:
       - '**'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
TeeeJay, line 12: ref is branch name then?

jsolaas: Yes, `refs/heads/branch-name`

PR #804, opened 2025-02-25
refactor: move models to domain

frodehk:

Part of libecalc refactoring: Move models from core and dto to domain. 

Move models to domain.


src/libecalc/domain/process/core/results/compressor.py
```diff
@@ -130,7 +130,7 @@ def create_empty(cls, number_of_periods: int) -> CompressorStageResult:
             asv_recirculation_loss_mw=nans,
             fluid_composition={},
             is_valid=[True] * number_of_periods,
-            chart_area_flags=[libecalc.core.models.chart.chart_area_flag.ChartAreaFlag.NOT_CALCULATED]
+            chart_area_flags=[libecalc.domain.process.core.chart.chart_area_flag.ChartAreaFlag.NOT_CALCULATED]
             * number_of_periods,
```
olelod, line 134: Is there a reason why we don't do

`from libecalc.domain.process.core.chart.chart_area_flag import ChartAreaFlag`

and

`chart_area_flags=[ChartAreaFlag.NOT_CALCULATED]`

here?

TeeeJay (approved):

olelod (approved):

src/libecalc/domain/process/core/results/compressor.py
frodehk: No reason. Fixed now.

olelod: It was more a question out of curiosity than you needing to fix it, but ok.

PR #809, opened 2025-02-25
fix: remove v2 consumers

jsolaas:
Remove unfinished v2 consumers


frodehk (approved):
Nice! As you see, recent merged will cause some conflicts, i.e. moving pump models and removing imports in original dto folder.

PR #812, opened 2025-02-27
fix: error message when variables are unsolvable

jsolaas:
Catch the error and convert it to an expected error.

Further improvement could be to evaluate variables when needed, but it
would still be nice if we could make sure we _can_ solve in
validating.

### DIFF
--- a/src/libecalc/presentation/yaml/mappers/variables_mapper/variables_mapper.py
+++ b/src/libecalc/presentation/yaml/mappers/variables_mapper/variables_mapper.py
@@ -12,6 +12,11 @@ from libecalc.presentation.yaml.yaml_types.yaml_variable import (
 )
 
 
+class InvalidVariablesException(Exception):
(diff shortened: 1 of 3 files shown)


src/libecalc/presentation/yaml/mappers/variables_mapper/variables_mapper.py
```diff
@@ -12,6 +12,11 @@
 )
 
 
+class InvalidVariablesException(Exception):
```
TeeeJay, line 15: Not EcalcException or ValidationException indicating that it is "our" exception, and that we have handled it (safely)?

TeeeJay: nver mind, saw it is being caught intrernally and sent as validationerror after that. all good :)

TeeeJay (approved):

PR #813, opened 2025-02-28
fix: update mix_neqsim_streams calculation

kjbrak:
Updated mix_neqsim_streams to correctly calculate component mole fractions based on mass flow rates and molar masses. Added test to verify the mixing calculations against manually calculated expected values.

The `mix_neqsim_streams` function in the NeqSim wrapper had an incorrect implementation of stream mixing calculations. 

- Updates the `mix_neqsim_streams` function to correctly calculate component mole fractions.
- Add a test case to unit test the function.
- Change snapshot for all test cases v3
- Change desired density in a fluid mix test
- Minor change in pytest config to get rid of an import error that appeared.
- Minor change in docker compose file to get rid of warning about architecture "mismatch"

refs:
https://github.com/equinor/ecalc-internal/issues/545


kjbrak:
Get two failing tests. I will check..

pyproject.toml
```diff
@@ -74,9 +74,8 @@ build-backend = "poetry.core.masonry.api"
 
 # plugins, plugin config etc
 [tool.pytest.ini_options]
-testpaths = [
-    "tests",
-]
+testpaths = ["tests"]
+pythonpath = ["."]
```
TeeeJay, line 78: Nødvendig? Hva var dette før du satte det eksplisitt? Hvorfor måtte det endres? Fint med inline kommentar mtp forståelse :)

src/ecalc_neqsim_wrapper/thermo.py
```diff
@@ -446,17 +441,20 @@ def mix_neqsim_streams(
     mol_per_hour_1 = mass_rate_stream_1 / stream_1.molar_mass
     mol_per_hour_2 = mass_rate_stream_2 / stream_2.molar_mass
 
-    fraction_1 = mol_per_hour_1 / (mol_per_hour_1 + mol_per_hour_2)
-    fraction_2 = mol_per_hour_2 / (mol_per_hour_1 + mol_per_hour_2)
+    component_moles: dict[str, float] = defaultdict(float)
```
TeeeJay, line 444:
```suggestion
    fluid_component_moles: dict[str, float] = defaultdict(float)
```

gir seg vel av kontekst, men er det "fluid component" det heter?

test/docker-compose.yml
```diff
@@ -8,6 +8,7 @@ services:
     volumes:
       - ../:/project/libecalc
     command: ["poetry", "run", "pytest", "-m", "dockersnapshot", "--snapshot-update"]
+    platform: linux/amd64
```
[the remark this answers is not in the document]
kjbrak: Det var ingenting i docker compose filen, men vi har platform i dockerfilen som gav en warning om noe platform mismatch, den warningen forsvant da jeg inkluderte denne linjen i docker compose

TeeeJay (approved):
LGTM, Ole Petter får sjekke når han er tilbake :)

test/docker-compose.yml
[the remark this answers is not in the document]
kjbrak: Ja det var derfor vi la til den der i dockerfilen tror jeg

kjbrak:
I see that there is a mixing method on the GERG2008FluidProperties class also. I will update this as well.

kjbrak:
I made some improvements and simplifications in the method and the test. Improved naming and added an important todo comment about accessing molar mass without having to create a new fluid (including a flash calculation).

TeeeJay:
@jsolaas and @TeeeJay realized that we appearantly do not use Neqsim when mixing fluids, or do we? Could this have been done in eCalc only wo. using neqsim?

TeeeJay:
> @jsolaas and @TeeeJay realized that we appearantly do not use Neqsim when mixing fluids, or do we? Could this have been done in eCalc only wo. using neqsim?

Having said that, I think we can include this and fix that later though.

src/ecalc_neqsim_wrapper/thermo.py
```diff
@@ -416,6 +414,23 @@ def _get_enthalpy_joule_for_GERG2008_joule_per_kg(enthalpy: float, thermodynamic
...
+        composition: A FluidComposition object containing mole fractions of components
+
+    Returns:
+        float: The molar mass of the mixture in kg/mol
+    """
+    comp_dict = composition.model_dump()
+    molar_mass = 0.0
+    for component, mole_fraction in comp_dict.items():
+        if mole_fraction > 0:  # Skip zero components
+            molar_mass += mole_fraction * getattr(ComponentMolecularWeight, component.upper())
```
olelod, line 430: Would not this require the composition given for the fluid model in the YAML file to always be fractions which sum to 1? I think it is common to also give the composition as percentages that sums to 100. Do we have any kind of normalisation somewhere? Or should we have it - to make sure that it is mole fractions when we do calculations in core.

kjbrak: Thats a good point. If its possible to give composition as percentage, then we should handle that here. I will take a look.

PR #814, opened 2025-02-28
chore: remove pydantic from domain process dto

frodehk:

Part of libecalc refactoring and code improvements.

- Remove pydantic from libecalc.domain.process.dto (previously named models in main dto folder)
- Update affected classes and methods
- Update tests and fixtures

### DIFF
--- a/src/libecalc/domain/process/dto/chart.py
+++ b/src/libecalc/domain/process/dto/chart.py
@@ -1,18 +1,56 @@
 from typing import Literal
 
-from pydantic import Field
-
 from libecalc.common.chart_type import ChartType
-from libecalc.dto.base import EcalcBaseModel
 
 
-class GenericChartFromDesignPoint(EcalcBaseModel):
+class GenericChartFromDesignPoint:
     typ: Literal[ChartType.GENERIC_FROM_DESIGN_POINT] = ChartType.GENERIC_FROM_DESIGN_POINT
-    polytropic_efficiency_fraction: float = Field(..., gt=0, le=1)
-    design_rate_actual_m3_per_hour: float = Field(..., ge=0)
-    design_polytropic_head_J_per_kg: float = Field(..., ge=0)
+
+    def __init__(
+        self,
+        polytropic_efficiency_fraction: float,
+        design_rate_actual_m3_per_hour: float,
+        design_polytropic_head_J_per_kg: float,
+    ):
+        self.polytropic_efficiency_fraction = polytropic_efficiency_fraction
+        self.design_rate_actual_m3_per_hour = design_rate_actual_m3_per_hour
+        self.design_polytropic_head_J_per_kg = design_polytropic_head_J_per_kg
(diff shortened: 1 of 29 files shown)


src/libecalc/domain/process/dto/chart.py
```diff
@@ -1,18 +1,106 @@
...
+    ):
+        self.polytropic_efficiency_fraction = polytropic_efficiency_fraction
+        self.design_rate_actual_m3_per_hour = design_rate_actual_m3_per_hour
+        self.design_polytropic_head_J_per_kg = design_polytropic_head_J_per_kg
+        self._validate_polytropic_efficiency_fraction()
+        self._validate_design_rate_actual_m3_per_hour()
+        self._validate_design_polytropic_head_J_per_kg()
+
+    def _validate_polytropic_efficiency_fraction(self):
+        if not isinstance(self.polytropic_efficiency_fraction, int | float):
```
TeeeJay, line 25: in general we should be able to avoid non-numbers to be added here, otherwise we would have to have that all over the place. wrt python it could of course happen, but I think we can/should skip that check, otherwise we need that everywhere.

src/libecalc/domain/process/dto/chart.py
```diff
@@ -1,18 +1,106 @@
...
+            msg = "polytropic_efficiency_fraction must be a number"
+
+            raise ProcessChartValueValidationException(
+                errors=[
+                    ModelValidationError(name=self.typ.value, location=Location([self.typ.value]), message=str(msg))
+                ],
+            )
+
+        if not (0 < self.polytropic_efficiency_fraction <= 1):
+            msg = "polytropic_efficiency_fraction must be greater than 0 and less than or equal to 1"
```
TeeeJay, line 35: we should include the nr that was 0 < nr < 1 here.

TeeeJay: same for other exceptions

src/libecalc/domain/process/dto/compressor/sampled.py
```diff
@@ -29,19 +50,52 @@ def validate_equal_list_lengths(self):
...
+            "energy_usage_values",
+            "rate_values",
+            "suction_pressure_values",
+            "discharge_pressure_values",
+            "power_interpolation_values",
+        ):
+            variable = getattr(self, variable_name)
+            if variable is not None:
+                if any(value < 0 for value in variable):
+                    msg = f"All values in {variable_name} must be greater than or equal to 0"
```
TeeeJay, line 91: incl the err nr

TeeeJay (approved):
nicely scoped work I think. Make sure to add enough information in exceptions for us/user to get enough context :)

src/libecalc/domain/process/dto/chart.py
frodehk: Ref discussions: Let us keep it until we are sure that it is properly validated earlier, e.g. when reading the actual resource. Agree?

PR #818, opened 2025-03-10
Chore/pyupgrade pre commit

Aleksander-Karlsson:


src/ecalc_cli/commands/show.py
```diff
@@ -35,7 +34,7 @@ def show_yaml(
 
 @app.command("schema")
 def show_schema(
-    output_file: Optional[Path] = typer.Option(
+    output_file: Path | None = typer.Option(
```
TeeeJay, line 37: I always thought Optional[Path] was better than Path | None, more readable, more Pythonic, but seems that the latter one is preferred. Oh well :)

TeeeJay (approved):

src/ecalc_cli/commands/show.py
Aleksander-Karlsson: I think it has a tendency to look cleaner when there are more than one Union/Optional in the same line as it tends to get less crowded with the 'new' annotation 🤩 
But yeah, its just a matter of taste, though Python seem to be moving in this direction :)

PR #819, opened 2025-03-11
Create stream and fluid domain with thermodynamic adapter

kjbrak:

To create centralized domain entity around fluid and stream and to create a thermodynamic adapter for future modifications of thermodynamic engine or simplifications and optimizations related to thermodynamic calculations.

- Centralizes domain entities for fluid and stream processing by defining dedicated classes for process conditions, fluid, and stream.
- Implements the NeqSim thermodynamic adapter to encapsulate interactions with the NeqSim Java service, paving the way for future modifications/optimizations or changing thermo engine more easily.
- Introduces a simplified stream mixing strategy with molar balance with simplifications similar to the ones used earlier in the neqsim fluid mix method, but a little enhanced by assuming outlet set to lowest inlet pressure and a simplified mass weighted temperature instead of setting equal to one of the streams (not requiring thermodynamic equilibrium calculations)
- Enhances fluid composition handling by adding a normalization method to `FluidComposition`, ensuring mole fractions always sum to 1 and making the logic agnostic to changes in component names or count.
- Testing across the new domain entities, mixing strategies, and thermodynamic adapter to ensure robustness and maintainability.

refs.:
?

### DIFF
--- a/src/libecalc/domain/process/core/stream/conditions.py
+++ b/src/libecalc/domain/process/core/stream/conditions.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 
-from libecalc.common.units import UnitConstants
+from libecalc.common.units import Unit, UnitConstants
 
 
 @dataclass(frozen=True)
@@ -28,22 +28,22 @@ class ProcessConditions:
     @property
     def temperature_celsius(self) -> float:
         """Get temperature in Celsius."""
-        return self.temperature - 273.15
+        return Unit.KELVIN.to(Unit.CELSIUS)(self.temperature)
 
     @property
     def pressure_pascal(self) -> float:
         """Get pressure in Pascal."""
-        return self.pressure * 1e5
+        return Unit.BARA.to(Unit.PASCAL)(self.pressure)
 
     @property
     def pressure_atm(self) -> float:
         """Get pressure in atmospheres."""
-        return self.pressure / 1.01325
+        return self.pressure / UnitConstants.STANDARD_PRESSURE_BARA
 
     @classmethod
     def from_celsius(cls, temperature_celsius: float, pressure: float) -> ProcessConditions:
         """Create instance with temperature specified in Celsius."""
-        return cls(temperature=temperature_celsius + 273.15, pressure=pressure)
+        return cls(temperature=Unit.CELSIUS.to(Unit.KELVIN)(temperature_celsius), pressure=pressure)
 
     @classmethod
     def standard_conditions(cls) -> ProcessConditions:
--- a/src/libecalc/domain/process/core/stream/conditions.py
+++ b/src/libecalc/domain/process/core/stream/conditions.py
@@ -15,40 +15,40 @@ class ProcessConditions:
         pressure: Pressure in bara (absolute bar)
     """
 
-    temperature: float
-    pressure: float
+    temperature_kelvin: float
+    pressure_bara: float
 
     def __post_init__(self):
         """Validate conditions"""
-        if self.temperature <= 0:
-            raise ValueError(f"Temperature must be positive, got {self.temperature}")
-        if self.pressure <= 0:
-            raise ValueError(f"Pressure must be positive, got {self.pressure}")
+        if self.temperature_kelvin <= 0:
+            raise ValueError(f"Temperature must be positive, got {self.temperature_kelvin}")
+        if self.pressure_bara <= 0:
+            raise ValueError(f"Pressure must be positive, got {self.pressure_bara}")
 
     @property
     def temperature_celsius(self) -> float:
         """Get temperature in Celsius."""
-        return Unit.KELVIN.to(Unit.CELSIUS)(self.temperature)
+        return Unit.KELVIN.to(Unit.CELSIUS)(self.temperature_kelvin)
 
     @property
     def pressure_pascal(self) -> float:
         """Get pressure in Pascal."""
-        return Unit.BARA.to(Unit.PASCAL)(self.pressure)
+        return Unit.BARA.to(Unit.PASCAL)(self.pressure_bara)
 
     @property
     def pressure_atm(self) -> float:
         """Get pressure in atmospheres."""
-        return self.pressure / UnitConstants.STANDARD_PRESSURE_BARA
+        return self.pressure_bara / UnitConstants.STANDARD_PRESSURE_BARA
 
     @classmethod
-    def from_celsius(cls, temperature_celsius: float, pressure: float) -> ProcessConditions:
+    def from_celsius(cls, temperature_celsius: float, pressure_bara: float) -> ProcessConditions:
         """Create instance with temperature specified in Celsius."""
-        return cls(temperature=Unit.CELSIUS.to(Unit.KELVIN)(temperature_celsius), pressure=pressure)
+        return cls(temperature_kelvin=Unit.CELSIUS.to(Unit.KELVIN)(temperature_celsius), pressure_bara=pressure_bara)
 
     @classmethod
     def standard_conditions(cls) -> ProcessConditions:
         """Create instance with standard conditions."""
         return cls(
-            temperature=UnitConstants.STANDARD_TEMPERATURE_KELVIN,
-            pressure=UnitConstants.STANDARD_PRESSURE_BARA,
+            temperature_kelvin=UnitConstants.STANDARD_TEMPERATURE_KELVIN,
+            pressure_bara=UnitConstants.STANDARD_PRESSURE_BARA,
         )
--- a/src/libecalc/domain/process/core/stream/conditions.py
+++ b/src/libecalc/domain/process/core/stream/conditions.py
@@ -1,0 +1,54 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from libecalc.common.units import UnitConstants
+
+
+@dataclass(frozen=True)
+class ProcessConditions:
+    """
+    Represents the physical conditions of a process stream.
+
+    Attributes:
+        temperature: Temperature in Kelvin
+        pressure: Pressure in bara (absolute bar)
+    """
+
+    temperature: float
+    pressure: float
+
+    def __post_init__(self):
+        """Validate conditions"""
+        if self.temperature <= 0:
+            raise ValueError(f"Temperature must be positive, got {self.temperature}")
+        if self.pressure <= 0:
+            raise ValueError(f"Pressure must be positive, got {self.pressure}")
+
+    @property
+    def temperature_celsius(self) -> float:
+        """Get temperature in Celsius."""
+        return self.temperature - 273.15
+
+    @property
+    def pressure_pascal(self) -> float:
+        """Get pressure in Pascal."""
+        return self.pressure * 1e5
+
+    @property
+    def pressure_atm(self) -> float:
+        """Get pressure in atmospheres."""
+        return self.pressure / 1.01325
+
+    @classmethod
+    def from_celsius(cls, temperature_celsius: float, pressure: float) -> ProcessConditions:
+        """Create instance with temperature specified in Celsius."""
+        return cls(temperature=temperature_celsius + 273.15, pressure=pressure)
+
+    @classmethod
+    def standard_conditions(cls) -> ProcessConditions:
+        """Create instance with standard conditions."""
+        return cls(
+            temperature=UnitConstants.STANDARD_TEMPERATURE_KELVIN,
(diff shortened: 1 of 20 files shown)


src/libecalc/domain/process/core/stream/conditions.py
```diff
@@ -0,0 +1,54 @@
...
+        """Get pressure in Pascal."""
+        return Unit.BARA.to(Unit.PASCAL)(self.pressure)
+
+    @property
+    def pressure_atm(self) -> float:
+        """Get pressure in atmospheres."""
+        return self.pressure / UnitConstants.STANDARD_PRESSURE_BARA
+
+    @classmethod
+    def from_celsius(cls, temperature_celsius: float, pressure: float) -> ProcessConditions:
```
TeeeJay, line 44: not provided info which unit pressure is expected to be

kjbrak: What do you mean? The output is atmosphere (atm). The self.pressure is in bara is mentioned in docsting:
pressure: Pressure in bara (absolute bar)

TeeeJay: hmm, here you send in `pressure` as parameter. I know it is assumed bara from docstring in constructor, but that is not being informed about in this method ...

src/libecalc/domain/process/core/stream/conditions.py
```diff
@@ -0,0 +1,54 @@
...
+class ProcessConditions:
+    """
+    Represents the physical conditions of a process stream.
+
+    Attributes:
+        temperature: Temperature in Kelvin
+        pressure: Pressure in bara (absolute bar)
+    """
+
+    temperature: float
```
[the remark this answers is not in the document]
kjbrak: Might be considered yes. I was thinking about dropping the whole ProcessConditions class since it might just be a redundant layer. I realised when working with the new streams now that it might just be bloat. Maybe have a small talk about it later.

You don't think it will be a penalty for passing around these value objects instead of floats, or instead of the Proc.Cond. object with these value objects instead of floats?

src/libecalc/domain/process/core/stream/thermo_adapters.py
```diff
@@ -0,0 +1,138 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from ecalc_neqsim_wrapper.thermo import NeqsimFluid
+from libecalc.domain.process.core.stream.conditions import ProcessConditions
+from libecalc.domain.process.core.stream.fluid import ThermodynamicEngine
+
+# Only import Fluid for type checking to avoid circular import
+if TYPE_CHECKING:
```
TeeeJay, line 10: it is fair to do this when we dont know why we get ciruclar imports, but here I think it is because we mix ports/interface and adapters/implementation - neqsim should only be within its neqsim module - but fulfill the requirements of the port/interface/protocols ...

src/libecalc/domain/process/core/stream/conditions.py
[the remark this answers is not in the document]
TeeeJay: There is definitely always a perforamnce penalty when boxing/unboxing data types, but it may be negligible? I do not have overview here, so I am not sure myself. In general I prefer readability > performance first, and if there is a performance hit (that we cannot ignore), then we have to make a trade-off perhaps?

TeeeJay (approved):
huge improvement - in general easy to read and understand - which paradoxically leads to more comments ;)

src/libecalc/domain/process/core/stream/stream.py
```diff
@@ -0,0 +1,208 @@
...
+    Attributes:
+        fluid: Fluid object containing composition and EoS model
+        conditions: Process conditions (temperature and pressure)
+        mass_rate: Mass flow rate [kg/h]
+        name: Optional identifier for the stream
+    """
+
+    fluid: Fluid
+    conditions: ProcessConditions
+    mass_rate: float
```
olelod, line 38: Today all input is given in standard rates. This means that today we need a call to neqsim (or something else) first to find mass_rate. Should this take standard rate? Or should it be some kind of object prepared also to handle molar rates?

src/libecalc/domain/process/core/stream/conditions.py
kjbrak: Its not an argument (no pun intended) since its using the instance attribute itself.

src/libecalc/domain/process/core/stream/stream.py
olelod: Ok, now I noticed the from_standard_rate method also....

src/libecalc/domain/process/core/stream/conditions.py
kjbrak: I changed the attributes to be explicit: pressure_bara and temperature_kelvin

src/libecalc/domain/process/core/stream/fluid.py
```diff
@@ -0,0 +1,106 @@
...
+    _thermodynamic_engine: ThermodynamicEngine | None = None  # Will be initialized in __post_init__
+
+    def __post_init__(self):
+        """Initialize the thermodynamic engine if not provided"""
+        # If engine is already provided, we don't need to create one
+        if self._thermodynamic_engine is not None:
+            return
+
+        # Import adapters here to avoid circular import at runtime
+        from libecalc.domain.process.core.stream.thermo_adapters import NeqSimThermodynamicAdapter
```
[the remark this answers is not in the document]
kjbrak: I have made a new fluid_factory.py with a fluid factory, this keeps fluid domain object and and adapter separated

src/libecalc/domain/process/core/stream/stream.py
kjbrak: I hope that method can be used, and maybe its "cleaner" to have mass rate the main rate attribute. Yes, if we are to implement feed molar flow rates then we could make a "from molar rates" method or something..

src/libecalc/domain/process/core/stream/fluid.py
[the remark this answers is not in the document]
kjbrak: I chose to keep the fluid_factory in the domain stream dir since I think its cleaner than having it somewhere else. Maybe take a look at the changes? I will resolve the conversation.

src/libecalc/domain/process/core/stream/conditions.py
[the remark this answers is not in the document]
kjbrak: I feel like adding classes for temperature and pressure would be bloated/overkill or an abstraction step too much. Doesnt feel like the right thing to do.

src/libecalc/domain/process/core/stream/stream.py
```diff
@@ -0,0 +1,208 @@
...
+from functools import cached_property
+from typing import ClassVar, Protocol
+
+from libecalc.common.units import UnitConstants
+from libecalc.domain.process.core.stream.conditions import ProcessConditions
+from libecalc.domain.process.core.stream.fluid import Fluid
+from libecalc.domain.process.core.stream.mixing import SimplifiedStreamMixing
+
+
+class StreamMixingStrategy(Protocol):
```
jsolaas, line 13: Could we use an abc instead, and be explicit in `SimplifiedStreamMixing`? We could do the same with a protocol, just specifying it in simplified.

src/libecalc/domain/process/core/stream/stream.py
```diff
@@ -0,0 +1,208 @@
...
+    def mix(cls, streams: list[Stream]) -> Stream:
+        """Mix multiple streams into one.
+
+        Args:
+            streams: List of streams to mix, see MixingStrategy for more details
+
+        Returns:
+            A new Stream instance representing the mixed stream
+        """
+        return cls._stream_mixing_strategy.mix_streams(streams)
```
jsolaas, line 179: Do we need this? Could we instead use MixingStrategy outside of stream?

src/libecalc/domain/process/core/stream/stream.py
```diff
@@ -0,0 +1,208 @@
...
+        mass_rate: Mass flow rate [kg/h]
+        name: Optional identifier for the stream
+    """
+
+    fluid: Fluid
+    conditions: ProcessConditions
+    mass_rate: float
+
+    # Default mixing strategy (class variable)
+    _stream_mixing_strategy: ClassVar[StreamMixingStrategy] = SimplifiedStreamMixing()
```
jsolaas, line 41: If keeping mixing strategy, shouldn't it be something we expect to be passed in to Stream? So not a private property?

jsolaas (approved):
Looks very good!

src/libecalc/domain/process/core/stream/stream.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def __post_init__(self):
+        """Validate stream properties"""
+        if self.mass_rate < 0:
+            raise ValueError(f"Mass rate must be non-negative, got {self.mass_rate}")
```
jsolaas, line 46: Should probably be a Stream specific exception, easier to capture it. InvalidstreamException as parent, `NegativeMassRateException` in this case? Or something like that

src/libecalc/domain/process/core/stream/stream.py
kjbrak: Yes, the way I'm thinking now we choose a mixing strategy that is normally used (simplified mixing, no enthalpy balance), but include the possibility to change it later for specific use cases if we see the need for a full enthalpy balance, i.e. in cases where temperature and/or Cp value or composition is expected to be notably different. This way we dont care about the mixing strategy being used normally, but have the option to use another one later if needed.

You made me realize I must remove the ClassVar to make it possible to set a per instance strategy, as is the intention now. I'm keeping it a private property since it indicates its an internal property not meant for "external" acccess. Ok?

jsolaas: Yes, it makes sense to have it as a private property, but as input to `__init__`. I think it makes sense to not use dataclass here. 

```python
class Stream:
    def __init__(
        self,
        fluid: Fluid,
        conditions: ProcessConditions,
        mass_rate: float,
        mixing_strategy: StreamMixingStrategy,
    ):
        self.fluid = fluid
        self.conditions = conditions

        if self.mass_rate < 0:
            raise ValueError(f"Mass rate must be non-negative, got {self.mass_rate}")

        self.mass_rate = mass_rate
        self._mixing_strategy = mixing_strategy
```

kjbrak: I re-introduced frozen=True on the dataclass here since it had disappeared for some reason. That makes it immutable, which makes sense? It was also mentioned by Ole Petter to have immutability on new stream object because he had seen that this could be wise. Would it make sense with frozen dataclass here to make it immutable, if not why?

kjbrak: Now fluid, stream and process cond. are all immutable, which seems like a good idea to protect from changing the instances in an unpredictable manner. It also makes sense physically to not be able to change a fluid or stream, but be forced to create new ones.

jsolaas: I guess we already have the separate class `StreamMixingStrategy`

kjbrak: Yes, If we remove the mix method from Stream then Stream doesnt need a mixing strategy.. Maybe that is preferable. Then mixing class would be like a "process unit operation" (like a kind of equipment). Was thinking about this earlier too, its probably cleaner 👍

src/libecalc/domain/process/core/stream/stream.py
kjbrak: Now its explicit that the mixing strategy "inherits" the protocol.

src/libecalc/domain/process/core/stream/stream.py
kjbrak: I removed it, now we use only the separate class and method for mixing

src/libecalc/domain/process/core/stream/stream.py
kjbrak: Now it looks better. Removed mix method from Stream and hence also the mixing strategy.

src/libecalc/domain/process/core/stream/thermo_adapters.py
kjbrak: This has been fixed when making fluid factory

src/libecalc/domain/process/core/stream/conditions.py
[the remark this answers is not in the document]
kjbrak: Will take a look at this as a potential separate add-on feature after we discuss it.

PR #832, opened 2025-03-26
make sure composition is normalised before calculating avg Mw

kjbrak:

Need to make sure composition is normalised before calculating avg mw.


TeeeJay (approved):
remember to use conventional commits in commit message, as here this is a `fix:`

src/ecalc_neqsim_wrapper/thermo.py
```diff
@@ -432,7 +432,7 @@ def calculate_molar_mass(composition: FluidComposition) -> float:
     Returns:
         float: The molar mass of the mixture in kg/mol
     """
-    comp_dict = composition.model_dump()
+    comp_dict = composition.normalized().model_dump()
```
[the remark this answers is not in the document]
kjbrak: Ok.
- Changed the Mw constants to be imported from fluid domain instead and deleted the redundant one in common fluid. I added a get_component_molecular_weight method on that one.
- Deleted the redundant Mw constants and the test for that one.
- Added items method on the fluid composition class.

tests/libecalc/domain/core/stream/test_thermo_constants.py
```diff
@@ -9,8 +9,4 @@ def test_validate_components(self):
 
         This test simply verifies that the component definitions are structurally valid.
         """
-        # Should run without raising exceptions
-        ThermodynamicConstants.validate_components()
-
-        # If we reach here, the validation passed
-        assert True
+        assert ThermodynamicConstants.validate_components()
```
jsolaas, line 12: I would move the contents of validate_components into the test, unless we need to use that method runtime. Seems to be there to keep the classes in sync?

jsolaas (approved):
Nice!

tests/libecalc/domain/core/stream/test_thermo_constants.py
kjbrak: Yes to make sure constants have all the components in our FluidComposition.
Ok, so not validate at module import? It makes sense, but I thought it was such a critical validation check that just having it in a test seemed too weak and that it made sense to explicitly validated in the code at import of the constants? But I see your point. Can move remove it from the class itself and use it only in the test.

kjbrak: done

PR #833, opened 2025-03-26
ci: update to release-please v4 with default behavior

Aleksander-Karlsson:

Make libecalc follow standard semantic versioning.

Config for release-please (and updates version)

https://github.com/equinor/ecalc-internal/issues/593


.github/workflows/on-push-release-branch.yml
```diff
@@ -16,7 +16,5 @@ jobs:
 
   release-please:
     uses: ./.github/workflows/release-please.yml
-    with:
```
TeeeJay, line 19: we need a way to be able to make patches AND to avoid conflict with versions on main branch. Didnt think of that... how? Possible to add a new suffix that is both supported by release-please AND pypi AND pip etc? @jsolaas

.github/workflows/release-please.yml
```diff
@@ -1,54 +1,19 @@
...
+  workflow_dispatch: # Workflow dispatch is used for manual triggers.
+  workflow_call: # Workflow call is used for called from another workflow.
 
 jobs:
   release-please:
     permissions: write-all
     runs-on: ubuntu-24.04
     steps:
-      - uses: google-github-actions/release-please-action@v3
+      - uses: googleapis/release-please-action@v4
```
TeeeJay, line 12: superb that you fix this as welL! 💯

.github/workflows/on-push-release-branch.yml
jsolaas: Is this an issue? How does release-please select the next version? We can override the release version if that happens. Then look into a solution? 

https://github.com/googleapis/release-please/issues/719

TeeeJay: release-please only knows about previous version, on the given branch it resides (right?) - and would not know about a potential conflict in versions ... so the developer would have to have the overview to avoid that to happen...and it would be better if he/she didnt have to do that. What happens if there is a conflict? Probably sth that needs to be fixed...idk.

Aleksander-Karlsson: Hmm yeah I think a dev would need to have some oversight but I think release-please will 'fail' if there is a conflict. Could it be enough to resolve conflicts then?

TeeeJay: mhm, there is nothing wrong about "lets test and see what happens", to make a better decision (if needed)

PR #842, opened 2025-04-03
feat: implement ProcessSystem and ProcessUnit interfaces

jsolaas:


src/libecalc/domain/energy/energy_component.py
```diff
@@ -44,6 +46,12 @@ def is_electricity_consumer(self) -> bool:
         """Returns True if the component consumes electricity"""
         ...
 
+    @abc.abstractmethod
+    def get_process_change_events(self) -> list[ProcessChangeEvent]: ...
```
TeeeJay, line 50: Event! <3

src/libecalc/domain/energy/process_change_event.py
```diff
@@ -0,0 +1,14 @@
+from dataclasses import dataclass
+
+from libecalc.common.time_utils import Period
+
+
+@dataclass(frozen=True)
+class ProcessChangeEvent:
```
TeeeJay, line 7: I think I would want the events to be in past tense; because that reflects the immutability, that it "has happened" ...

src/libecalc/domain/energy/process_change_event.py
```diff
@@ -0,0 +1,14 @@
...
+    """
+    An event representing a change in the process system
+    """
+
+    period: Period
```
TeeeJay, line 12: I think an event is defined at a timestep, and that all proceeding timesteps&periods are affected, or until a new event ... not sure if we think the same here, lets talk ...

PR #867, opened 2025-04-11
refactor: remove generic from process unit and system

jsolaas:
It doesn't make sense with generic without an input of the generic type


frodehk (approved):
Nice

PR #870, opened 2025-04-24
chore: add compressor stage abc

jsolaas:
Create CompressorStage abc to define what is expected from a compressor
stage. Currently only a compressor chart is expected specifically for
CompressorStage.

### DIFF
--- a/src/libecalc/domain/process/process_system.py
+++ b/src/libecalc/domain/process/process_system.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
 import abc
-from typing import Protocol, TypeVar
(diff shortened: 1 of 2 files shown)


src/libecalc/domain/process/process_system.py
```diff
@@ -29,9 +31,6 @@ class LiquidStream(Stream):
     ...
 
 
-TStream = TypeVar("TStream", bound=LiquidStream | MultiPhaseStream)
```
TeeeJay, line 32: not in use?

TeeeJay (approved):

PR #878, opened 2025-05-05
refactor: make compressor models point in time

olelod:
Preparation for passing stream object to compressor models instead of rate/pressures.

Make all the compressor models take single time step input (in essence using floats instead of numpy arrays).

Update compressor model test


tests/ecalc_cli/snapshots/test_app/test_json_advanced_model/test_json_advanced_model_v3.json
```diff
@@ -34115,28 +34115,28 @@
           },
           "unit": "Am3/h",
           "values": [
-            0.0,
```
olelod, line 34118: The differences in this snapshot is due to the fact that we used to calculate all time steps simultaneously for simplified trains, now we do it one-by-one. This is results for one operational setting for a simplified train, used only two times. Now all other time steps have empty results - before they were calculated with rate 0.

PR #879, opened 2025-05-07
Fix compressor solver issue

kjbrak:

There was issues with the solver in the compressor discharge pressure calculations in an edge case for Johan Sverdrup case study where inlet z factor is around 0.36. Both issue with eos call crashing due to too large initial guess on pressure and an issue with solver instability during the iterations.

Max cap on pressure in ph flash call to avoid crash and adaptive dampening algorithm implemented in solver.

ref.
https://github.com/equinor/ecalc-internal/issues/759


olelod (commented):
test_all_energy_usage_models does not fail locally for me (probably not for you either...?). all differences are in the 3rd, 4th or 5th decimal, so nothing serious, but I guess we should understand why.

olelod (approved):

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.4 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
```
TeeeJay, line 1: usikker på om denne fortsatt er i brukt (har ikke sjekka), iom at vi går over til `uv` isf Poetry. Mulig det ikke er satt opp for libecalc el CICD enda ...

PR #895, opened 2025-05-19
refactor: add evaluate method for stage given speed and target discha…

olelod:
…rge pressure

Today variable speed compressor trains use single speed compressor train at minimum speed for INDIVIDUAL_ASV_PRESSURE. This should allow for making a better method for variable speed compressor trains

Make a method on the CompressorStage class that can evaluate both single and variable speed compressor charts given a speed (for variable speed) and target discharge pressure. This method was "hidden" inside the single speed compressor train. It should now be possible to make a method of INDIVIDUAL_ASV_PRESSURE for variable speed compressor trains without going via a single speed compressor train.


TeeeJay:
nice :)

TeeeJay (approved):

PR #896, opened 2025-05-20
fix: key error speed

Aleksander-Karlsson:

https://github.com/equinor/ecalc-internal/issues/755

### DIFF
--- a/src/libecalc/presentation/yaml/mappers/utils.py
+++ b/src/libecalc/presentation/yaml/mappers/utils.py
@@ -157,6 +157,9 @@ def convert_control_margin_to_fraction(control_margin: float | None, input_unit:
 def chart_curves_as_resource_to_dto_format(resource: Resource) -> list[dict[str, list[float]]]:
     resource_headers = resource.get_headers()
 
+    if "SPEED" not in resource_headers:
+        raise InvalidColumnException(message="Resource is missing SPEED header!", header="SPEED")
+
     resource_data = []
     for header in resource_headers:
         column = resource.get_column(header)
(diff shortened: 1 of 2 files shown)


src/libecalc/presentation/yaml/mappers/utils.py
```diff
@@ -167,6 +167,10 @@ def chart_curves_as_resource_to_dto_format(resource: Resource) -> list[dict[str,
             msg = f"Resource contains non-numeric value: {e}"
             logger.error(msg)
             raise InvalidColumnException(message=msg, header=header) from e
+    # Ensure speed header is set and filled with ones.
+    if "SPEED" not in resource_headers:
+        resource_headers.append("SPEED")
+        resource_data.append([1] * len(resource.get_column(resource_headers[0])))
```
TeeeJay, line 173: Speed was 0 as default earlier. Why set to 1? I saw this recently occured, why? Why is speed missing in the first place?

Aleksander-Karlsson: I thought it was defaulted to 1, ref:
https://github.com/equinor/ecalc/blob/ae718821b4968ca6a9f12deb5cb7188f04183577/src/libecalc/presentation/yaml/mappers/utils.py#L273

But I've made a change as I realise it should only reach this point if this is in fact a Variable speed pump, and then it must contain SPEED right?

TeeeJay (approved):

src/libecalc/presentation/yaml/mappers/utils.py
TeeeJay: yes, variable speed we must know the speed. For single speed it is either left out or specified in yaml I guess?

PR #899, opened 2025-05-22
refactor: add energy network

jsolaas:


src/libecalc/domain/energy/energy_network.py
```diff
@@ -0,0 +1,52 @@
...
+
+class EnergyNetwork(abc.ABC):
+    @abc.abstractmethod
+    def get_energy_nodes(self) -> list[PowerConsumer | PowerProvider | FuelConsumer]: ...
+
+    @abc.abstractmethod
+    def get_power_providers(self) -> list[PowerProvider]: ...
+
+    @abc.abstractmethod
+    def get_power_consumers(self, energy_node_id: UUID) -> list[PowerConsumer]: ...
```
frodehk, line 47: Maybe more precise naming, do we mean all consumers (`get_all_power_consumers`), or downstream consumers for a given node (all that are supplied by that node -> `get_downstream_power_consumers`)?

PR #908, opened 2025-05-27
test: remove redundant test for operational settings used

frodehk:

The test `test_operational_settings_used_available` was originally added to verify the presence of the operational_settings_used column in CSV output. Since this attribute follows the same structure and export logic as other `TimeSeries` attributes, and is already covered by general CSV export and validation tests, the dedicated test is redundant.

This PR removes the `test_operational_settings_used_available` test from `test_app.py`. No functional code is changed; only the unnecessary test is deleted.


tests/ecalc_cli/test_app.py
```diff
@@ -176,20 +176,20 @@ def test_csv_temporal_default(self, simple_temporal_yaml_path, simple_yaml_path,
...
+            _get_args(model_file=pump_system_yaml_path, csv=True, output_folder=tmp_path, name_prefix=run_name_prefix),
             catch_exceptions=False,
         )
         run_csv_output_file = tmp_path / f"{run_name_prefix}.csv"
         assert run_csv_output_file.is_file()
         df = pd.read_csv(run_csv_output_file, index_col="timesteps")
         operational_settings_used = df["Water injection pump system A.operational_settings_used[N/A]"].tolist()
         is_valid = df["Water injection pump system A.is_valid[N/A]"].tolist()
-        assert operational_settings_used == [3, 3, 3, 3, 3, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
+        assert operational_settings_used == [3, 3, 3, 3, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
```
TeeeJay, line 192: it seems to me immediately that this test is too simple. Which edge cases are tested? Is crossover tested? What about unhappy paths, where none are valid etc

frodehk: It is basically checking that we are providing operational settings used for systems. There are no changes to what is being tested in this PR. The difference is that we only include the pump system that is part of the assert, and remove everything else (i.e. minimal yaml as opposed to the advanced yaml which is tested in `TestJsonOutput.test_json_advanced_model`)

tests/ecalc_cli/test_app.py
```diff
@@ -5,14 +5,25 @@
...
 
+import numpy as np
 import pandas as pd
 import pytest
 import yaml
 from typer.testing import CliRunner
 
 from ecalc_cli import main
 from libecalc.common.run_info import RunInfo
+from libecalc.common.serializable_chart import ChartCurveDTO, SingleSpeedChartDTO
```
TeeeJay, line 16: hmm, I assume these imports should not be here or?

PR #917, opened 2025-06-05
chore: upgrade dependencies for week 23. Auto-generated by devCalc

Aleksander-Karlsson:
Had to manually lock pandas versioning a bit to get around an error poetry had with installing pandas 2.3.0

```Note: This error originates from the build backend, and is likely not a problem with poetry but with pandas (2.3.0) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "pandas (==2.3.0)"'.```


TeeeJay:
ok, komodo still has pandas 2.2.3 as default, but may change soon

TeeeJay (approved):

PR #936, opened 2025-06-12
refactor: remove output frequency from global time vector

olelod:

- Removing `_get_end_boundary` where the selected output frequency was (potentially) used to find the end date in the global time vector. The end date should be independent of the output frequency. 
- Removing all tests connected to `_get_end_boundary`.
- Making END a required keyword in the YAML file


tests/ecalc_cli/snapshots/test_app/test_new_stp_export_properly/test.INST_A.stp.tsv
```diff
@@ -145,14 +145,14 @@ forecastYear	forecastMonth	co2Emission	methaneEmission
...
-2033	04	0	0
-2033	05	0	0
-2033	06	0	0
-2033	07	0	0
-2033	08	0	0
-2033	09	0	0
-2033	10	0	0
-2033	11	0	0
-2033	12	0	0
+2033	02	4714.785	2.033391
```
olelod, line 148: Here eCalc is run with MONTH as input frequency, EXTRAPOLATION is FALSE for time series, and END is specified to be 2034-01-01. This means that the input from 2033-01-01 will give values for all of 2033. Earlier the first day of each month would be added to the global time vector, so the values for the last date in the input would only be valid for Jan 2033 - the rest of the months would be set to 0.

TeeeJay (approved):
Oppdater i selve doccen for `END:` også?

PR #938, opened 2025-06-12
feat: provide co2 intensity results for cli

frodehk:
Refs equinor/ecalc-internal#534

This pull request adds functionality to calculate and report CO2 intensity as part of the CLI output. This feature is needed for providing CLI-users with detailed insights into emission intensities.

- Implements CO2 intensity calculation in the CLI using the `EmissionIntensityCalculator`.
- Adds support for exporting CO2 intensity results in **separate** CSV and JSON files, distinct from the main results.
- New test cases for CO2 intensity calculations, and move old tests. All tests for intensity are found in:
  - tests/ecalc_cli/test_emission_intensity.py
  - tests/libecalc/domain/emission/test_emission_intensity.py
- **Note:** The relocation of `TimeSeriesIntensity` **requires** corresponding **changes in backend** to ensure compatibility.
- **Note:** This functionality has already been available in the web interface but has not been accessible in the CLI until now.
- Consolidates functionality:
  - The CLI logic is organized under ecalc_cli.
  - Domain logic for calculating intensities is structured under domain.emission.
  - Tests are grouped under tests/ecalc_cli for CLI-related functionality and tests/libecalc/domain/emission for domain logic.


src/libecalc/domain/emission/emission_intensity.py
```diff
@@ -1,9 +1,68 @@
...
+
+
+class EmissionIntensityResults(EcalcResultBaseModel):
+    results: list[EmissionIntensityResult]
+
+    def resample(self, freq: Frequency) -> Self:
+        return EmissionIntensityResults(results=[r.resample(freq) for r in self.results])
+
+
+def write_emission_intensity_csv(emission_intensity_results, output_file, date_format):
```
TeeeJay, line 24: this is infrastrcture - and should not be in domain layer - is it possible to reuse `write_output` and `get_csv`?

src/libecalc/domain/emission/emission_intensity.py
```diff
@@ -1,9 +1,68 @@
...
+    for result in emission_intensity_results.results:
+        df = result.to_dataframe(prefix=result.name)
+        dfs.append(df)
+    # Join on index (timesteps)
+    combined_df = pd.concat(dfs, axis=1)
+    combined_df.index = combined_df.index.strftime(date_format)
+    combined_df.to_csv(output_file, index_label="timesteps")
+
+
+def calculate_emission_intensity(
```
TeeeJay, line 36: isnt this implemented somewhere else already? i may be missing something

frodehk: I is moved from the json_result mapper (where it was unused btw). So yes, it was implemented another place - but moved to here. Just to collect all intensity-stuff in one place.

src/libecalc/domain/emission/emission_intensity.py
TeeeJay: We might want to rename it to `to_csv` and return a string? Ideally serialization should be outside domain, but since this class is the only class that has the needed knowledge - it makes sense to keep it here. Otherwise some external class will need to have too much knowledge about internals in order to generate a proper csv ... I think ...

src/ecalc_cli/commands/run.py
```diff
@@ -167,6 +189,13 @@ def run(
             )
             write_output(output=csv_data, output_file=output_prefix.with_suffix(".csv"))
 
+            # Emission intensity CSV
```
TeeeJay, line 192: I guess it has to be like this, that it is included as a default when running model. Alternative would be to have a separate app that calculated intensity based on previously stored result, which is probably at least too early to do something about now ..?

frodehk: You want the logic of creating the intensity `csv_data` moved from `write_emission_intensity_csv` to run.py? Now `csv_data` is created inside `write_emission_intensity_csv` which in turn calls `write_output`.

TeeeJay: that is my suggestion :) I think that is better and clearer, and clear separtion of responsibilities. I dont see the benefit here to do the "Dependency Inversion Principle", it will make the code less clear imo.

tests/ecalc_cli/test_app.py
```diff
@@ -176,6 +177,34 @@ def test_csv_temporal_default(self, simple_temporal_yaml_path, simple_yaml_path,
 
         assert df_temporal.equals(df_basic)
 
+    def test_intensity_csv_content(self, simple_yaml_path, tmp_path):
```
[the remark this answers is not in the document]
frodehk: The test is removed, and its functionality is transferred to new tests in domain instead (no need to run CLI etc.)

src/libecalc/domain/emission/emission_intensity.py
frodehk: In backend I think the logic is handled in api/modules/run/domain/results.py in `EmissionResult.intensity`. It is using the `EmisionIntensity` class from libecalc, but not the `calculate_emission_intensity`.

src/libecalc/domain/emission/emission_intensity.py
frodehk: It is renamed to `emission_intensity_to_csv`.

PR #942, opened 2025-06-17
refactor: don't use dto.ConsumerFunciton in fuel/el-consumers

jsolaas:
Map directly from yaml to infrastructure ConsumerFunction. Next change
will be to remove the dto.ConsumerFunctions.


Aleksander-Karlsson (commented):
Included a couple of suggestions to maybe improve error messages

olelod (approved):

PR #946, opened 2025-06-18
refactor: remove dto.TabulatedConsumerFunction

jsolaas:


src/libecalc/presentation/yaml/mappers/consumer_function_mapper.py
```diff
@@ -431,24 +450,39 @@ def from_yaml_to_dto(
     ) -> TemporalModel[ConsumerFunction]:
         time_adjusted_model = define_time_model_for_period(data, target_period=self._target_period)
 
-        temporal_dict = {}
+        temporal_dict: dict[Period, ConsumerFunction] = {}
         for period, model in time_adjusted_model.items():
```
frodehk, line 454: Instead of many isinstance-checks, could we have to_dto-methods in the Yaml-classes instead?

frodehk (approved):

PR #959, opened 2025-06-19
Ci/use pytest split jpype

jsolaas:
Run JPype with pytest-split


.github/workflows/lib-ci.yml
```diff
@@ -41,32 +42,7 @@ jobs:
 
       - name: Run tests with coverage
```
frodehk, line 43: How do we plan to monitor code coverage going forward? If there’s an alternative solution, maybe mention it in the PR description or documentation?

tj098895: atm we are not using it, and the integration tests are misleading wrt. coverage imo. we should rather build this from ground up, good coverage on unit tests, make things testable at a unit level.

PR #974, opened 2025-06-24
fix: require END to be set in yaml

jsolaas:


olelod:
Included in https://github.com/equinor/ecalc/pull/936

PR #1003, opened 2025-07-04
refactor: introduce Condition class

frodehk:

This pull request introduces the `Condition` class as a central domain component for handling conditional logic on numerical arrays. The motivation is to centralize and standardize how conditions are evaluated and applied, replacing scattered implementations. This abstraction improves code clarity, maintainability, and testability, especially as conditional masking is a recurring requirement in the domain logic.

- Adds the `Condition` class, which encapsulates the logic for parsing, evaluating, and applying condition expressions to arrays.
- Provides a clear API for applying conditions, supporting both NumPy array and Python list outputs.
- Refactors relevant parts of the codebase to utilize the new `Condition` class, reducing code duplication and potential for errors:
  - [x] `TabularConsumerFunction`
  - [x]  `DirectExpressionConsumerFunction`
  - [x]  `PumpConsumerFunction`
  - [x]  `CompressorConsumerFunction`
  - [x]  `PumpSystemConsumerFunction`
  - [x]  `CompressorSystemConsumerFunction`
- Improves documentation and code structure around conditional operations, making future extensions and maintenance easier.

By introducing a dedicated `Condition` abstraction, the codebase gains a more robust and reusable foundation for conditional array operations, supporting both current and future needs.

<hr></hr> 

**Note:**
This is one of several steps in decoupling domain logic from the `ExpressionEvaluator` class. By introducing the `Condition` abstraction, the codebase is being prepared for further refactoring where domain components will rely less on direct usage of `ExpressionEvaluator`.

### DIFF
--- a/src/libecalc/domain/condition.py
+++ b/src/libecalc/domain/condition.py
@@ -8,21 +8,24 @@ from libecalc.expression.expression import ExpressionType
 
 class Condition:
     """
-    Represents a condition that can be applied to arrays based on an expression.
+    Represents a class for evaluating and applying conditional logic to numerical arrays.
 
-    The Condition class encapsulates logic for evaluating a condition expression
-    and applying it to numerical arrays, typically to mask or zero out values
-    where the condition is not met.
-
-    Attributes:
-        _expression_input (ExpressionType | None): The raw input for the condition expression.
-        expression: The parsed and validated expression, or None if not provided.
+    The `Condition` class encapsulates the logic for parsing, validating, and evaluating
+    a condition expression, typically used to mask or zero out values in arrays where
+    the condition is not met. This centralizes conditional logic, improving code clarity
+    and maintainability.
     """
 
     def __init__(
         self,
         expression_input: ExpressionType | None,
     ):
+        """
+        Initialize a Condition instance.
+
+        Args:
+            expression_input (ExpressionType | None): The raw input for the condition expression.
+        """
         self._expression_input = expression_input
         self.expression = convert_expression(self._expression_input)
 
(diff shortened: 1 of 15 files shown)


src/libecalc/domain/condition.py
```diff
@@ -0,0 +1,94 @@
+import numpy as np
+from numpy._typing import NDArray
+
+from libecalc.common.variables import ExpressionEvaluator
+from libecalc.dto.utils.validators import convert_expression
+from libecalc.expression.expression import ExpressionType
+
+
+class Condition:
```
TeeeJay, line 9: this depends heavily on stuff in `common` and should probably be placed there, along with `Expression`. I think it definitely makes sense, as you suggest, to have a common way of doing this, and that this could serve as a separate library, that could as well be used in "other apps"; A way of creating a mathematical expression language, provided as strings, that also has boolean logic (condition) as a part of it.

src/libecalc/domain/condition.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    Represents a class for evaluating and applying conditional logic to numerical arrays.
+
+    The `Condition` class encapsulates the logic for parsing, validating, and evaluating
+    a condition expression, typically used to mask or zero out values in arrays where
+    the condition is not met. This centralizes conditional logic, improving code clarity
+    and maintainability.
+    """
+
+    def __init__(
```
TeeeJay, line 19: looking at the code, I would expect that it accepted an expression (probably that would already have variables expanded to numbers?) - and made sure that the boolean logic was done correctly (correct order wrt. PEMDAS/precedence is correct) - and that it is evaluated as boolean logic vs. the normal way of evaluating expressions, that are mathematical ...

I think we have gotchas in the code, where this might not always be as expected by end users wrt. boolean logic for conditions, so that should be clear.

I think it would be better to leave out variables etc here, as that should have been expanded/extrapolated already.

frodehk:
Condition will be addressed in other pull requests, related to time series rates

PR #1065, opened 2025-08-25
fix: also supply gasTurbineGeneratorConsumption for ltp if empty

tj098895:
If we have no matching category and query filter for any periods for a given ecalc model, the gasTurbineGeneratorConsumption column was skipped. It will now always be included.

This is needed in order to have a consistent LTP format even when columns are empty, for comparison and for export to centuries.

Refs: equinor/ecalc-internal#1031

### DIFF
--- a/src/libecalc/presentation/exporter/queries.py
+++ b/src/libecalc/presentation/exporter/queries.py
@@ -21,6 +21,20 @@ class Query(abc.ABC):
         unit: Unit,
         frequency: Frequency,
     ) -> dict[Period, float] | None:
+        """
+        Query the graph (result) to aggregate values based on the query parameters such as CATEGORY.
+
+        Only returns None if the installation category does not match.
+        Otherwise, it returns a dict with all periods and 0.0 values.
+
+        Args:
+            installation_graph:
+            unit:
+            frequency:
+
+        Returns:
+
+        """
         pass
 
 


tests/ecalc_cli/snapshots/test_app/test_new_ltp_export_properly/test.INST_A.ltp.tsv.json
```diff
@@ -385,18 +705,18 @@
   },
   "loadedAndStoredOil":{
     "0":"Total Oil Loaded\/Stored[Sm3]",
-    "1":"1095000",
```
tj098895, line 388: @frodehk should we also perhaps change the LTP model to use venting emitter, so that we get the same results as before?

tests/libecalc/presentation/exporter/test_ltp.py
```diff
@@ -827,105 +827,6 @@ def test_boiler_heater_categories(
         ltp_test_helper.assert_consumption(ltp_result, 0, "boilerFuelGasCo2Mass", ltp_test_helper.co2_from_boiler)
         ltp_test_helper.assert_consumption(ltp_result, 0, "heaterFuelGasCo2Mass", ltp_test_helper.co2_from_heater)
 
-    def test_total_oil_loaded_old_method(
```
tj098895, line 830: @frodehk hmm, do we have an equivalent test for "new_method"?

tj098895:
to be updated and merged next week

PR #1074, opened 2025-08-29
fix/refactor time series cable loss category

frodehk:
Refs equinor/ecalc-internal#1038

Previously, the cable loss was applied regardless of the electricity category, which led to incorrect results when the generator set was not categorized as "POWER_FROM_SHORE". The cable loss should only be considered for periods where the category is "POWER_FROM_SHORE". This fix is needed to ensure the domain logic matches the intended behavior and to prevent misleading power calculations.

- Refactors how cable loss and maximum usage from shore are handled in the generator set component.
- Introduces abstract interfaces (`TimeSeriesCableLoss`, `TimeSeriesMaxUsageFromShore`) for time series logic.
- Implements logic so that cable loss is only applied when the category is POWER_FROM_SHORE; otherwise, zero cable loss is used.
- Updates mapping and construction of generator set components to use the new time series abstractions.
- Adds tests to verify correct handling of cable loss depending on electricity category, including support for temporal switching between categories.

### DIFF
--- a/src/libecalc/domain/infrastructure/energy_components/generator_set/generator_set_component.py
+++ b/src/libecalc/domain/infrastructure/energy_components/generator_set/generator_set_component.py
@@ -33,9 +33,10 @@ from libecalc.domain.infrastructure.energy_components.generator_set import Gener
 from libecalc.domain.infrastructure.path_id import PathID
 from libecalc.domain.installation import ElectricityProducer, FuelConsumer, FuelConsumption
 from libecalc.domain.regularity import Regularity
+from libecalc.domain.time_series_cable_loss import TimeSeriesCableLoss
+from libecalc.domain.time_series_max_usage_from_shore import TimeSeriesMaxUsageFromShore
 from libecalc.dto.component_graph import ComponentGraph
 from libecalc.dto.fuel_type import FuelType
-from libecalc.expression import Expression
 
 
 class GeneratorSetEnergyComponent(Emitter, EnergyComponent, ElectricityProducer, FuelConsumer):
@@ -55,8 +56,8 @@ class GeneratorSetEnergyComponent(Emitter, EnergyComponent, ElectricityProducer,
         expression_evaluator: ExpressionEvaluator,
         consumers: list[ElectricityConsumer],
         fuel: TemporalModel[FuelType],
-        cable_loss: Expression | None = None,
-        max_usage_from_shore: Expression | None = None,
+        cable_loss: TimeSeriesCableLoss | None = None,
+        max_usage_from_shore: TimeSeriesMaxUsageFromShore | None = None,
         component_type: Literal[ComponentType.GENERATOR_SET] = ComponentType.GENERATOR_SET,
     ):
         self._uuid = id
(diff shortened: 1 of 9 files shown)


src/libecalc/presentation/yaml/mappers/component_mapper.py
```diff
@@ -390,12 +394,30 @@ def create_error(message: str, key: str) -> ModelValidationError:
...
+                )
+            else:
+                cable_loss = None
+
         except InvalidExpressionError as e:
             raise ComponentValidationException(errors=[create_error(str(e), key="CABLE_LOSS")]) from e
 
         try:
-            max_usage_from_shore = convert_expression(data.max_usage_from_shore)
+            max_usage_from_shore = (
```
tj098895, line 412: so, this was previously not an expression, or it was, but "not supported"? should update docs?

tests/ecalc_cli/snapshots/test_app/test_new_ltp_export_properly/test.POWER_FROM_SHORE_EVENT.ltp.tsv
```diff
@@ -1,7 +1,7 @@
 forecastYear	forecastMonth	turbineFuelGasConsumption	turbineFuelGasCo2Mass	turbineFuelGasNoxMass	turbineFuelGasNmvocMass	turbineFuelGasCh4Mass	gasTurbineGeneratorConsumption	fromShoreConsumption	powerSupplyOnshore
 #Years	Months	Fuel Consumption[Sm3]	CO2 From Fuel Gas[t]	NOX From Fuel Gas[t]	NMVOC From Fuel Gas[t]	CH4 From Fuel Gas[t]	Total Electricity Generated[GWh]	Total Electricity Consumed From Power-From-Shore[GWh]	Power Supply Onshore[GWh]
 2021	01	0	0	0	0	0	0	0	0
-2022	01	50156012	105829.2	459.4291	12.03744	45.64197	175.3752	0	0
+2022	01	50156012	105829.2	459.4291	12.03744	45.64197	159.432	0	0
```
tj098895, line 4: assume this is expected change, due to limitation that PfS is only active from 2023?

src/libecalc/presentation/yaml/mappers/component_mapper.py
frodehk: It was an expression earlier as well.

src/libecalc/domain/infrastructure/energy_components/generator_set/generator_set_component.py
```diff
@@ -267,11 +268,11 @@ def get_power_production(self) -> TimeSeriesRate:
     def get_maximum_power_production(self) -> TimeSeriesRate | None:
         if self.max_usage_from_shore is None:
             return None
-        max_power_production = self.expression_evaluator.evaluate(self.max_usage_from_shore)
+        max_power_production = self.max_usage_from_shore.get_values()
```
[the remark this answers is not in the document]
frodehk: Not sure if I understand, could you elaborate a bit more?

PR #1076, opened 2025-09-01
ci: dependabot

Aleksander-Karlsson:


requirements.txt
```diff
@@ -0,0 +1,86 @@
+annotated-types==0.7.0
```
tj098895, line 1: commit and push this to a dedicated branch? Can set `branch` in dependabot.yml

Aleksander-Karlsson: Ahh that should work, was hoping we could drop pip/requirments.txt entirely, but that does not seem to work for some reason 🤔

PR #1078, opened 2025-09-01
refactor: validate input before evaluating compressor model

jsolaas:
Cleans up mapping by removing some implementation specific checks. It
also makes the validation better for rate and pressure inputs as we can
check that they are provided when required.

### DIFF
--- a/src/libecalc/domain/process/compressor/core/sampled/compressor_model_sampled.py
+++ b/src/libecalc/domain/process/compressor/core/sampled/compressor_model_sampled.py
@@ -5,6 +5,7 @@ import pandas as pd
 from numpy.typing import NDArray
 from scipy.interpolate import interp1d
 
+from libecalc.common.consumption_type import ConsumptionType
 from libecalc.common.decorators.feature_flags import Feature
 from libecalc.common.energy_usage_type import EnergyUsageType
 from libecalc.common.list.adjustment import transform_linear
@@ -58,8 +59,7 @@ class CompressorModelSampled(CompressorModel):
             Often needed for energy reporting in e.g. LTP, STP.
         """
         logger.debug("Creating CompressorModelSampled")
-        self.data_transfer_object = data_transfer_object
-        self.function_values_are_power = self.data_transfer_object.energy_usage_type == EnergyUsageType.POWER
+        self.function_values_are_power = data_transfer_object.energy_usage_type == EnergyUsageType.POWER
         self.power_interpolation_values = data_transfer_object.power_interpolation_values
 
         function_values_adjusted: NDArray[np.float64] = transform_linear(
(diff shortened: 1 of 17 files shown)


src/libecalc/domain/process/compressor/core/sampled/compressor_model_sampled.py
```diff
@@ -146,12 +149,27 @@ def get_max_standard_rate(  # type: ignore[override]
...
+    def set_evaluation_input(
         self,
         rate: NDArray[np.float64],
-        suction_pressure: NDArray[np.float64],
-        discharge_pressure: NDArray[np.float64],
+        suction_pressure: NDArray[np.float64] | None,
+        discharge_pressure: NDArray[np.float64] | None,
         intermediate_pressure: NDArray[np.float64] | None = None,
+    ):
+        some_input = any(input is not None for input in [rate, suction_pressure, discharge_pressure])
```
tj098895, line 159: we dont use the variable, so the functional code can be used directly instead of using a temp var? Also, when needed for both check and variable, we can use the "walrus operator"

src/libecalc/domain/process/compressor/core/train/variable_speed_compressor_train_common_shaft_multiple_streams_and_pressures.py
```diff
@@ -83,6 +86,25 @@ def __init__(
...
 
+    def set_evaluation_input(
+        self,
+        rate: NDArray[np.float64],
+        suction_pressure: NDArray[np.float64] | None,
+        discharge_pressure: NDArray[np.float64] | None,
+        intermediate_pressure: NDArray[np.float64] | None = None,
+    ):
+        has_interstage_pressure = any(stage.has_control_pressure for stage in self.data_transfer_object.stages)
+        if has_interstage_pressure and intermediate_pressure is None:
```
tj098895, line 97: interstage and intermediate is "the same"?

src/libecalc/domain/process/compressor/core/train/variable_speed_compressor_train_common_shaft_multiple_streams_and_pressures.py
```diff
@@ -83,6 +86,25 @@ def __init__(
...
+            raise ValueError("Energy model requires interstage control pressure to be defined")
```
tj098895, line 98: nice :)

tj098895 (approved):

src/libecalc/domain/process/compressor/core/train/variable_speed_compressor_train_common_shaft_multiple_streams_and_pressures.py
tj098895: @olelod

src/libecalc/presentation/yaml/mappers/consumer_function_mapper.py
```diff
@@ -491,40 +393,37 @@ def _map_compressor_system(
...
         compressors = []
-        compressor_power_usage_type = set()
+        compressor_consumption_types: set[ConsumptionType] = set()
         for compressor in model.compressors:
             compressor_train = self.__references.get_compressor_model(compressor.compressor_model)
 
-            check_for_generic_from_input_compressor_chart_in_simplified_train_compressor_system(
-                compressor_train, name=compressor.name
-            )
+            # check_for_generic_from_input_compressor_chart_in_simplified_train_compressor_system(
```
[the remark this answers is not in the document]
tj098895: so, now it will "just (not) work" without any warning?

PR #1088, opened 2025-09-02
refactor: remove dto dependency from compressor train model

frodehk:

Tight coupling to DTOs in core compressor train models leads to unnecessary dependencies and reduced flexibility.

- Refactors core compressor train models to take explicit parameters instead of DTOs.
- Moves DTO-to-domain mapping outside the parent model (`CompressorTrainModel`) and into relevant child classes in the domain.
- Updates instantiation and tests to use new constructor arguments.

### DIFF
--- a/src/libecalc/domain/process/compressor/core/train/variable_speed_compressor_train_common_shaft.py
+++ b/src/libecalc/domain/process/compressor/core/train/variable_speed_compressor_train_common_shaft.py
@@ -458,7 +458,6 @@ class VariableSpeedCompressorTrainCommonShaft(CompressorTrainModel):
                 # iterate between rate with minimum power, and the previously found rate to return, to find the
                 # maximum rate that gives power consumption below maximum power
 
-                assert maximum_power is not None
                 result = self.fluid_factory.mass_rate_to_standard_rate(
                     find_root(
                         lower_bound=result_with_minimum_rate.stage_results[0].mass_rate_asv_corrected_kg_per_hour,


src/libecalc/domain/process/compressor/core/train/variable_speed_compressor_train_common_shaft.py
```diff
@@ -445,21 +456,26 @@ def _calculate_train_result_given_speed_at_stone_wall(
...
-                        upper_bound=rate_to_return,
-                        func=lambda x: self.evaluate_given_constraints(
-                            constraints=constraints.create_conditions_with_new_input(
-                                new_rate=self.fluid_factory.mass_rate_to_standard_rate(x),  # type: ignore[arg-type]
-                            )
-                        ).power_megawatt
-                        - self.data_transfer_object.maximum_power * (1 - POWER_CALCULATION_TOLERANCE),
-                        relative_convergence_tolerance=1e-3,
-                        maximum_number_of_iterations=20,
+                maximum_power = self.data_transfer_object.maximum_power
```
frodehk, line 459: I assume it is correct to return the rate directly if maximum_power is not set? (i.e. no limitations) - this check is due to mypy-feedback (maximum_power can be None, and None cannot be multiplied with a number)

olelod: Can we use a default value for maximum_power of float("inf") instead of None?

frodehk: I think the `if not self.data_transfer_object.maximum_power` is not sufficient for mypy (it checks for 0, None or other faults). I guess `if self.data_transfer_object.maximum_power is not None `will solve it (no need for the extra maximum power check (added in this PR)

PR #1090, opened 2025-09-03
perf: move validation of pressure values to before calculations

olelod:
- Validate that pressures, when we know they are pressures, are positive.
- Validate that rates, when we know they are rates, are non negative.
- Validate that the pressure ratio over a compressor is larger than 1, when we know that two pressures are on either side of a compressor (or pump, compressor system or pump system).


src/libecalc/presentation/yaml/domain/expression_time_series_pressure.py
```diff
@@ -14,6 +24,14 @@ def __init__(
         time_series_expression: TimeSeriesExpression,
     ):
         self._time_series_expression = time_series_expression
+        self._pressure_values = self._time_series_expression.get_evaluated_expressions()
+        self._validate()
+
+    def _validate(self):
+        """Validate that all pressure values are positive."""
+        for pressure in self._pressure_values:
+            if pressure is not None and pressure <= 0:
```
tj098895, line 33: so, `None` is ok?

src/libecalc/presentation/yaml/domain/expression_time_series_pressure.py
```diff
@@ -1,8 +1,18 @@
 from libecalc.common.time_utils import Periods
 from libecalc.domain.time_series_pressure import TimeSeriesPressure
+from libecalc.expression import Expression
 from libecalc.presentation.yaml.domain.time_series_expression import TimeSeriesExpression
 
 
+class InvalidPressureException(Exception):
```
tj098895, line 7: Is this caught further up, so that context such as (YAML) location is given?

src/libecalc/presentation/yaml/domain/expression_time_series_pressure.py
```diff
@@ -1,8 +1,18 @@
...
+    """Exception raised for invalid pressure values."""
+
+    def __init__(self, pressure: float, expressions: list[Expression]):
+        super().__init__(
+            f"All pressure values must be positive, got {pressure} in {[str(expression) for expression in expressions]}"
+        )
+
+
 class ExpressionTimeSeriesPressure(TimeSeriesPressure):
```
tj098895, line 16: An alternative to send an expression as a argument in the default constructor, is that the `TimeSeriesPressure` class instead has a `TimeSeriesPressure.from_expression(Expression) -> Self` that creates a `TimeSeriesPressure` instance. Because, after we have validated and instantiated the Timeseries, do we need to "worry about" it being initialized from an Expression? ... (honest question)

src/libecalc/presentation/yaml/domain/expression_time_series_pressure.py
tj098895: It can, but I think we should catch them now, in validation step.... Ie `,,,` represents to "empty" values. Whether the empty value is accepted and parsed as 0 is another thing, but I think we now do not accept it, and use need to fill in. Hence you are right, we can assume that the pressure values at this time has a valid number as a value ...

src/libecalc/presentation/yaml/domain/expression_time_series_flow_rate.py
```diff
@@ -27,8 +38,16 @@ def __init__(
...
         assert isinstance(consumption_rate_type, RateType)
         self._consumption_rate_type = consumption_rate_type
+        self._rate_values = self._get_stream_day_values()
+        self._validate()
 
-    def get_stream_day_values(self) -> list[float | None]:
+    def _validate(self):
+        """Validate that all flow rate values are positive."""
+        for rate in self._rate_values:
+            if rate is not None and rate < 0:
```
olelod, line 47: _get_stream_day_values returns list[float|None] so needs to handle None here as well, i think

frodehk (approved):

PR #1100, opened 2025-09-05
refactor: simplify validation exceptions

jsolaas:
DomainValidationException for a single error, without context.
ModelValidationException for one or more errors, with context.


frodehk (approved):
Nice work!!

PR #1120, opened 2025-09-12
1067 neqsim fix

tj098895:


src/ecalc_neqsim_wrapper/java_service.py
```diff
@@ -40,51 +40,183 @@ def _start_server(maximum_memory: str = "4G") -> JavaGateway:
...
+    Implemented by Neqsim Team
+    """
+
+    def __new__(cls, maximum_memory: str = "4G") -> "NeqsimJPypeService":
+        instance = super().__new__(cls)
+        return instance
+
+    @classmethod
+    def initialize(cls, maximum_memory: str = "4G") -> Self:
+        print("NeqsimJPypeService.initialize called")
```
tj098895, line 112: will remove

src/libecalc/common/services/feature_flag_service.py
```diff
@@ -0,0 +1,58 @@
+"""
+A feature flag singleton service available when needed
```
tj098895, line 2: Not relevant now, will remove/readd later, differently.

src/ecalc_neqsim_wrapper/java_service.py
```diff
@@ -40,51 +40,183 @@ def _start_server(maximum_memory: str = "4G") -> JavaGateway:
...
+    # _gateway: "JavaGateway"  # type: ignore # noqa: F821
+
+    def __new__(cls, maximum_memory: str = "4G") -> "NeqsimPy4JService":
+        instance = super().__new__(cls)
+        instance._gateway = _start_server(maximum_memory=maximum_memory)
+        _logger.info(
+            f"Started neqsim process with PID '{instance._gateway.java_process.pid}' "
+            f"on port '{instance._gateway.gateway_parameters.port}'"
+        )
+        return instance
```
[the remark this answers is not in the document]
tj098895: to avoid direct instantiation of the class with `NeqsimPy4JService()`, since the instance is already created when `init()` is being called. So would need to override `__new__` in any case. Also, it is easier and more transparent if I use class methods to get an instance of the service, instead of handling that through init, since we only need to send params when creating the instance (initially). There are many ways to get this singleton behaviour, I just thought this was closest to what I am used to from Java (where we set the constructor to `private` and use static variable and method to access the "class variable")

PR #1129, opened 2025-09-16
refactor: remove PathID

jsolaas:
Using a generated id instead, while also keeping track of yaml-context
based on that id.


src/libecalc/dto/fuel_type.py
```diff
@@ -15,22 +14,6 @@ class FuelType(EcalcBaseModel, Fuel):
     user_defined_category: FuelTypeUserDefinedCategoryType | None = Field(default=None, validate_default=True)
     emissions: list[Emission] = Field(default_factory=list)
 
-    @field_validator("user_defined_category", mode="before")
```
tj098895, line 18: Removed or moved? If removed, why?

src/libecalc/dto/utils/validators.py
```diff
@@ -58,13 +58,3 @@ def uppercase_user_defined_category(value):
     elif value is not None and is_temporal_model(value):
         return {timestep: category.upper() for timestep, category in value.items()}
     return value
-
-
-TModel = TypeVar("TModel")
-
-
-def validate_temporal_model(model: dict[Period, TModel]) -> dict[Period, TModel]:
```
tj098895, line 66: not in use?

PR #1154, opened 2025-09-29
refactor: don't use extend functionality for consumer results

jsolaas:


src/libecalc/core/result/results.py
```diff
@@ -50,107 +62,203 @@ def __init__(
...
         self.id = id
+        self.power_capacity_margin = power_capacity_margin
 
 
-class GeneratorSetResult(GenericComponentResult):
-    """The Generator set result component."""
+T = TypeVar("T", bound=TimeSeries)
+
+
+class AggregatedProperty(Generic[T]):
```
tj098895, line 75: Seems to be a good idea and start to get rid of a lot of verbose "copying of values" etc.

Can you add some tests to both demonstrate correctness and usage/documentation?

src/libecalc/core/result/results.py
```diff
@@ -50,107 +62,203 @@ def __init__(
...
+
+        if self._post_processor is not None:
+            return self._post_processor(res)
 
-    typ: Literal["genset"] = "genset"
+        return res
+
+
+class GenericComponentResult:
+    energy_usage = AggregatedProperty(TimeSeriesStreamDayRate, fill_value=0.0)
```
tj098895, line 151: so, for most of the "results", I got how it worked by "having a straw directly to the operational/raw data" - but I dont get how these data are set? Is the idea that they do not belong here, and should instead be added to the turbine etc - and are just being kept for now for backwardscompatibility - until the other is in place? Am I missing sth?

tests/libecalc/integration/snapshots/test_all_energy_usage_models/test_all_results/all_energy_usage_models_v3.json
```diff
@@ -71981,10 +71981,10 @@
                 },
                 "unit": "GWh",
                 "values": [
-                    0.0,
```
tj098895, line 71984: I guess there is no "correct answer" here, when aggregating/summing the totals; should the invalid data be included or not. Given that invalid data is never something that should end up in a correct model, it doesnt matter, but I guess it is (more) correct to include it :) Otherwise it can also give a "false" impression of what are the total numbers.

tj098895: Should be a heads up to inform users about a "semantic" change, as they are used to the opposite though

PR #1244, opened 2025-11-25
fix: compressor sampled evaluation input validation

frodehk:
Implement validation to ensure that the evaluation input for sampled compressor (qhull-based) models matches the required variables. The new validation raises a dedicated exception if any required input for the compressor's evaluation (rate, suction pressure, or discharge pressure) is missing.

Previously, the code did not check that the correct evaluation variables were always set. This likely hasn't caused issues because 'rate' is typically present in the user's input data (e.g., facility input). Several tests are added to confirm that the required inputs are enforced across 1D, 2D, and 3D models.

### DIFF
--- a/src/libecalc/domain/process/compressor/core/sampled/compressor_model_sampled.py
+++ b/src/libecalc/domain/process/compressor/core/sampled/compressor_model_sampled.py
@@ -14,6 +14,7 @@ from libecalc.common.list.list_utils import array_to_list
 from libecalc.common.logger import logger
 from libecalc.common.units import Unit
 from libecalc.domain.component_validation_error import (
+    CompressorModelSampledEvaluationInputValidationException,
     ProcessEqualLengthValidationException,
     ProcessMissingVariableValidationException,
     ProcessNegativeValuesValidationException,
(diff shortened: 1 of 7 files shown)


src/libecalc/domain/process/compressor/core/sampled/compressor_model_sampled.py
```diff
@@ -171,12 +172,31 @@ def get_max_standard_rate(
...
+        }
+        missing = [var for var in self.required_variables if input_vars[var] is None]
+        extra = [var for var in input_vars if var not in self.required_variables and input_vars[var] is not None]
+
+        if missing:
+            raise CompressorModelSampledEvaluationInputValidationException(
+                f"Missing required input(s) for sampled compressor energy usage model: {missing}"
+            )
+        if extra:
+            logger.warning(
```
kjbrak: If a table has multiple independent variables/dimensions and one dependent variable (fuel/power), how is it possible to interpolate between one independent variable and the dependent variable in such a dataset without knowing the other independent variable(s)?

Perhaps this should raise error also?

kjbrak (approved):

frodehk:
Closing this PR for now, as it has gone stale and this area is currently being refactored, so it’s unlikely to be merged in its current form.

PR #1470, opened 2026-04-10
refactor: pydantic Field in Annotated

jsolaas:


frodehk (approved):
Nice! I notice fields with alias and default_factory are not migrated — intentional scoping, or left for a follow-up?

jsolaas:
Dropped changing old code

PR #1481, opened 2026-04-13
refactor: invert shaft-compressor relationship

frodehk:
Previously each `Compressor` received a `Shaft` via its constructor and read speed through `self._shaft.get_speed()`. The ownership was implicit — multiple compressors shared a mutable reference.

Now `Shaft` explicitly owns its compressors. Speed is propagated to all registered units when `set_speed()` is called, enforcing the invariant (same speed on every compressor) in one place.

---

### Changes

**`shaft.py`**
- Added `register(unit)` — registers a compressor (or any object with `set_speed()`) on the shaft
- Added `get_speed_boundary()` — returns the overlapping speed range across all registered compressors
- `set_speed()` now propagates speed to all registered units via `_apply_speed()`

**`compressor.py`**
- Removed `shaft` from constructor
- Removed `shaft` property and `get_speed_boundary()`
- Added public `set_speed(speed)` — called by Shaft when speed changes
- `speed` property now reads from internal `_speed` instead of `self._shaft.get_speed()`

---

## Type of Work

### DIFF
--- a/src/libecalc/domain/process/entities/shaft/shaft.py
+++ b/src/libecalc/domain/process/entities/shaft/shaft.py
@@ -33,12 +33,12 @@ class Shaft(ABC):
     def set_speed(self, value: float) -> None:
         pass
 
-    def register(self, compressor) -> None:
+    def connect(self, compressor: Compressor) -> None:
         if compressor in self._compressors:
             raise ProgrammingError("Compressor is already registered on this shaft.")
         self._compressors.append(compressor)
 
-    def get_compressors(self) -> list:
+    def get_compressors(self) -> list[Compressor]:
         return list(self._compressors)
 
     def get_speed(self) -> float:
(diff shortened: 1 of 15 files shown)


src/libecalc/domain/process/entities/shaft/shaft.py
```diff
@@ -28,6 +33,14 @@ def get_id(self) -> ShaftId:
     def set_speed(self, value: float) -> None:
         pass
 
+    def register(self, compressor) -> None:
```
tj098895:
```suggestion
    def connect(self, compressor: Compressor) -> None:
```

?

src/libecalc/domain/process/entities/shaft/shaft.py
```diff
@@ -28,6 +33,14 @@ def get_id(self) -> ShaftId:
...
+        if compressor in self._compressors:
+            raise ProgrammingError("Compressor is already registered on this shaft.")
+        self._compressors.append(compressor)
+
+    def get_compressors(self) -> list:
```
tj098895: does this method need to exist? not currently used? Might wait, but I think the shaft in general should hold IDs to compressors it is connected to.

tj098895 (approved):

PR #1483, opened 2026-04-14
feat: add PressureRatioCompressor and CommonPressureRatioSolver

olelod:
What

  Introduces a ratio-based solving path for compressor trains that operate at a fixed pressure ratio — no shaft speed modelling. This is the domain foundation for the upcoming SIMPLIFIED_SERIAL YAML type (follow-up PR).

  Why

  The existing solver (CommonSpeedWithPressureControlSolver, formerly OutletPressureSolver) searches for a shaft speed to meet a target outlet pressure. Simplified compressor trains (using GENERIC_FROM_INPUT or GENERIC_FROM_DESIGN_POINT charts) don't model shaft speed — they distribute the total pressure ratio equally
  across stages. This PR adds domain support for that approach.

  Changes

  New domain entities:

   - PressureRatioCompressor — a ProcessUnit that evaluates at a given pressure ratio via propagate_stream_at_pressure_ratio(). Only accepts generic chart types.
   - CommonPressureRatioSolver — distributes equal pressure ratio (geometric mean) across all PressureRatioCompressor stages, applies minimum recirculation per stage (individual anti-surge). Exposes get_max_standard_rate() for stonewall-based feasibility.

  Solver hierarchy refactored:

   - New ProcessSystemSolver ABC with find_solution(pressure_constraint, inlet_stream) → Solution
   - OutletPressureSolver → renamed to CommonSpeedWithPressureControlSolver (speed-based path)
   - CommonPressureRatioSolver (ratio-based path, no speed search)

  FeasibilitySolver split:

   - RunnerBasedFeasibility — uses runner + stonewall limits (speed-based path)
   - RatioBasedFeasibility — uses per-stage stonewall limits from chart data directly

  Cross-validation guards:

   - CommonPressureRatioSolver rejects speed-based Compressor units
   - ProcessSystemRunner rejects PressureRatioCompressor units

  Both raise DomainValidationException at construction time.

  Tests

   - test_pressure_ratio_compressor.py — unit tests for the new process unit
   - test_common_pressure_ratio_solver.py — solver tests with 2-stage systems
   - Existing solver tests updated for the rename

### DIFF
--- a/src/libecalc/domain/process/entities/process_units/pressure_ratio_compressor.py
+++ b/src/libecalc/domain/process/entities/process_units/pressure_ratio_compressor.py
@@ -19,7 +19,7 @@ _ALLOWED_CHART_TYPES = (ChartType.GENERIC_FROM_INPUT, ChartType.GENERIC_FROM_DES
 class PressureRatioCompressor(PressureRatioUnit):
     """Compressor stage that operates at a given pressure ratio.
 
-    Provides the two methods CommonPressureRatioSolver needs:
+    Provides the two methods OutletPressureSolverRatio needs:
       - propagate_stream_at_pressure_ratio()
       - get_recirculation_range_at_pressure_ratio()
 
--- a/src/libecalc/domain/process/entities/process_units/pressure_ratio_compressor.py
+++ b/src/libecalc/domain/process/entities/process_units/pressure_ratio_compressor.py
@@ -1,0 +1,117 @@
+from libecalc.common.chart_type import ChartType
+from libecalc.domain.component_validation_error import (
+    ProcessChartTypeValidationException,
+)
+from libecalc.domain.process.compressor.core.train.utils.enthalpy_calculations import (
+    calculate_enthalpy_change_head_iteration,
+)
+from libecalc.domain.process.process_solver.boundary import Boundary
+from libecalc.domain.process.process_system.process_unit import ProcessUnit, ProcessUnitId
+from libecalc.domain.process.value_objects.chart.chart import ChartData
+from libecalc.domain.process.value_objects.chart.compressor import CompressorChart
+from libecalc.domain.process.value_objects.fluid_stream import FluidService, FluidStream
+from libecalc.domain.process.value_objects.fluid_stream.fluid import Fluid
+
+_ALLOWED_CHART_TYPES = (ChartType.GENERIC_FROM_INPUT, ChartType.GENERIC_FROM_DESIGN_POINT)
+
+
+class PressureRatioCompressor(ProcessUnit):
+    """Compressor stage that operates at a given pressure ratio.
+
+    Provides the two methods CommonPressureRatioSolver needs:
+      - propagate_stream_at_pressure_ratio()
+      - get_recirculation_range_at_pressure_ratio()
+
+    propagate_stream() raises NotImplementedError — a pressure ratio is always
+    required to evaluate this unit.
+    """
+
+    def __init__(
+        self,
+        process_unit_id: ProcessUnitId,
+        compressor_chart: ChartData,
+        fluid_service: FluidService,
+    ) -> None:
+        if compressor_chart.origin_of_chart_data not in _ALLOWED_CHART_TYPES:
+            raise ProcessChartTypeValidationException(
+                f"PressureRatioCompressor requires a generic chart, "
+                f"got {compressor_chart.origin_of_chart_data.value}."
+            )
+        self._id = process_unit_id
+        self._compressor_chart = CompressorChart(compressor_chart)
+        self._fluid_service = fluid_service
+
+    def get_id(self) -> ProcessUnitId:
+        return self._id
+
+    def propagate_stream(self, inlet_stream: FluidStream) -> FluidStream:
+        raise NotImplementedError(
+            f"PressureRatioCompressor '{self._id}' cannot be propagated with propagate_stream(). "
+            "Use propagate_stream_at_pressure_ratio() — a pressure ratio is always required."
+        )
+
+    @property
+    def compressor_chart(self) -> CompressorChart:
+        return self._compressor_chart
+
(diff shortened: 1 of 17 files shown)


tj098895:
Can this be rather solved by:

- Not use a shaft (and hence no shaft / speed strategy)
- Use a "PressureRatioStrategy" across compressors instead?

Can we then reuse the existing compressor etc and separate solving from simulation better?

But at the same time, this compressor is a "synthetic generic compressor", so might be relevant to keep that as a separate thing anyways somehow? Not sure how this aligns with reality and what is the use of something "unreal"?

olelod:
> Can this be rather solved by:
> 
> * Not use a shaft (and hence no shaft / speed strategy)
> * Use a "PressureRatioStrategy" across compressors instead?
> 
> Can we then reuse the existing compressor etc and separate solving from simulation better?
> 
> But at the same time, this compressor is a "synthetic generic compressor", so might be relevant to keep that as a separate thing anyways somehow? Not sure how this aligns with reality and what is the use of something "unreal"?

We of course could. But would not that mean having to deal with `shaft: Shaft | None`, `speed: float | None` etc?

The generic chart is fundamentally not a real compressor — it's a synthetic stand-in used when you don't have actual vendor curves. You give it an efficiency and it tells you how much work it does for a given inlet stream and pressure ratio. There's no shaft speed involved because the machine isn't modelled at that level of detail. We don't
   take the work performed and the rate put in, to impose a speed (interpolate it in the compressor chart) - and find three different speeds for three generic compressors in a train. We simply don't care about speed at all - only the work performed by the generic compressor.

The typical use case is early-phase studies or concept screening — you want a rough energy estimate for a compression stage before you have a vendor quote or a real performance curve. You're not pretending to know the machine; you're explicitly saying you don't.

src/libecalc/domain/process/process_solver/outlet_pressure_solver_ratio.py
```diff
@@ -0,0 +1,188 @@
...
+                        fluid_model=current_inlet.fluid_model,
+                        pressure_bara=current_inlet.pressure_bara,
+                        temperature_kelvin=current_inlet.temperature_kelvin,
+                        standard_rate_m3_per_day=current_inlet.standard_rate_sm3_per_day + recirculation_rate,
+                    )
+                else:
+                    inner_inlet = current_inlet
+
+                outlet_with_recirc = unit.propagate_stream_at_pressure_ratio(
+                    inner_inlet, pressure_ratio=pressure_ratio_per_stage
```
tj098895, line 102: can pressure ratio be a config that is being set instead, calculated by the compressor pressure ratio solver, in order to always use the `propagate_stream` method? I see now however, that the compressor is a bit different wrt. behaviour, so probably a good idea to separate it from a "normal compressor process impl"

olelod: Maybe PressureRatioCompressor should not inherit from ProcessUnit at all? Make it something separate?

tj098895: will such a system use ASV/recirc loop and other chokes, or not relevant?

olelod: There is of course individual recirculation to surge line (if the rate is too low), but this is just using min/max_rate_given_head on the compressor chart - no need for recirc ProcessUnits.

olelod: and upstream/downstream choke is not relevant. multiple streams and pressures should not be relevant for simplified trains.

src/libecalc/domain/process/entities/process_units/pressure_ratio_compressor.py
```diff
@@ -0,0 +1,117 @@
...
+            fluid_model=inlet_stream.fluid_model,
+            mass_rate_kg_per_h=max_actual_rate * density,
+        )
+        inlet_sm3_per_day = inlet_stream.standard_rate_sm3_per_day
+        return Boundary(
+            min=max(0.0, min_sm3_per_day - inlet_sm3_per_day),
+            max=max(0.0, max_sm3_per_day - inlet_sm3_per_day),
+        )
+
+    def propagate_stream_at_pressure_ratio(self, inlet_stream: FluidStream, pressure_ratio: float) -> FluidStream:
```
frodehk, line 89: Suggestion: Have a setter for pressure_ratio, and use propagate_stream() to fulfill the ProcessUnit contract. Other process units are already doing this, e.g. Choke (set_pressure_change).

Alternatively, split the ProcessUnit interface? e.g. RatioBasedUnit

src/libecalc/domain/process/process_solver/common_pressure_ratio_solver.py
```diff
@@ -0,0 +1,188 @@
...
+    Solution,
+    SolverFailureStatus,
+    TargetNotAchievableEvent,
+)
+from libecalc.domain.process.process_solver.solvers.recirculation_solver import RecirculationConfiguration
+from libecalc.domain.process.process_system.process_system import ProcessSystem
+from libecalc.domain.process.value_objects.fluid_stream import FluidService, FluidStream
+
+
+class CommonPressureRatioSolver(ProcessSystemSolver):
```
frodehk: OutletPressureSolverRatio?

src/libecalc/domain/process/process_solver/common_speed_with_pressure_control_solver.py
```diff
@@ -26,7 +27,7 @@
 from libecalc.domain.process.value_objects.fluid_stream import FluidStream
 
 
-class OutletPressureSolver:
+class CommonSpeedWithPressureControlSolver(ProcessSystemSolver):
```
frodehk: OutletPressureSolverSpeed?

olelod:
replaced by https://github.com/equinor/ecalc/pull/1527

PR #1496, opened 2026-04-22
refactor: separate configuration handlers from process units

jsolaas:
Shaft and RecirculationLoops are now implemented as
ConfigurationHandlers. Since the only relevant consequence of a shaft
and recirculation loop is the constraint that either the compressors
should have a common speed or the mixer and splitter should have a
common rate, we can implement RecirculationLoop and Shaft as aggregates
for configurations, i.e. the RecirculationLoop and Shaft classes make
  sure the correct configuration is set for each process unit. The
  configuration handlers is not part of the process pipeline itself.


src/libecalc/domain/process/process_solver/process_system_runner.py
```diff
@@ -1,99 +1,58 @@
...
-            unit = self._shaft
+        self._apply_config_for_unit(configuration_handler=unit, configuration=configuration)
 
-        if unit is None:
-            unit = self._units.get(unit_id)  # type: ignore[arg-type]
-
-        if unit is None:
-            raise ValueError(f"Unit with id '{unit_id}' does not exist.")
-        return unit
+    def _get_unit(self, unit_id: SimulationUnitId) -> ConfigurationHandler:
```
tj098895:
```suggestion
    def _get_simulation_unit(self, simulation_unit_id: SimulationUnitId) -> ConfigurationHandler:
```

to make clear when we use process unit and when we use simulation unit?

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -99,62 +102,83 @@ def get_excess_rate(self, inlet_stream: FluidStream) -> float:
...
+                process_unit_id=create_process_unit_id(),
             )
-            for recirculation_loop_id, compressor in zip(
-                recirculation_loop_ids, self._process_system.get_process_units(), strict=True
+            splitter = DirectSplitter(
+                process_unit_id=create_process_unit_id(),
+            )
+            recirculation_loop = RecirculationLoop(
+                configuration_handler_id=create_simulation_unit_id(),
+                inner_process=self._process_pipeline,
```
tj098895: inner_process=compressor?

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -99,62 +102,83 @@ def get_excess_rate(self, inlet_stream: FluidStream) -> float:
 class CompressorTrainBuilder:
     def __init__(self, compressors: Sequence[Compressor], fluid_service: FluidService):
         self._fluid_service = fluid_service
-        self._process_system = SerialProcessSystem(
-            process_system_id=create_process_system_id(), propagators=compressors
-        )
+        self._process_pipeline: Sequence[ProcessUnit] = compressors
```
tj098895, line 104: is this correct? rather assign to compressors variable that we use as source to build the pipeline? maybe there is sth i dont understand.

tj098895 (approved):
very good, bits are falling into place :) some bugs though as far as I see.

PR #1508, opened 2026-04-27
refactor: update boundary calculations for Common ASV strategies

frodehk:
Compute the recirculation boundary on the compressor's actual inlet stream (after `TemperatureSetter` +  `LiquidRemover`) instead of the train-inlet stream, matching what the Individual ASV strategies already do.

#### Changes

anti_surge/common_asv.py: propagate to first compressor inlet before calling `get_recirculation_range`.
pressure_control/common_asv.py: same.
No behavioural change expected for converged cases; eliminates fallback searches in `RecirculationSolver` for `COMMON_ASV`.

Refs: equinor/ecalc-internal#1721

## Type of Work

- [x] Patch: X.Y.**Z+1**. **NEGLIGIBLE** visible changes, does not change input or output - OR changes behaviour. Use chore:, refactor: etc
- [ ] Minor: X.**Y+1**.Z. Minor changes, might ADD new input (YAML), or other **backwards-compatible** changes. Use feat:, fix:
- [ ] Major: **X+1**.Y.Z. Major and most likely **BREAKING** changes, wo. backwards compatibility, or removing temporary backwards compatibility functionality. Use ! or BREAKING:.

See here (internal): https://github.com/equinor/ecalc-internal/discussions/1044

## Have you remembered and considered?

- [ ] IF FEAT: I have remembered to update documentation
- [ ] IF FIX OR FEAT: I have remembered to update manual changelog (`docs/drafts/next.draft.md`)
- [ ] IF BREAKING: I have remembered to update migration guide (`docs/docs/migration_guides/`)
- [ ] IF BREAKING: I have committed with `BREAKING:` in footer or `!` in header
- [ ] I have added tests (if not, comment why)
- [ ] I have used conventional commits syntax (if you squash, make sure that conventional commit is used)
- [ ] I have included the Github issue nr in the footer!

PR #1515, opened 2026-04-29
fix: reset recirculation before computing pressure control boundary

kjbrak:
## Problem

`CommonASVPressureControlStrategy.apply()` computes the recirculation boundary via `run(inlet_stream, to_id=unit)` which passes through the `DirectMixer`. When anti-surge had previously set a non-zero recirculation rate, the mixer inflated the unit's inlet rate, shrinking the available recirculation range.

This is a latent bug introduced by #1508 (which changed boundary computation to use `run(to_id=...)` instead of the raw inlet stream). It's hard to trigger with compressors because their anti-surge recirculation is typically small relative to the chart range, but it becomes visible when the anti-surge fraction is large — e.g., when inlet rate is well below the chart minimum.

## Fix

Reset recirculation to zero before the boundary computation, so the range is always based on the base inlet rate. This matches the pattern anti-surge already uses (`reset()` is called before `apply()` in `OutletPressureSolver`).

## Testing

- Added regression test: `test_pressure_control_boundary_not_affected_by_residual_anti_surge_recirculation`
- Uses a compressor with chart min 50 m³/h and a very low inlet rate (5,000 sm³/day ≈ 7 m³/h), forcing large anti-surge recirculation
- Verified the test **fails without the fix** (solver reports achievable=60.4 vs target=60.0) and **passes with it**
- All 46 process solver tests pass

### DIFF
--- a/src/libecalc/domain/process/process_solver/pressure_control/common_asv.py
+++ b/src/libecalc/domain/process/process_solver/pressure_control/common_asv.py
@@ -51,6 +51,15 @@ class CommonASVPressureControlStrategy(PressureControlStrategy):
             return self._simulator.run(inlet_stream=inlet_stream)
 
         # Check feasibility: can we reach target pressure at maximum recirculation?
+        # Reset recirculation before computing the boundary — the mixer may carry
+        # residual state from anti-surge, which inflates the inlet rate and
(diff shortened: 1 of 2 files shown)


frodehk (approved):
Nice!

PR #1519, opened 2026-04-29
feat: add EqualRatioOutletPressureCalculator

olelod:
What

Adds an analytic outlet-pressure calculator for compressor trains that operate at a fixed pressure ratio per stage — no shaft speed search. Same idea as the legacy `CompressorTrainSimplified`, repackaged as a `Solution`-returning calculator that plugs into the `ProcessRunner` / configuration-handler architecture.

Why

The existing `OutletPressureSolver` searches for a shaft speed to meet a target outlet pressure. Simplified compressor trains using `GENERIC_FROM_INPUT` / `GENERIC_FROM_DESIGN_POINT` charts don't model shaft speed — they distribute the total pressure ratio equally across stages. This PR adds a calculator for that path, sitting alongside the speed-based solver.

Changes

New domain entity:

 - `EqualRatioOutletPressureCalculator` — distributes equal pressure ratio (geometric mean) across all `Compressor` units in the pipeline, applies minimum recirculation per stage (individual anti-surge), and returns a `Solution[Sequence[Configuration]]` of `RecirculationConfiguration` per stage. Exposes `get_max_standard_rate()` for stonewall-based feasibility (per-stage max → bottleneck `min`).
 - Rejects `Choke` units and non-generic compressor charts at construction time.

Refactor:

 - Moved `get_max_standard_rate` from `FeasibilitySolver` to `ProcessRunner`, so both the speed-based and ratio-based paths can share it.

Tests

 - `test_equal_ratio_outlet_pressure_calculator.py` — unit tests covering equal-ratio split, individual anti-surge, max-rate bottleneck, choke/chart-type rejection.

### DIFF
--- a/src/libecalc/domain/process/process_pipeline/process_unit.py
+++ b/src/libecalc/domain/process/process_pipeline/process_unit.py
@@ -1,10 +1,11 @@
 import abc
-from typing import NewType, Self
+from typing import NewType, Protocol, Self, runtime_checkable
 from uuid import UUID
 
 from libecalc.common.ddd.entity import Entity
 from libecalc.common.utils.ecalc_uuid import ecalc_id_generator
 from libecalc.domain.process.process_pipeline.stream_propagator import StreamPropagator
+from libecalc.domain.process.value_objects.fluid_stream import FluidStream
 
 ProcessUnitId = NewType("ProcessUnitId", UUID)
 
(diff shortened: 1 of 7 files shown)


olelod:
replaced by https://github.com/equinor/ecalc/pull/1527

PR #1540, opened 2026-05-05
fix: harden compressor PH flashes

kjbrak:
## Type of Work

## Summary
Note: BREAKING!

This PR hardens compressor PH flash handling after debugging a production compressor model crash.

The work happened in three steps:

1. **Fix the original compressor PH crash**
   - The original failure came from a compressor outlet PH flash where NeqSim returned an invalid outlet state near the phase envelope.
   - That invalid state contained bad thermodynamic properties, including non-finite `kappa`.
   - eCalc then used those properties in the Campbell pressure iteration, which produced a non-finite outlet pressure and eventually crashed in NeqSim.
   - The fix is to seed compressor PH flashes from the compressor inlet temperature instead of starting from the cached standard-condition reference state.

2. **Add validation so bad PH results cannot silently propagate**
   - PH flash results are now validated before they are cached or used.
   - Validation checks finite/positive pressure, temperature, density, `z`, `kappa`, standard density, finite vapor fraction and enthalpy, and that the resulting enthalpy matches the PH target within tolerance.
   - Compressor outlet calculations now also reject non-finite Campbell pressure guesses before they are sent back into NeqSim.

3. **Handle additional invalid/non-physical states exposed by validation**
   - After validation was added, the full model exposed another failure in a max-speed/capacity probe: the chart point was already invalid/outside capacity, but the code still required a physical outlet PH state for the extrapolated probe.
   - That PH target was not physically reachable, causing NeqSim to drive temperature toward near-zero K and fail. The code now preserves the original invalid chart/capacity result instead of crashing on the outlet flash for an already-invalid chart point.
   - Validation also exposed an existing simplified-train test case where extreme pressure ratios produced an invalid near-zero-temperature PH result. The simplified max-rate helper now avoids using those invalid PH properties once the required head is already above the maximum-speed chart head, preserving the existing chart-limited scalar result.

## Changes

- Add optional `temperature_guess_kelvin` to `FluidService.flash_ph(...)`.
- Seed compressor PH flashes from inlet stream temperature.
- Validate NeqSim PH flash results before cache insertion/use.
- Add `CompressorOutletCalculationError` for compressor outlet thermodynamic failures.
- Reject invalid/non-finite compressor PH results and Campbell pressure guesses.
- In legacy compressor evaluation, skip failed outlet PH only when the chart point is already invalid; valid chart points still raise.
- Preserve simplified max-rate behavior for above-chart-head estimates without relying on invalid NeqSim properties.

## Testing

Added and updated tests for seeded PH flash cache behavior, compressor outlet PH validation, invalid-chart outlet failure handling, and simplified max-rate behavior for invalid high-head PH states.

All tests pass.

## Verification

- All UAT models run successfully with no validation triggers — the new guards are targeted and do not affect existing production models.
- Unified Java bridge error handling: PH/TP flash handlers now catch both Py4J and JPype exceptions via a shared `JAVA_ERRORS` tuple.

## Githus projects issue
refs.:
https://github.com/equinor/ecalc-internal/issues/1765

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -396,38 +396,6 @@
             {
                 "code": "reportAttributeAccessIssue",
                 "range": {
-                    "startColumn": 36,
-                    "endColumn": 41,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportAttributeAccessIssue",
-                "range": {
-                    "startColumn": 75,
-                    "endColumn": 80,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportAttributeAccessIssue",
-                "range": {
-                    "startColumn": 62,
-                    "endColumn": 67,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportAttributeAccessIssue",
-                "range": {
-                    "startColumn": 62,
-                    "endColumn": 67,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportAttributeAccessIssue",
-                "range": {
                     "startColumn": 35,
                     "endColumn": 42,
                     "lineCount": 1
--- a/src/ecalc_neqsim_wrapper/fluid_service.py
+++ b/src/ecalc_neqsim_wrapper/fluid_service.py
@@ -13,11 +13,13 @@ import logging
 from typing import ClassVar
 
 from ecalc_neqsim_wrapper.cache_service import CacheConfig, CacheName, CacheService, LRUCache
+from ecalc_neqsim_wrapper.exceptions import NeqsimFlashCalculationError
 from ecalc_neqsim_wrapper.thermo import NeqsimFluid
 from libecalc.process.fluid_stream.constants import ThermodynamicConstants
 from libecalc.process.fluid_stream.fluid import Fluid
 from libecalc.process.fluid_stream.fluid_model import EoSModel, FluidComposition, FluidModel
 from libecalc.process.fluid_stream.fluid_properties import FluidProperties
+from libecalc.process.fluid_stream.fluid_property_validation import validate_ph_flash_result
 from libecalc.process.fluid_stream.fluid_service import FluidService
 from libecalc.process.fluid_stream.fluid_stream import FluidStream
 
@@ -56,7 +58,7 @@ class NeqSimFluidService(FluidService):
     Usage:
         service = NeqSimFluidService.instance()
         props = service.flash_pt(fluid_model, pressure_bara, temperature_kelvin)
-        props = service.flash_ph(fluid_model, pressure_bara, target_enthalpy)
+        props = service.flash_ph(fluid_model, pressure_bara, target_enthalpy_joule_per_kg)
     """
 
     _instance: ClassVar[NeqSimFluidService | None] = None
@@ -200,7 +202,7 @@ class NeqSimFluidService(FluidService):
         )
 
     def _make_ph_cache_key(
-        self, composition_key: tuple, eos_model: EoSModel, pressure: float, target_enthalpy: float
+        self, composition_key: tuple, eos_model: EoSModel, pressure: float, target_enthalpy_joule_per_kg: float
     ) -> tuple:
         """Create cache key for PH flash with proper rounding.
 
@@ -209,13 +211,15 @@ class NeqSimFluidService(FluidService):
             ("PH", composition_key, eos_model, rounded_pressure, rounded_enthalpy)
 
             This ensures unique cache entries for each distinct thermodynamic state.
+            The temperature guess is intentionally not part of this key; it is
+            a convergence aid, not part of the PH target state.
         """
         return (
             "PH",
             composition_key,
             eos_model,
             round(pressure, _PRESSURE_DECIMALS),
-            round(target_enthalpy, _ENTHALPY_DECIMALS),
+            round(target_enthalpy_joule_per_kg, _ENTHALPY_DECIMALS),
         )
 
     def _get_standard_density(self, fluid_model: FluidModel) -> float:
@@ -310,12 +314,13 @@ class NeqSimFluidService(FluidService):
         self,
         fluid_model: FluidModel,
         pressure_bara: float,
-        target_enthalpy: float,
+        target_enthalpy_joule_per_kg: float,
+        temperature_guess_kelvin: float | None = None,
     ) -> FluidProperties:
         """PH flash to target pressure and enthalpy.
 
-        Note: The target_enthalpy is reference-state dependent. NeqSim uses an arbitrary
-        enthalpy reference state, so the caller must ensure that target_enthalpy was
+        Note: The target_enthalpy_joule_per_kg is reference-state dependent. NeqSim uses an arbitrary
+        enthalpy reference state, so the caller must ensure that target_enthalpy_joule_per_kg was
         computed from enthalpy values obtained from the same EoS/fluid model session.
         Mixing enthalpy values from different thermodynamic packages or sessions may
         produce incorrect results.
@@ -323,14 +328,24 @@ class NeqSimFluidService(FluidService):
         Args:
             fluid_model: The fluid model (composition + EoS)
             pressure_bara: Target pressure in bara
-            target_enthalpy: Target specific enthalpy in J/kg (must be from same EoS session)
-
+            target_enthalpy_joule_per_kg: Target specific enthalpy in J/kg (must be from same EoS session)
+            temperature_guess_kelvin: Optional initial temperature for PH. When provided, the cached reference fluid is
+                first TP-flashed at the target pressure and this temperature before the same PH target is solved. This
+                is a convergence aid for callers with a physically relevant temperature estimate
+                (or a T closer to target T than T_standard at least); it will not change
+                a correctly converged PH result and is intentionally not part of the PH cache key.
+                note: this can avoid some bad initial states for the PH flash in some edge cases
         Returns:
             FluidProperties at the specified conditions.
         """
         composition = fluid_model.composition.normalized()
         composition_key = _make_composition_key(composition)
-        cache_key = self._make_ph_cache_key(composition_key, fluid_model.eos_model, pressure_bara, target_enthalpy)
+        cache_key = self._make_ph_cache_key(
+            composition_key,
+            fluid_model.eos_model,
+            pressure_bara,
+            target_enthalpy_joule_per_kg,
+        )
 
         # Check cache first
         cached = self._flash_cache.get(cache_key)
@@ -338,12 +353,28 @@ class NeqSimFluidService(FluidService):
             return cached
 
         ref = self._get_reference_fluid(fluid_model)
-        flashed = ref.set_new_pressure_and_enthalpy(
-            new_pressure=pressure_bara,
-            new_enthalpy_joule_per_kg=target_enthalpy,
-        )
+        if temperature_guess_kelvin is not None:
+            seeded_fluid = ref.set_new_pressure_and_temperature(
+                new_pressure_bara=pressure_bara,
+                new_temperature_kelvin=temperature_guess_kelvin,
+            )
+            flashed = seeded_fluid.set_new_pressure_and_enthalpy(
+                new_pressure=pressure_bara,
+                new_enthalpy_joule_per_kg=target_enthalpy_joule_per_kg,
+            )
+        else:
+            flashed = ref.set_new_pressure_and_enthalpy(
+                new_pressure=pressure_bara,
+                new_enthalpy_joule_per_kg=target_enthalpy_joule_per_kg,
+            )
 
         result = self._extract_properties(flashed, fluid_model)
+        validate_ph_flash_result(
+            result,
+            target_enthalpy_joule_per_kg,
+            "NeqSimFluidService.flash_ph",
+            error_factory=NeqsimFlashCalculationError,
+        )
         self._flash_cache.put(cache_key, result)
         return result
 
--- a/src/libecalc/domain/process/compressor/core/train/simplified_train/simplified_train.py
+++ b/src/libecalc/domain/process/compressor/core/train/simplified_train/simplified_train.py
@@ -3,12 +3,14 @@ import numpy as np
 from libecalc.common.errors.exceptions import IllegalStateException
 from libecalc.common.logger import logger
 from libecalc.common.units import UnitConstants
+from libecalc.domain.process.compressor.core.exceptions import CompressorThermodynamicCalculationError
 from libecalc.domain.process.compressor.core.results import (
     CompressorTrainResultSingleTimeStep,
 )
 from libecalc.domain.process.compressor.core.train.base import CompressorTrainModel
 from libecalc.domain.process.compressor.core.train.stage import CompressorTrainStage
 from libecalc.domain.process.compressor.core.train.train_evaluation_input import CompressorTrainEvaluationInput
+from libecalc.domain.process.compressor.core.train.utils.common import flash_ph_for_compressor_calculation
 from libecalc.domain.process.compressor.core.train.utils.enthalpy_calculations import (
     calculate_polytropic_head_campbell,
 )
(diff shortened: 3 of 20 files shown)


src/ecalc_neqsim_wrapper/fluid_service.py
```diff
@@ -36,6 +40,8 @@
 # Standard conditions for reference fluids and standard density calculations
 _STANDARD_TEMPERATURE_KELVIN = 288.15
 _STANDARD_PRESSURE_BARA = 1.01325
+_PH_ENTHALPY_REL_TOLERANCE = 1e-4
```
tj098895: What are the rationales behind these tolerances? From literature? Testing? Would it be important to document this for some expert users?

src/ecalc_neqsim_wrapper/fluid_service.py
```diff
@@ -218,6 +226,48 @@ def _make_ph_cache_key(
...
+            raise IllegalStateException(f"{context}: expected finite positive {name}, got {value}.")
+
+    @classmethod
+    def _validate_ph_result(
+        cls,
+        properties: FluidProperties,
+        target_enthalpy: float,
+        context: str,
+    ) -> None:
+        cls._require_positive_finite(properties.pressure_bara, "pressure_bara", context)
```
tj098895: could these checks be moved to `__post__init__` for `FluidProperties` instead? Possibly add custom or Pydantic validation if that is what we want?

src/ecalc_neqsim_wrapper/fluid_service.py
```diff
@@ -218,6 +226,48 @@ def _make_ph_cache_key(
...
+    @staticmethod
+    def _require_positive_finite(value: float, name: str, context: str) -> None:
+        if not math.isfinite(value) or value <= 0:
+            raise IllegalStateException(f"{context}: expected finite positive {name}, got {value}.")
+
+    @classmethod
+    def _validate_ph_result(
+        cls,
+        properties: FluidProperties,
+        target_enthalpy: float,
```
tj098895:
```suggestion
        target_enthalpy_joule_per_kg: float,
```

to be consistent?

src/ecalc_neqsim_wrapper/fluid_service.py
```diff
@@ -311,6 +361,7 @@ def flash_ph(
         fluid_model: FluidModel,
         pressure_bara: float,
         target_enthalpy: float,
+        temperature_guess_kelvin: float | None = None,
```
tj098895, line 318: is it the user that should decide this? or us? not sure if I understand when it should happen, and when it should not happen?

tj098895: will this have (big) consequences on results from neqsim?

src/ecalc_neqsim_wrapper/fluid_service.py
```diff
@@ -324,26 +375,59 @@ def flash_ph(
...
+                    f"pressure_bara={pressure_bara}, target_enthalpy_joule_per_kg={target_enthalpy}, "
+                    f"temperature_guess_kelvin={temperature_guess_kelvin}, eos_model={fluid_model.eos_model}."
+                ) from error
+        else:
+            try:
+                flashed = ref.set_new_pressure_and_enthalpy(
+                    new_pressure=pressure_bara,
+                    new_enthalpy_joule_per_kg=target_enthalpy,
+                )
+            except (Py4JJavaError, Py4JError) as error:
```
tj098895: nice to catch these errors. Are those the only errors we can get, all neqsim errors wrapped in either one of them?

however, here we are slightly violating the rule about this code being completely ignorant about py4j, neqsim, jpype etc. Could we just change to catch a generic exception, or can we request that the "reference fluid implementation" HAS TO throw an eCalc exception instead? If we change implementation etc, these exceptions will suddenly not exist any more.

src/libecalc/domain/process/compressor/core/train/simplified_train/simplified_train.py
```diff
@@ -279,7 +280,34 @@ def calculate_maximum_rate_for_stage(
             enthalpy_change_joule_per_kg = polytropic_head / polytropic_efficiency
 
             target_enthalpy = inlet_stream.enthalpy_joule_per_kg + enthalpy_change_joule_per_kg
-            props = fluid_service.flash_ph(inlet_stream.fluid_model, outlet_pressure, target_enthalpy)
+            try:
+                props = fluid_service.flash_ph(
+                    inlet_stream.fluid_model,
```
tj098895: kw_args only to make sure we get it right?

src/libecalc/domain/process/compressor/core/train/simplified_train/simplified_train.py
```diff
@@ -279,7 +280,34 @@ def calculate_maximum_rate_for_stage(
...
+                    maximum_head_on_maximum_speed_curve,
+                    exc_info=True,
+                )
+                maximum_actual_volume_rate = float(
+                    maximum_rate_function(
+                        heads=polytropic_head,  # type: ignore[arg-type]
+                        extrapolate_heads_below_minimum=False,
+                    )
+                )
+                break
```
tj098895, line 321: a bit difficult to see what will happen here imo

src/libecalc/domain/process/compressor/core/train/utils/common.py
```diff
@@ -14,9 +18,78 @@
 RECIRCULATION_BOUNDARY_TOLERANCE = 1e-6
 # -----------------------------------------------------------------------------
 # Impossible pressure sanity cap.  Protects the PH flash when the first
-# Campbell estimate is evaluated for an artificial max‑speed head in edge cases.
+# Campbell estimate is evaluated for an artificial max-speed head in edge cases.
 # -----------------------------------------------------------------------------
 MAX_FIRST_GUESS_BAR = 2_000.0  # [bara]
+COMPRESSOR_PH_ENTHALPY_REL_TOLERANCE = 1e-4
```
tj098895: relevant to reuse both the constants and the util to test for positive+finite here instead of duplicating to fluid_service? or do you not want fluid service to dep on common?

src/libecalc/domain/process/compressor/core/train/utils/common.py
```diff
@@ -1,8 +1,12 @@
+import math
+
+from libecalc.common.errors.exceptions import IllegalStateException
```
tj098895: I didnt know we had this exception. It is a bit misleadign to me, coming from Java. Then it means that the class/method has been called in an inproper way. Could we change to so something more domain related? Which state are we talking about? fluid/gas state?

src/libecalc/domain/process/compressor/core/train/utils/common.py
```diff
@@ -14,9 +18,78 @@
...
+) -> FluidProperties:
+    try:
+        return fluid_service.flash_ph(
+            inlet_stream.fluid_model,
+            outlet_pressure_bara,
+            target_enthalpy,
+            temperature_guess_kelvin=inlet_stream.temperature_kelvin,
+        )
+    except IllegalStateException as error:
+        raise CompressorOutletCalculationError(f"{context}: PH flash failed.") from error
```
tj098895: I like this way of handling exceptions, more context and better fit msg/exception for more tailored handling

src/ecalc_neqsim_wrapper/fluid_service.py
kjbrak: We decide to use it when we have a good guess, when we know a temperature closer to the target than standard temp.

kjbrak: There is a description of it in the docstring

src/ecalc_neqsim_wrapper/fluid_service.py
kjbrak: Good comment. I will check.

src/libecalc/domain/process/compressor/core/train/utils/common.py
kjbrak: Good feedback, i will sort this out.

src/libecalc/domain/process/compressor/core/train/utils/common.py
kjbrak: I saw it had been used many places and it seemed to fit ok, but i agree, will take a look.

tests/ecalc_neqsim_wrapper/unit_tests/test_seeded_ph_flash.py
```diff
@@ -0,0 +1,75 @@
+from __future__ import annotations
+
+from ecalc_neqsim_wrapper.cache_service import CacheService
+from ecalc_neqsim_wrapper.fluid_service import NeqSimFluidService
+from libecalc.process.fluid_stream.fluid_model import EoSModel, FluidComposition, FluidModel
+from libecalc.process.fluid_stream.fluid_properties import FluidProperties
+
+
+class FakeNeqsimFluid:
```
tj098895: FakeFluid?

tj098895:
just curious, since you didnt check the "breaking" checkbox, but indicating with **!** that it is breaking, what is breaking?

src/ecalc_neqsim_wrapper/fluid_service.py
kjbrak: Centralized the PH flash validation tolerances and documented that these are defensive backend-result checks, not process-model tolerances.

src/ecalc_neqsim_wrapper/fluid_service.py
kjbrak: Kept this at the PH flash operation boundary since validation needs the target enthalpy and operation context, which FluidProperties does not have.

src/ecalc_neqsim_wrapper/fluid_service.py
kjbrak: Clarified the docstring: the temperature guess is only a convergence aid for the same PH target state and is intentionally not part of the cache key.

src/ecalc_neqsim_wrapper/fluid_service.py
kjbrak: Moved backend-specific flash failures behind the NeqSim wrapper boundary and expose fluid-service flash errors to domain code instead.

src/libecalc/domain/process/compressor/core/train/simplified_train/simplified_train.py
kjbrak: Refactored this path to make the chart-limited fallback explicit and only use it when the PH flash fails above chart maximum head.

src/libecalc/domain/process/compressor/core/train/utils/common.py
kjbrak: Moved the shared PH validation and tolerances out of compressor/common duplication so fluid service and compressor code reuse the same validation logic.

src/libecalc/domain/process/compressor/core/train/utils/common.py
kjbrak: Replaced this with compressor-domain thermodynamic errors carrying operation context and physical inputs.

src/libecalc/domain/process/compressor/core/train/utils/common.py
```diff
@@ -14,9 +18,78 @@
...
+class CompressorOutletCalculationError(IllegalStateException):
+    """Raised when compressor outlet thermodynamics cannot produce a usable state."""
+
+
+def _require_positive_finite(value: float, name: str, context: str) -> None:
+    if not math.isfinite(value) or value <= 0:
+        raise CompressorOutletCalculationError(f"{context}: expected finite positive {name}, got {value}.")
+
+
+def _validate_compressor_ph_flash_result(
```
olelod: Doesn't this and `_validate_ph_result` in `fluid_service.py` basically validate the same thing? The only extra here is the vapor fraction inside [0,1]. Move that to `_validate_ph_result` and drop this validation?

src/ecalc_neqsim_wrapper/fluid_service.py
```diff
@@ -324,26 +375,59 @@ def flash_ph(
...
+            try:
+                seeded_fluid = ref.set_new_pressure_and_temperature(
+                    new_pressure_bara=pressure_bara,
+                    new_temperature_kelvin=temperature_guess_kelvin,
+                )
+                flashed = seeded_fluid.set_new_pressure_and_enthalpy(
+                    new_pressure=pressure_bara,
+                    new_enthalpy_joule_per_kg=target_enthalpy,
+                )
+            except (Py4JJavaError, Py4JError) as error:
```
olelod: Did we drop JPype? Can errors there still leak through?

src/libecalc/domain/process/compressor/core/train/utils/common.py
```diff
@@ -14,9 +18,78 @@
 RECIRCULATION_BOUNDARY_TOLERANCE = 1e-6
 # -----------------------------------------------------------------------------
 # Impossible pressure sanity cap.  Protects the PH flash when the first
-# Campbell estimate is evaluated for an artificial max‑speed head in edge cases.
+# Campbell estimate is evaluated for an artificial max-speed head in edge cases.
 # -----------------------------------------------------------------------------
 MAX_FIRST_GUESS_BAR = 2_000.0  # [bara]
```
olelod, line 26: Have you checked if this one is still needed now?

src/libecalc/process/fluid_stream/fluid_service.py
```diff
@@ -66,6 +67,12 @@ def flash_ph(
             fluid_model: The fluid model (composition + EoS)
             pressure_bara: Target pressure in bara
             target_enthalpy: Target specific enthalpy in J/kg (must be from same EoS session)
+            temperature_guess_kelvin: Optional initial temperature for the PH flash.
+                Implementations may use this to precondition the underlying thermodynamic
+                system at (pressure_bara, temperature_guess_kelvin) before solving PH.
+                Usfeul if knowing a temperature closer to final state than the reference fluid
```
olelod: Useful

src/ecalc_neqsim_wrapper/fluid_service.py
tj098895: specific implementation errors should be handled in the specific implementation (wrapper). JPype is atm inactive, but it is still there and can be configured to be used.

src/libecalc/domain/process/compressor/core/train/utils/common.py
kjbrak: @olelod : I did come changes yesterday before leaving the office and forgot to push, so the state this morning has been stale. New version pushed now. This issue has been resolved as the validation logic has been consolidated to `src/libecalc/process/fluid_stream/fluid_property_validation.py`

src/libecalc/domain/process/compressor/core/train/utils/common.py
kjbrak: I would think its still needed? Is there anything we have done that would make the first gas in the campbell iterations not risk being very high and crash eos liek we saw earlier? I dont think this PR changes anything regarding that.

src/ecalc_neqsim_wrapper/fluid_service.py
kjbrak: Good catch. The catch blocks in thermo.py only caught Py4JError/Py4JJavaError. JPype JException was not handled, so Java exceptions would leak through when using the JPype backend. Fixed in e8dd37f3: added a unified JAVA_ERRORS tuple in exceptions.py covering both Py4JError and JException. PH/TP flash handlers now catch both backends. Added 5 tests.

olelod (approved):
Should be good to go now.

Either make it a fix! or remove the Note Breaking (i think removing the note about breaking is most correct - since there is no change in input/output, and only edge cases are affected and treated more correctly)

.basedpyright/baseline.json
```diff
@@ -393,38 +393,6 @@
                     "lineCount": 1
                 }
             },
-            {
```
olelod, line 396: Should this be here?

kjbrak: this file was updated automatically by pyright when i fixed the type error.. hmm

kjbrak: should probably not edit it manually?

kjbrak: Looking at it, what is the problem here really, i dont see it.

jsolaas (approved):

PR #1543, opened 2026-05-06
feat(libecalc): handle unachievable outlet fluid states

olelod:
## What is this PR all about?

Compressor catches `CompressorThermodynamicCalculationError` and wraps it as `OutletFluidNotAchievableError` with operating-point context. `ShaftSpeedFinder` and `PipelineSectionSolver` handle the new failure type.

When equinor/ecalc#1543 was originally opened, the solver stack had no handling for EOS/flash failures from the compressor. Most of the surrounding infrastructure has since landed on main (typed failure hierarchy, rate-recovery helpers, solvers→finders rename). This update rebases onto current main and addresses only the remaining gap: catching thermodynamic calculation failures and propagating them as typed failures.

---

<details>
<summary>Previous PR description</summary>

## Type of Work

- [x] I have used conventional commits syntax

## What is this PR all about?

When the compressor's flash/EOS step fails (NaN properties, non-convergence, NeqSim Java exceptions near the dense/supercritical boundary), the failure used to escape the solver uncaught. `Compressor.propagate_stream`'s `except EcalcError` did not cover `Py4JJavaError`/JPype exceptions, so the original guard missed the cases that actually occur in production.

This PR introduces a typed `OutletFluidNotAchievableError` (carrying a `CompressorOperatingPoint`) raised by `Compressor.propagate_stream` on any thermodynamic-calculation failure, and threads it through the solver stack:

- `SpeedSolver` handles it at all three call sites (max-speed probe, min-speed probe, root-finding loop). When a boundary probe fails, it binary-searches inward to the highest/lowest EOS-feasible speed, anchored at the corresponding capacity-feasible speed so capacity and EOS narrowing don't interfere.
- A new `OutletFluidNotAchievableEvent` is produced as a typed solver failure.
- `OutletPressureSolver.find_solution` short-circuits on that event, so the unrunnable speed is not retried through anti-surge / pressure control.
- `get_anti_surge_solution` becomes `Optional`, propagating the EOS failure cleanly instead of asserting.

Catching only `EcalcError` at the flash boundary — but Java exceptions from NeqSim are not `EcalcError` subclasses, so this leaves the original bug in place.

Handling the recovery inside `Compressor` — but the solver already owns the speed-search logic and is the right place to decide what to do with a failed operating point.

</details>

### DIFF
--- a/src/libecalc/process/process_pipeline/process_error.py
+++ b/src/libecalc/process/process_pipeline/process_error.py
@@ -1,5 +1,17 @@
+from dataclasses import dataclass
+
 from libecalc.common.errors.exceptions import EcalcError
 from libecalc.process.process_pipeline.process_unit import ProcessUnitId
+
+
+@dataclass(frozen=True)
+class CompressorOperatingPoint:
+    inlet_pressure_bara: float
+    inlet_temperature_kelvin: float
+    actual_rate_m3_per_hour: float
+    polytropic_head_joule_per_kg: float
+    polytropic_efficiency: float
+    speed: float
 
 
 class ProcessError(EcalcError):
@@ -24,6 +36,18 @@ class RateTooLowError(OutsideCapacityError):
         self.actual_rate = actual_rate
         self.boundary_rate = boundary_rate
         self.process_unit_id = process_unit_id
+        super().__init__(reason)
+
+
+class OutletFluidNotAchievableError(ProcessError):
+    def __init__(
+        self,
+        process_unit_id: ProcessUnitId,
+        unachievable_operating_point: CompressorOperatingPoint,
+        reason: str = "Outlet fluid state is not achievable.",
+    ):
+        self.process_unit_id = process_unit_id
+        self.unachievable_operating_point = unachievable_operating_point
         super().__init__(reason)
 
 
--- a/src/libecalc/process/process_pipeline/process_error.py
+++ b/src/libecalc/process/process_pipeline/process_error.py
@@ -1,5 +1,19 @@
+from dataclasses import dataclass
+
 from libecalc.common.errors.exceptions import EcalcError
 from libecalc.process.process_pipeline.process_unit import ProcessUnitId
+
+
+@dataclass(frozen=True)
+class CompressorOperatingPoint:
+    """Context about the compressor state when a failure occurred."""
+
+    inlet_pressure_bara: float
+    inlet_temperature_kelvin: float
+    actual_rate_m3_per_hour: float
+    polytropic_head_joule_per_kg: float
+    polytropic_efficiency: float
+    speed: float
 
 
 class ProcessError(EcalcError):
@@ -92,3 +106,21 @@ class InsufficientInletPressureError(ProcessError):
             f"Inlet pressure {inlet_pressure_bara:.3f} bara is insufficient for required pressure drop "
             f"{required_delta_pressure_bara:.3f} bara."
         )
+
+
+class OutletFluidNotAchievableError(ProcessError):
+    """The compressor's EOS calculation could not produce a valid outlet stream.
+
+    This occurs when the thermodynamic flash at the computed outlet conditions
+    fails to converge, indicating a physically unachievable state.
+    """
+
+    def __init__(
+        self,
+        process_unit_id: ProcessUnitId,
+        unachievable_operating_point: CompressorOperatingPoint,
+        reason: str = "Outlet fluid state is not achievable.",
+    ):
+        self.process_unit_id = process_unit_id
+        self.unachievable_operating_point = unachievable_operating_point
+        super().__init__(reason)
--- a/src/libecalc/process/process_solver/solvers/speed_solver.py
+++ b/src/libecalc/process/process_solver/solvers/speed_solver.py
@@ -3,18 +3,26 @@ from collections.abc import Callable
 from dataclasses import dataclass
 
 from libecalc.process.fluid_stream.fluid_stream import FluidStream
-from libecalc.process.process_pipeline.process_error import RateTooHighError, RateTooLowError
+from libecalc.process.process_pipeline.process_error import (
+    OutletFluidNotAchievableError,
+    RateTooHighError,
+    RateTooLowError,
+)
 from libecalc.process.process_solver.boundary import Boundary
 from libecalc.process.process_solver.configuration import SpeedConfiguration
 from libecalc.process.process_solver.search_strategies import (
+    ACCEPT_AND_GO_HIGHER,
     ACCEPT_AND_GO_LOWER,
     REJECT_AND_GO_HIGHER,
     REJECT_AND_GO_LOWER,
     BinarySearchResult,
     RootFindingStrategy,
     SearchStrategy,
+    find_highest_true,
+    find_lowest_true,
 )
 from libecalc.process.process_solver.solver import (
+    OutletFluidNotAchievableFailure,
     Solution,
     Solver,
     TargetDirection,
@@ -63,6 +71,8 @@ class SpeedSolver(Solver[SpeedConfiguration]):
             return self._target_not_achievable_at_maximum_speed(maximum_speed_result)
 
         minimum_speed_result = self._minimum_speed_result(func, get_outlet_stream)
+        if isinstance(minimum_speed_result, Solution):
+            return minimum_speed_result
 
         if minimum_speed_result.outlet_stream.pressure_bara > self._target_pressure:
             return self._target_not_achievable_at_minimum_speed(minimum_speed_result)
@@ -76,6 +86,7 @@ class SpeedSolver(Solver[SpeedConfiguration]):
         return self._solve_within_pressure_bracket(
             get_outlet_stream=get_outlet_stream,
             minimum_speed_result=minimum_speed_result,
+            maximum_speed_result=maximum_speed_result,
         )
 
     def _maximum_speed_result(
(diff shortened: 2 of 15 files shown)


src/libecalc/process/process_units/compressor.py
```diff
@@ -56,12 +60,25 @@ def propagate_stream(self, inlet_stream: FluidStream) -> FluidStream:
...
-            fluid_service=self._fluid_service,
-        )
+        try:
+            return calculate_outlet_pressure_and_stream(
+                polytropic_efficiency=polytropic_efficiency,
+                polytropic_head_joule_per_kg=polytropic_head,
+                inlet_stream=inlet_stream,
+                fluid_service=self._fluid_service,
+            )
+        except Exception as exc:
```
olelod: This one can be tightened once fix/sverdrup-bug is merged, i guess.

src/libecalc/process/process_pipeline/process_error.py
```diff
@@ -27,6 +27,39 @@ def __init__(
...
+    chart capacity — the fluid itself is the problem.
+
+    All fields describe the exact operating point at failure, enough to reproduce
+    the issue in isolation.
+    """
+
+    def __init__(
+        self,
+        process_unit_id: ProcessUnitId,
+        inlet_pressure_bara: float | None = None,
```
tj098895: useful to gather these attributes in a reusable data structure? `InvalidCompressor/FluidState`. Does it make sense to use that other places?

src/libecalc/process/process_solver/solvers/speed_solver.py
```diff
@@ -34,6 +39,25 @@ def solve(self, func: Callable[[SpeedConfiguration], FluidStream]) -> Solution[S
         def get_outlet_stream(speed: float) -> FluidStream:
             return func(SpeedConfiguration(speed=speed))
 
+        def _fluid_not_achievable_solution(
+            e: OutletFluidNotAchievableError, configuration: SpeedConfiguration
+        ) -> Solution[SpeedConfiguration]:
+            logger.debug(f"Outlet fluid not achievable at speed {configuration.speed}", exc_info=e)
```
tj098895:
```suggestion
            logger.warning(f"Outlet fluid not achievable at speed {configuration.speed}", exc_info=e)
```
? Or not, since we might handle it? best that user is not aware of this? Or are there time when the user should be aware?

src/libecalc/process/process_solver/solvers/speed_solver.py
```diff
@@ -61,6 +85,34 @@ def get_outlet_stream(speed: float) -> FluidStream:
...
+            # Pre-check: if boundary.min also fails, no speed in the range produces a valid outlet.
+            try:
+                get_outlet_stream(speed=self._boundary.min)
+            except OutletFluidNotAchievableError as e:
+                return _fluid_not_achievable_solution(e, SpeedConfiguration(speed=self._boundary.min))
+
+            def max_speed_fluid_func(x: float) -> tuple[bool, bool]:
+                try:
+                    get_outlet_stream(speed=x)
+                    return True, True  # EOS succeeded — can try higher
```
tj098895: does the search support named tuples?

src/libecalc/process/process_solver/solvers/speed_solver.py
```diff
@@ -95,6 +147,34 @@ def bool_speed_func(x: float):
             )
             minimum_speed_configuration = SpeedConfiguration(speed=minimum_speed_within_capacity)
             minimum_speed_outlet_stream = func(minimum_speed_configuration)
+        except OutletFluidNotAchievableError:
```
tj098895: this method is getting long, would it make sense to split it into (reusable) steps/phases - to easier understand/read each step?

src/libecalc/process/process_units/compressor.py
```diff
@@ -56,12 +60,25 @@ def propagate_stream(self, inlet_stream: FluidStream) -> FluidStream:
...
+            # Any exception from the EOS flash layer (including Java/NeqSim exceptions
```
tj098895: are we entirely sure about that? what about temporary errors or bugs in neqsim code? it doesnt matter? we treat them as EOS errors anyway? (at least it is a third party error and in general we are not able to handle them, it should be logged though, if it is not.

tests/libecalc/process/process_solver/solvers/test_speed_solver.py
```diff
@@ -85,3 +87,217 @@ def speed_func(configuration: SpeedConfiguration):
     assert solution.configuration.speed == expected_speed
     outlet_stream = speed_func(solution.configuration)
     assert outlet_stream.pressure_bara == expected_pressure
+
+
+class FluidNotAchievableProcessUnit(ProcessUnit):
```
tj098895: is this like a "probe" to test the stream?

tj098895: does it need to be a processunit? can we rather mock the `propagate_stream` method on a `DummyProcessUnit` with "no" logic, since what we are testing is that the "framework" around can handle the error state correctly?

src/libecalc/process/process_units/compressor.py
olelod: Added a comment myself. Waiting for the PH flash hardening PR - to use the error/exception that will be introduced there.

src/libecalc/process/process_solver/solvers/speed_solver.py
olelod: You are right. Here we actually return a failure, and the user should know. The other logger.debugs in the same method we can recover from.

src/libecalc/process/process_pipeline/process_error.py
olelod: Good idea. Adding a CompressorOperatingPoint. Could not find one anywhere in the code. Placing it in process_error.py for now. Maybe it should be elsewhere?

src/libecalc/process/process_solver/solvers/speed_solver.py
olelod: Added a BinarySearchResult(NamedTuple). Deliberately did not update other sites where this could be useful (can see at least a couple). Leave it for another PR.

src/libecalc/process/process_solver/solvers/speed_solver.py
olelod: Yes!

src/libecalc/process/process_solver/solvers/speed_solver.py
```diff
@@ -121,8 +146,85 @@ def root_speed_func(x: float) -> float:
...
+    def _fluid_not_achievable_solution(
+        self,
+        e: OutletFluidNotAchievableError,
+        configuration: SpeedConfiguration,
+    ) -> Solution[SpeedConfiguration]:
+        logger.warning(f"Outlet fluid not achievable at speed {configuration.speed}", exc_info=e)
+        return Solution(
+            success=False,
+            configuration=configuration,
+            failure_event=OutletFluidNotAchievableEvent(
```
kjbrak: This new failure event still needs a caller-side short-circuit. `OutletPressureSolver.find_solution()` currently treats every failed speed solution as a pressure-control case: it applies the returned speed, runs anti-surge, and then re-runs the pipeline. For an `OUTLET_FLUID_NOT_ACHIEVABLE` result, that speed is specifically not runnable, so the typed solver failure can become an escaping `OutletFluidNotAchievableError` again. Consider returning this failure directly from `OutletPressureSolver` before anti-surge/pressure-control handling, preserving `speed_solution.failure_event` and the speed configuration.

src/libecalc/process/process_solver/solvers/speed_solver.py
```diff
@@ -61,6 +74,12 @@ def get_outlet_stream(speed: float) -> FluidStream:
                     source_id=e.process_unit_id,
                 ),
             )
+        except OutletFluidNotAchievableError:
+            result = self._search_highest_eos_feasible_speed(get_outlet_stream)
+            if isinstance(result, Solution):
+                return result
+            max_speed_configuration = result
+            maximum_speed_outlet_stream = get_outlet_stream(speed=result.speed)
```
kjbrak: When max speed fails EOS, this block adjusts `max_speed_configuration` to the highest EOS-feasible speed and computes `maximum_speed_outlet_stream` at that adjusted speed. If the target is still above that adjusted maximum, the failure branch below currently returns the original boundary max, which makes the returned configuration inconsistent with `achievable_value` and can cause callers to re-run the EOS-invalid speed. Consider returning `max_speed_configuration` from that target-not-achievable branch instead.

src/libecalc/process/process_solver/solvers/speed_solver.py
```diff
@@ -121,8 +146,85 @@ def root_speed_func(x: float) -> float:
...
+        try:
+            get_outlet_stream(self._boundary.min)
+        except OutletFluidNotAchievableError as e:
+            return self._fluid_not_achievable_solution(e, SpeedConfiguration(speed=self._boundary.min))
+
+        def feasible(x: float) -> BinarySearchResult:
+            try:
+                get_outlet_stream(x)
+                return ACCEPT_AND_GO_HIGHER
+            except (OutletFluidNotAchievableError, RateTooHighError, RateTooLowError):
```
kjbrak: This search can still leak a capacity error when the lower end of the speed range is invalid for capacity while the upper end is invalid for EOS. For example, `boundary.min` may raise `RateTooHighError`, an intermediate speed may be valid, and high speeds may raise `OutletFluidNotAchievableError`; the pre-check only catches EOS failures, and treating `RateTooHighError` as "go lower" moves in the wrong direction for speed-dependent max-flow capacity. Consider first narrowing to the lowest capacity-feasible speed, then searching for the highest EOS-feasible speed within that narrowed interval.

kjbrak (changes requested):
These potential issues were found together with AI, and I would like to discuss them together with you @olelod

src/libecalc/process/process_solver/solvers/speed_solver.py
olelod: Done — added in outlet_pressure_solver.py (lines 134–139):
 
```
  if isinstance(speed_solution.failure_event, OutletFluidNotAchievableEvent):
      return Solution(
          success=False,
          configuration=list(configurations.values()),
          failure_event=speed_solution.failure_event,
      )
```

src/libecalc/process/process_solver/solvers/speed_solver.py
olelod: The EOS-adjustment now flows through `_maximum_speed_result`: on `OutletFluidNotAchievableError` at `boundary.max`, it returns the `SpeedResult` from `_highest_speed_with_achievable_outlet`, i.e. the highest
 EOS-feasible speed and its outlet stream. The target-not-achievable branch then calls `_target_not_achievable_at_maximum_speed(result) `with that adjusted result, so both configuration (the adjusted max
 speed) and achievable_value (result.outlet_stream.pressure_bara) describe the same point — no inconsistency, no risk of callers re-running the EOS-invalid speed.

src/libecalc/process/process_solver/solvers/speed_solver.py
olelod: Done — `_highest_speed_with_achievable_outlet` now does exactly that two-step narrowing:
 
 
  1. `lower_bound = self._lowest_speed_within_capacity(get_outlet_stream) `— narrows up from `boundary.min` to the lowest capacity-feasible speed; only fails if no speed in the range avoids RateTooHighError.
  2. Binary-searches for the EOS boundary within `[lower_bound, boundary.max]`, treating `RateTooHighError` / `RateTooLowError` / `OutletFluidNotAchievableError` uniformly as REJECT_AND_GO_LOWER, so a capacity 
error never escapes as the answer.
 
 Symmetrically, `_lowest_speed_with_achievable_outlet` anchors on `_highest_speed_within_capacity` first. The classic failure mode you described (low end raises `RateTooHighError`, intermediate speeds valid, high
 speeds raise `OutletFluidNotAchievableError`) is covered by the parametrised cases in `test_speed_solver.py`.

PR #1546, opened 2026-05-07
fix(process): scale LiquidRemover mass rate by gas mass fraction

olelod:
## Type of Work

## What is this PR all about?

When `LiquidRemover` activates on a two-phase inlet, it removes the liquid phase from the composition but previously kept the total mass rate unchanged — so the dropped-out liquid mass was effectively re-injected into the gas stream downstream.

This PR scales the outlet mass rate by the gas mass fraction so the removed liquid is no longer carried by the gas:

```
gas_mass_fraction = vapor_fraction_molar * outlet_molar_mass / inlet_molar_mass
new_mass_rate     = inlet.mass_rate * gas_mass_fraction
```

Pure-vapor inlets are unchanged.

Also exposing the removed liquid as a separate outlet stream (e.g. for routing to an oil pump or emissions accounting). Left as a docstring note — it would change `propagate_stream`'s contract and there is no consumer for it yet.

## Between the lines?

Resolves equinor/ecalc-internal#1767

### DIFF
--- a/src/libecalc/process/process_units/liquid_remover.py
+++ b/src/libecalc/process/process_units/liquid_remover.py
@@ -1,6 +1,7 @@
 from typing import Final
 
 from libecalc.process.fluid_stream.constants import ThermodynamicConstants
+from libecalc.process.fluid_stream.exceptions import InvalidStreamException
 from libecalc.process.fluid_stream.fluid_service import FluidService
 from libecalc.process.fluid_stream.fluid_stream import FluidStream
 from libecalc.process.process_pipeline.process_unit import ProcessUnit, ProcessUnitId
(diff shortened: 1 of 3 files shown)


tj098895 (approved):
looks good. leaves it for @kjbrak to reverify

src/libecalc/process/process_units/liquid_remover.py
```diff
@@ -26,6 +33,12 @@ def propagate_stream(self, inlet_stream: FluidStream) -> FluidStream:
         """
         if inlet_stream.vapor_fraction_molar < ThermodynamicConstants.PURE_VAPOR_THRESHOLD:
             new_fluid = self._fluid_service.remove_liquid(inlet_stream.fluid)
-            return inlet_stream.with_new_fluid(new_fluid)
+            inlet_molar_mass = inlet_stream.fluid.molar_mass
+            if inlet_molar_mass > 0.0:
+                gas_mass_fraction = inlet_stream.vapor_fraction_molar * new_fluid.molar_mass / inlet_molar_mass
+            else:
+                gas_mass_fraction = 1.0
```
kjbrak: Consider failing explicitly when `inlet_molar_mass <= 0.0` rather than preserving the full inlet mass rate. A non-positive molar mass is non-physical, and `gas_mass_fraction = 1.0` would hide the invalid thermodynamic state while carrying all removed-liquid mass downstream again for that edge case.

If using `InvalidStreamException`, the matching import also needs to be added.

```suggestion
            if inlet_molar_mass <= 0.0:
                raise InvalidStreamException(
                    f"Cannot remove liquid from stream with non-positive molar mass: {inlet_molar_mass}"
                )
            gas_mass_fraction = inlet_stream.vapor_fraction_molar * new_fluid.molar_mass / inlet_molar_mass
```

kjbrak (approved):

PR #1550, opened 2026-05-11
chore: upgrade dependencies for week 20. Auto-generated by devCalc

markusrf:


uv.lock
```diff
@@ -11,7 +11,7 @@ resolution-markers = [
 ]
 
 [options]
-exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
```
tj098895, line 14: are you running old uv?

tj098895 (approved):

uv.lock
markusrf: 0.11.7, upgraded now

PR #1554, opened 2026-05-13
chore: expose configuration handlers

jsolaas:


src/libecalc/process/process_solver/temperature_setter_configuration_handler.py
```diff
@@ -0,0 +1,32 @@
...
+        return self._id
+
+    def handle_configuration(self, configuration: Configuration) -> None:
+        assert configuration.configuration_handler_id == self._id, (
+            f"Configuration id '{configuration.configuration_handler_id}' does not match configuration handler id '{self._id}'"
+        )
+        assert isinstance(configuration.value, TemperatureSetterConfiguration), (
+            f"Expected configuration value to be of type '{TemperatureSetterConfiguration.__name__}', got '{type(configuration.value)}'"
+        )
+        self._temperature_setter.set_temperature(configuration.value.temperature)
```
tj098895, line 29: should we also check that it is only set once and not manipulated?

src/libecalc/process/process_units/temperature_setter.py
```diff
@@ -19,6 +19,9 @@ def __init__(
     def get_id(self) -> ProcessUnitId:
         return self._id
 
+    def set_temperature(self, temperature: float) -> None:
```
tj098895:
```suggestion
    def set_temperature(self, temperature_kelvin: float) -> None:
```

be consisstent wrt variable naming?

tj098895 (approved):

src/libecalc/process/process_solver/temperature_setter_configuration_handler.py
jsolaas: That would be correct currently, but seems a bit weird to enforce here. What if we want to 'solve' temperature_setter instead?

PR #1557, opened 2026-05-15
chore: improve resource discovery and reference handling for new sections

frodehk:
## Summary

Small fixes and improvements to YAML resource discovery, reference handling, and mapper robustness. Discovered while running the migration tooling against existing models, but valuable independently.

## Changes

- **`yaml_reference_service.py`**: register `FLUID_MODELS` entries so they are resolvable through the reference service.
- **`pyyaml_yaml_model.py`**:
  - structure-agnostic helper to find `FILE` references nested inside the new YAML sections (`PROCESS_UNITS`, `PROCESS_SYSTEMS`)
  - auto-detect `is_single_speed` from chart resource headers
  - declare `fluid_models` as an abstract property on `YamlValidator`
- **`process_simulation_mapper.py`**: defensive `None`-handling in `_map_conditions`.

No behavioral change for existing models — these only enable correct handling of newer YAML constructs and improve type safety.

## Type of Work

### DIFF
--- a/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
+++ b/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
@@ -211,7 +211,9 @@ class ProcessSimulationMapper:
         else:
             return ref
 
-    def _map_conditions(self, condition: YamlExpressionType | None, conditions: list[YamlExpressionType] | None):
+    def _map_conditions(
+        self, condition: YamlExpressionType | None, conditions: list[YamlExpressionType] | None
+    ) -> YamlExpressionType | None:
         if condition is not None:
             assert isinstance(condition, ExpressionType)
             return condition
--- a/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
+++ b/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
@@ -172,12 +172,16 @@ class ProcessSimulationMapper:
             resource = self._resources.get(resource_name)
             if resource is None:
                 raise EcalcValidationException(f"Resource '{resource_name}' not found for variable speed chart.")
+            if "SPEED" not in resource.get_headers():
+                raise EcalcValidationException(
+                    f"Chart resource '{resource_name}' is missing required 'SPEED' column. "
+                    f"For single-speed charts, use the same speed value for all rows."
+                )
             try:
-                is_single_speed = "SPEED" not in resource.get_headers()
                 return UserDefinedChartData.from_resource(
                     resource,
                     units=yaml_chart.units,
-                    is_single_speed=is_single_speed,
+                    is_single_speed=False,
                     control_margin=control_margin_fraction,
                 )
             except InvalidResourceException as e:
--- a/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
+++ b/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
@@ -173,8 +173,12 @@ class ProcessSimulationMapper:
             if resource is None:
                 raise EcalcValidationException(f"Resource '{resource_name}' not found for variable speed chart.")
             try:
+                is_single_speed = "SPEED" not in resource.get_headers()
                 return UserDefinedChartData.from_resource(
-                    resource, units=yaml_chart.units, is_single_speed=False, control_margin=control_margin_fraction
+                    resource,
+                    units=yaml_chart.units,
+                    is_single_speed=is_single_speed,
+                    control_margin=control_margin_fraction,
                 )
             except InvalidResourceException as e:
                 raise InvalidChartResourceException(
@@ -208,12 +212,13 @@ class ProcessSimulationMapper:
             return ref
 
     def _map_conditions(self, condition: YamlExpressionType | None, conditions: list[YamlExpressionType] | None):
-        if condition:
+        if condition is not None:
             assert isinstance(condition, ExpressionType)
             return condition
-        else:
+        if conditions is not None:
             assert isinstance(conditions, list)
             return handle_condition_list(conditions)
+        return None
 
     def _get_regularity(self) -> Regularity:
         return Regularity(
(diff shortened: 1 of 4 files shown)


src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -180,8 +173,12 @@ def _get_compressor_chart(self, yaml_compressor_model_chart: YamlCompressorModel
             if resource is None:
                 raise EcalcValidationException(f"Resource '{resource_name}' not found for variable speed chart.")
             try:
+                is_single_speed = "SPEED" not in resource.get_headers()
```
tj098895: cant or doesnt it sometimes happen that speed is set even though it is the same? I seem to remember to have seen that, but may be wrong. I dont think that is robust enough...?

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -215,12 +212,13 @@ def _resolve_fluid_model_reference(self, ref: str | YamlFluidModel) -> YamlFluid
             return ref
 
     def _map_conditions(self, condition: YamlExpressionType | None, conditions: list[YamlExpressionType] | None):
-        if condition:
+        if condition is not None:
             assert isinstance(condition, ExpressionType)
             return condition
-        else:
+        if conditions is not None:
```
tj098895, line 224: I had to read this 3 times before I noticed the extra `s` :p

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -215,12 +212,13 @@ def _resolve_fluid_model_reference(self, ref: str | YamlFluidModel) -> YamlFluid
             return ref
 
     def _map_conditions(self, condition: YamlExpressionType | None, conditions: list[YamlExpressionType] | None):
```
tj098895:
```suggestion
    def _map_conditions(self, condition: YamlExpressionType | None, conditions: list[YamlExpressionType] | None) -> ???:
```

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -215,12 +212,13 @@ def _resolve_fluid_model_reference(self, ref: str | YamlFluidModel) -> YamlFluid
...
-        if condition:
+        if condition is not None:
             assert isinstance(condition, ExpressionType)
             return condition
-        else:
+        if conditions is not None:
             assert isinstance(conditions, list)
             return handle_condition_list(conditions)
+        return None
```
tj098895, line 227: Is this handled?

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
jsolaas: Could we require speed instead? Cleans up the chart definition if we can.

src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
```diff
@@ -310,10 +313,16 @@ def facility_resource_names(self) -> list[str]:
...
         resource_data = facility_input_data + model_curves_data
-        resource_names = [
+        resource_names: list[str] = [
             data.get(EcalcYamlKeywords.file) for data in resource_data if data.get(EcalcYamlKeywords.file) is not None
         ]
-        return resource_names
+
+        # Pick up FILE references nested in the new YAML sections (PROCESS_UNITS, PROCESS_SYSTEMS, ...).
+        for section in _NEW_SECTIONS_WITH_FILE_REFS:
+            resource_names.extend(_find_file_references(self._internal_datamodel.get(section)))
```
jsolaas, line 322: Also replace facility and model 'search' above? Search everything except TIME_SERIES? Or do we need to separate TIME_SERIES and FACILITY when looking at resources?

jsolaas (approved):

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
frodehk: Yes, `_map_rate` passes to `TimeSeriesExpression`, which handles `ExpressionType | None`

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
frodehk: Added -> `YamlExpressionType | None.`

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
frodehk: I agree. Requiring speed is better, but it is a breaking change — and better as separate PR? For now this is at least more correct than the previous hardcoded is_single_speed=False.

jsolaas: Since this is new code it is breaking anyway. Isn't requiring speed the same as not implementing this?

src/libecalc/presentation/yaml/yaml_models/pyyaml_yaml_model.py
frodehk: Good point. 

I kept the new search to the new sections (`PROCESS_UNITS`, `PROCESS_SYSTEMS`) on purpose. I think replacing the existing facility/model search will be a refactor that may require some test coverage - better to keep in a separate PR? I think `TIME_SERIES` probably needs to stay separate, since they map to a different resource type.

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
frodehk: I agree, s i too easy to miss.

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
frodehk: Added check for SPEED in `_get_compressor_chart`, i.e. speed is now required.

PR #1571, opened 2026-05-20
chore: use global period instead of global time vector

tj098895:
## Type of Work

### DIFF
--- a/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
+++ b/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
@@ -21,6 +21,7 @@ from libecalc.ecalc_model.process_simulation import (
     ProcessSimulation,
 )
 from libecalc.ecalc_model.time_series_configuration import (
+    TimeSeriesPressureDropperConfiguration,
     TimeSeriesTemperatureSetterConfiguration,
 )
 from libecalc.ecalc_model.time_series_stream import TimeSeriesStream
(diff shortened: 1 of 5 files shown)


src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -334,45 +334,52 @@ def map_process_simulation(
...
-                problem_configuration_handlers.append(choke_configuration_handler)
 
                 liquid_remover = LiquidRemover(fluid_service=self._fluid_service)
                 process_unit_map[temperature_setter.get_id()] = temperature_setter
-                process_unit_map[choke.get_id()] = choke
+                process_unit_map[pressure_dropper.get_id()] = pressure_dropper
                 process_unit_map[liquid_remover.get_id()] = liquid_remover
                 process_unit_map[compressor.get_id()] = compressor
 
+                # TODO: Pressure dropper AFTER temperature setter. Correct? Should not it be before? Does it matter?
```
jsolaas: Looking at CompressorTrainStage you are correct, the order is choke, tempsetter, liquidremover.

jsolaas (approved):

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
olelod: Today tempsetter and liquidremover are considered part of a **Stage**. Pressuredropper (ahead of stage) should  should ideally be taken out of **Stage** (if we want to keep liquidremover and tempsetter there) - and probably be included as an own item in Yaml, together with splitter and mixer, as something that happens between compression stages.

tj098895: good point. We will figure out :)

PR #1579, opened 2026-05-26
refactor(process): process units return FluidStream | PropagationFailure

olelod:
## Type of Work

- [x] I have added tests (if not, comment why) — existing tests that asserted on raised exceptions were ported to assert on returned `PropagationFailure` variants instead; no new coverage added because no new behaviour was introduced

## What is this PR all about?

Process units no longer raise exceptions when they can't propagate a stream — they return a typed `PropagationFailure` value instead.

**Signature change.** Every concrete `ProcessUnit.propagate_stream(...)` now returns:

```python
FluidStream | PropagationFailure
```

instead of returning `FluidStream` and raising on failure.

**Deletions.** `process_pipeline/process_error.py` is removed. The three exception classes it held — `RateTooHighError`, `RateTooLowError`, `OutsideOperationalCapacityError` — are gone. Each had a one-to-one counterpart variant on `PropagationFailure` introduced in PR 1; the migration is mechanical.

**Call-site updates.**
- `stream_propagator` forwards the union.
- Solvers (`speed`, `recirculation`, `upstream_choke`, `downstream_choke`, `multi_pressure`, `multi_shaft`, `outlet_pressure`, `feasibility`) drop `try/except Rate*Error` blocks in favour of `is_failure(...)` / `isinstance(..., PropagationFailure)` branches, and wrap the result in `Solution.failed(...)`.
- Pressure-control strategies (`common_asv`, `individual_asv`, `upstream_choke`, `downstream_choke`) do the same. `IndividualASVRateControlStrategy` keeps a private `_InfeasibleDuringSearch` exception **scoped inside** `get_outlet_pressure` to short-circuit `find_root` mid-iteration when a propagated failure makes continuing pointless; it never escapes the method.
- Tests that previously used `pytest.raises(Rate*Error)` were ported to assert on returned variants.

**No behavioural change.** Same inputs produce the same outputs, including the same failure outcomes. The change is in *how* the failure is delivered (return value vs. raised exception), not *what* the failure is.

## Between the lines?

**Stack**

1. [#1578](https://github.com/equinor/ecalc/pull/1578) — rename `SolverFailure` → `PropagationFailure`.
2. **This PR** — process units return `FluidStream | PropagationFailure`; unit exception classes deleted.
3. `typed-failures/03-search-strategies` — add `DidNotConverge` variant; search/root-finding return `float | DidNotConverge`; `DidNotConvergeError` deleted.


src/libecalc/domain/process/compressor/core/train/stage.py
```diff
@@ -130,7 +131,11 @@ def get_compressor_inlet_stream(
 
         # Then the stream passes through the PressureModifier (if defined),
         if self.choke is not None:
-            inlet_stream_after_choke = self.choke.propagate_stream(inlet_stream_after_mixer)
+            choke_result = self.choke.propagate_stream(inlet_stream_after_mixer)
+            if isinstance(choke_result, PropagationFailure):
+                # Legacy compressor train code has no PropagationFailure channel; surface as an exception
+                raise IllegalStateException(f"Choke could not propagate stream: {choke_result}")
```
tj098895, line 137: I think we created a new `IllegalStateException` class called `IllegalProcessStreamStateException` or sth, as the `IllegalStateException` is imo sth related to programming errors

tj098895: Also, it should send choke_result as a separate parameter instead of serializing it here.

tj098895: also include process_unit_id ?

src/libecalc/process/process_solver/solvers/recirculation_solver.py
```diff
@@ -33,79 +36,103 @@ def bool_func(x: float, mode: Literal["minimize", "maximize"]) -> tuple[bool, bo
...
             """
-            try:
-                func(RecirculationConfiguration(recirculation_rate=x))
-                return False if mode == "minimize" else True, True
-            except RateTooLowError:
+            result = func(RecirculationConfiguration(recirculation_rate=x))
+            if isinstance(result, RateTooLow):
                 return True, False
-            except RateTooHighError:
+            if isinstance(result, RateTooHigh):
```
tj098895, line 42: Rename `RateTooHigh` to `RateTooHighStreamErrorEvent` or similar? We need one error representing all the states/errors that can occur with a stream - ie violating policies of a process_unit;

- Rate too high/low
- Pressure too high/low
- Temperature too high/low
- Illegal composition? Is that possible? Incompatible composition? ie that fluid reaches a compressor?

Imo. every process unit should have policies against these, ie. need to make a decision on what are the limits, either hard limits or limits calculated such as compressor chart. We need to make sure that every process unit makes a decision on what do to in those cases, and that we know that all edge cases are checked and taken care of ... possibly enums of PropagationFailure, in order for a solver to make decisions on what to test next or to give up?

src/libecalc/domain/process/compressor/core/train/stage.py
olelod: this is the legacy compressor stage

PR #1581, opened 2026-05-26
refactor: introduce YamlProcessUnit types and rename to YamlProcessPipeline

frodehk:
Introduce `YamlProcessUnit` (`YamlPressureDropper`, `YamlTemperatureSetter`,
`YamlLiquidRemover`, `YamlCompressor`) in a new `yaml_process_units.py`
module, and rename `YamlSerialProcessSystem` → `YamlProcessPipeline`.

Preparation for removing `YamlCompressorStageProcessSystem` and letting
units stand directly in the pipeline. The mapper and `items` field are
not yet updated — that follows in a separate PR.

Refs:
equinor/ecalc-internal#1831

---

## Type of Work

### DIFF
--- a/src/libecalc/presentation/yaml/yaml_types/components/yaml_process_system.py
+++ b/src/libecalc/presentation/yaml/yaml_types/components/yaml_process_system.py
@@ -4,50 +4,13 @@ from pydantic import Field
 
 from libecalc.presentation.yaml.yaml_types import YamlBase
 from libecalc.presentation.yaml.yaml_types.components.yaml_expression_type import YamlExpressionType
-from libecalc.presentation.yaml.yaml_types.models.yaml_compressor_chart import UnitsField, YamlCurve, YamlUnits
-from libecalc.presentation.yaml.yaml_types.models.yaml_compressor_stages import YamlControlMarginUnits
+from libecalc.presentation.yaml.yaml_types.components.yaml_process_units import (
+    ProcessUnitReference,
+    YamlCompressor,
+)
 from libecalc.presentation.yaml.yaml_types.streams.yaml_inlet_stream import YamlInletStream
-from libecalc.presentation.yaml.yaml_types.yaml_data_or_file import DataOrFile
 
 StreamRef = str
-
-
-class YamlControlMargin(YamlBase):
-    unit: YamlControlMarginUnits
-    value: float
-
-
-class YamlCompressorChart(YamlBase):
-    curves: Annotated[
-        DataOrFile[list[YamlCurve]],
-        Field(description="Compressor chart curves, one per speed.", title="CURVES"),
-    ]
-    units: YamlUnits = UnitsField()
-
-
-class YamlCompressorModelChart(YamlBase):
-    type: Literal["COMPRESSOR_CHART"]
-    chart: YamlCompressorChart
-    control_margin: YamlControlMargin
-
-
-ProcessUnitReference = str
-
-
-class YamlCompressor(YamlBase):
-    """
-    A Compressor process unit
-    """
-
-    type: Literal["COMPRESSOR"]
-    name: Annotated[
-        ProcessUnitReference,
-        Field(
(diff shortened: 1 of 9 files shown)


src/libecalc/presentation/yaml/yaml_types/components/yaml_process_system.py
```diff
@@ -4,52 +4,15 @@
...
+    ProcessUnitReference,
+    YamlCompressor,
+)
 from libecalc.presentation.yaml.yaml_types.streams.yaml_inlet_stream import YamlInletStream
-from libecalc.presentation.yaml.yaml_types.yaml_data_or_file import DataOrFile
 
 StreamRef = str
 
 
-class YamlControlMargin(YamlBase):
```
tj098895, line 15: this is just new stuff? can we create a new directory for "new yaml" in order to make that more clear what is now and what is old ?

src/libecalc/presentation/yaml/yaml_types/components/yaml_process_units.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class YamlPressureDropper(YamlBase):
+    """A pressure dropper unit — reduces stream pressure by a fixed amount."""
+
+    type: Literal["PRESSURE_DROPPER"]
+    pressure_drop: Annotated[
+        YamlExpressionType,
+        Field(
+            description="Pressure drop across the unit [bar].",
```
tj098895, line 56: should we specify barA vs barG? I guess bar is too little info, it is not given of context if it is barg or bara? I do remember that some users have specified incorrect in the past. Should we enforce units...?

src/libecalc/presentation/yaml/yaml_types/components/yaml_process_units.py
```diff
@@ -0,0 +1,88 @@
...
+    the stream, leaving only the gas phase.
+
+    No-op if the stream is already pure vapor.
+    """
+
+    type: Literal["LIQUID_REMOVER"]
+
+
+YamlProcessUnit = Annotated[
+    YamlCompressor | YamlPressureDropper | YamlTemperatureSetter | YamlLiquidRemover,
```
tj098895, line 86: I assume it is on purpose to not include choke(s) here or recirculation loops, because that is given by strategy. I think that is a good choice for now.

tj098895 (approved):
nice, see comments :)

src/libecalc/presentation/yaml/yaml_types/components/yaml_process_units.py
tj098895: I see that we assume barA, so we should state that as well then :)

src/libecalc/presentation/yaml/yaml_types/components/yaml_process_units.py
frodehk: The idea was that I thought choke was reserved for solving purposes, pressure dropper for input. However, I see that the implementation is practically identical (PressureDropper and Choke) - so I am not sure why we need both.

src/libecalc/presentation/yaml/yaml_types/components/yaml_process_units.py
frodehk: Pressure drop is a difference, not an absolute value, so barA vs barG gives the same value. However, we may enforce unit - e.g. by introducing a YamlDeltaPressure (value, unit). Maybe wait till we see the need?

src/libecalc/presentation/yaml/yaml_types/components/yaml_process_system.py
frodehk: Good point. I will address this in another PR. I plan to create a yaml_types.process directory, and move e.g. yaml_process_units.py. Then split up yaml_process_system.py and create modules for process simulation, process pipeline, stream distribution etc. That will cover most of the new yaml as far as I can see.

PR #1582, opened 2026-05-26
refactor: add compressor solver-path matrix tests

kjbrak:
## Type of Work

## What is this PR all about?

Adds a **9-region × 5-mode test matrix** (45 trial cases) that exercises every solver path in both the legacy `CompressorTrainCommonShaft` and the new process-domain `OutletPressureSolver`.

Each trial case specifies an operating region on the compressor chart (internal, below surge, above stonewall, below/above speed limits, zero-rate, over-compression) and a pressure-control mode (upstream choke, downstream choke, individual ASV rate/pressure, common ASV). Tests assert:

- Success/failure status and failure reason
- Discharge pressure target achievement
- Speed boundary correctness (min/max/internal)
- Power output (MW)
- Pressure-control behavior (choke ΔP, recirculation rates, anti-surge)
- Chart area flag (legacy only)

Also extracts `ProcessSolverBuilder` into a reusable test helper (`tests/libecalc/process/helpers/`) and adds `PressureControlType` as a `Literal` type alias.

**Results:** All 45 legacy tests pass. 26 of 45 process tests pass; 19 are marked as strict xfails (see below).

### Discrepancies Found (xfails)

19 process solver xfails grouped by root cause hypothesis (to be investigated deeper):

| Regions | Modes | Count | Root Cause Hypothesis |
|---------|-------|-------|------------|
| R1 (all) | All 5 | 5 | SpeedSolver bracketing failure — min-speed evaluation hits stonewall, breaking the root-find bracket |
| R6 (all) | All 5 | 5 | SpeedSolver convergence near max-speed boundary — similar bracketing issue at the upper speed limit |
| R8 (all) | All 5 | 5 | No zero-rate short-circuit — legacy returns `power=0, status=OK` for zero rate; process solver attempts full evaluation |
| R3 | Upstream choke | 1 | Decoupled solving — anti-surge is optimized at nominal suction pressure before the choke reduces it, leading to over-recirculation and 7.5% power mismatch |
| R5 | Common ASV | 1 | Common-ASV topology wraps the entire pipeline in one recirculation loop, masking the above-max-flow failure that legacy reports |
| R9 | Upstream choke | 1 | Upstream choke reports success for a stonewall-limited point where legacy correctly reports above-max-flow |
| R9 | Common ASV | 1 | Common-ASV reports pressure-unreachable where legacy reports above-max-flow for below-min-speed near-stonewall region |

### Two-Stage Tests

The current matrix uses a single-stage compressor train. Future two-stage tests could involve:
- Interstage cooling and pressure-drop-ahead-of-stage
- Per-stage vs common ASV recirculation behavior
- Stage-to-stage stream propagation through the process pipeline
- Multi-stage speed coupling on a shared shaft

These will validate the process solver for production-representative train configurations where interstage interactions create additional solver complexity.

### DIFF
--- a/tests/libecalc/process/helpers/process_solver_builder.py
+++ b/tests/libecalc/process/helpers/process_solver_builder.py
@@ -1,0 +1,230 @@
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import assert_never
+
+from libecalc.ecalc_model.process_simulation import PressureControlConfig, PressureControlType
+from libecalc.process.fluid_stream.fluid_service import FluidService
+from libecalc.process.process_pipeline.process_pipeline import ProcessPipeline
+from libecalc.process.process_pipeline.process_unit import ProcessUnit
+from libecalc.process.process_solver.anti_surge.common_asv import CommonASVAntiSurgeStrategy
+from libecalc.process.process_solver.anti_surge.individual_asv import IndividualASVAntiSurgeStrategy
+from libecalc.process.process_solver.choke_configuration_handler import ChokeConfigurationHandler
+from libecalc.process.process_solver.outlet_pressure_solver import OutletPressureSolver
+from libecalc.process.process_solver.pressure_control.common_asv import CommonASVPressureControlStrategy
+from libecalc.process.process_solver.pressure_control.downstream_choke import DownstreamChokePressureControlStrategy
+from libecalc.process.process_solver.pressure_control.individual_asv import (
+    IndividualASVPressureControlStrategy,
+    IndividualASVRateControlStrategy,
+)
+from libecalc.process.process_solver.pressure_control.pressure_control_strategy import PressureControlStrategy
+from libecalc.process.process_solver.pressure_control.upstream_choke import UpstreamChokePressureControlStrategy
+from libecalc.process.process_solver.process_pipeline_runner import ProcessPipelineRunner
+from libecalc.process.process_solver.recirculation_loop import RecirculationLoop
+from libecalc.process.process_solver.search_strategies import RootFindingStrategy, ScipyRootFindingStrategy
+from libecalc.process.process_units.choke import Choke
+from libecalc.process.process_units.compressor import Compressor
+from libecalc.process.process_units.direct_mixer import DirectMixer
+from libecalc.process.process_units.direct_splitter import DirectSplitter
+from libecalc.process.process_units.liquid_remover import LiquidRemover
+from libecalc.process.process_units.temperature_setter import TemperatureSetter
+from libecalc.process.shaft import VariableSpeedShaft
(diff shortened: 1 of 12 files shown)


olelod (approved):

tests/libecalc/process/helpers/process_solver_builder.py
```diff
@@ -0,0 +1,230 @@
...
+from libecalc.process.process_units.compressor import Compressor
+from libecalc.process.process_units.direct_mixer import DirectMixer
+from libecalc.process.process_units.direct_splitter import DirectSplitter
+from libecalc.process.process_units.liquid_remover import LiquidRemover
+from libecalc.process.process_units.temperature_setter import TemperatureSetter
+from libecalc.process.shaft import VariableSpeedShaft
+from tests.libecalc.process.helpers.process_solver_system import ProcessSolverSystem, StageConfig
+
+
+class ProcessSolverBuilder:
```
tj098895, line 35: this solver model may be different from the one used in code, is it not possible to use the one in production code instead? now we are testing a test solver instead of the real solver, which is not ideal ...

tj098895: if the production code solver is not testable like this, that should instead be rewritten to be testable. we should not need to set it up in test.

tests/libecalc/process/process_solver/compressor_solver_path_matrix/cases.py
```diff
@@ -0,0 +1,249 @@
+"""Shared data model and trial cases for the solver-path matrix.
+
+Defines 9 operating regions × 5 pressure-control modes = 45 trial cases
```
tj098895, line 3: this is gold 💠

tests/libecalc/process/process_solver/compressor_solver_path_matrix/cases.py
```diff
@@ -0,0 +1,249 @@
...
+def _fail(
+    outcome: ExpectedOutcome,
+    power_mw: float,
+    pressure: PressureExpectation = PressureExpectation.NOT_ASSERTED,
+    control: ExpectedControlAction = ExpectedControlAction.NONE,
+) -> ExpectedResult:
+    return ExpectedResult(outcome=outcome, power_mw=power_mw, pressure_expectation=pressure, control_action=control)
+
+
+EXPECTED_RESULTS: dict[tuple[str, PressureControlType], ExpectedResult] = {
```
tj098895, line 170: what are the sources for these results?

kjbrak: its from the legacy solver results

tj098895: ok, should document. As in, we assume the legacy solver is correct. If we find bugs, and the newer code is correct, then we should probably fix the bug in legacy as well ...

PR #1583, opened 2026-05-26
fix: recompute anti-surge recirculation at each upstream-choke probe

olelod:
## Type of Work

## What is this PR all about?

Fixes **systematic over-recirculation** in the process solver when using upstream-choke pressure control on surge-pinned operating points.

### Problem

`OutletPressureSolver.find_solution` applies anti-surge recirculation *once* at unchoked suction pressure, then invokes `UpstreamChokePressureControlStrategy.apply()` which iterates through choke ΔP candidates without ever re-evaluating anti-surge. The recirculation rate remains frozen at its pre-choke value — which is higher than necessary at reduced (post-choke) suction. This causes over-recirculation and elevated power.

**Observed impact** (R3-UPSTREAM_CHOKE scenario, single-speed train, surge-pinned):
- Before fix: 2.41 MW
- After fix: 2.24 MW  
- Legacy reference: 2.23 MW

### Fix

`UpstreamChokePressureControlStrategy` now receives an `anti_surge_strategy` and `recirculation_loop_ids` at construction. Inside the choke search function (called at each ΔP probe):

1. **Zero** all recirculation loops — clearing stale state from the previous probe
2. **Re-evaluate** `anti_surge_strategy.apply(inlet_stream)` so recirculation is computed at the actual post-choke compressor inlet conditions
3. **Return** recirculation configurations alongside the choke configuration so the outer solver overwrites stale auto-ASV entries by handler ID

### Why zeroing is necessary

`IndividualASVAntiSurgeStrategy.apply()` calls `compressor.get_recirculation_range()` which returns `Boundary(min=max(0, min_rate - current_standard_rate))`. If recirculation from a previous probe is still applied, `current_standard_rate` is inflated → `boundary.min` is too low → ASV under-recirculates relative to true conditions. Zeroing ensures the ASV sees the genuine user-rate at the compressor inlet.

1. **ASV self-resets** (adding `_apply_recirculation_configuration(loop_id, 0.0)` inside `IndividualASVAntiSurgeStrategy.apply`): rejected because it mutates ASV behavior for all callers, not just the upstream-choke path.

2. **Passing full `current_configurations` into `PressureControlStrategy.apply()`**: rejected because it requires changing the interface of all 4 strategy implementations + all callers — too invasive for this fix.

3. **Chosen approach**: pass `recirculation_loop_ids` into `UpstreamChokePressureControlStrategy` only. Minimal interface change, scoped to the one strategy that needs it.

### DIFF
--- a/src/libecalc/process/process_solver/pressure_control/upstream_choke.py
+++ b/src/libecalc/process/process_solver/pressure_control/upstream_choke.py
@@ -1,6 +1,7 @@
 from collections.abc import Sequence
 
 from libecalc.process.fluid_stream.fluid_stream import FluidStream
+from libecalc.process.process_solver.anti_surge.anti_surge_strategy import AntiSurgeStrategy
 from libecalc.process.process_solver.boundary import Boundary
 from libecalc.process.process_solver.configuration import Configuration, ConfigurationHandlerId
 from libecalc.process.process_solver.float_constraint import FloatConstraint
@@ -22,6 +23,11 @@ class UpstreamChokePressureControlStrategy(PressureControlStrategy):
     The strategy models pressure control by reducing suction pressure (upstream choking).
     It computes a safe ΔP search interval from the inlet stream pressure at solve time,
     then uses a root-finding solver to find the ΔP that makes outlet pressure match the target.
+
+    When the demanded flow lies below the surge line, the anti-surge strategy is re-evaluated
+    at every choke probe so that the compressor's recirculation rate is consistent with the
+    actual (post-choke) suction pressure — rather than the pre-choke recirculation that
+    would otherwise be frozen by the outer solver.
     """
 
     def __init__(
@@ -29,10 +35,13 @@ class UpstreamChokePressureControlStrategy(PressureControlStrategy):
         simulator: ProcessRunner,
         choke_configuration_handler_id: ConfigurationHandlerId,
         root_finding_strategy: RootFindingStrategy,
+        anti_surge_strategy: AntiSurgeStrategy,
     ):
         self._choke_configuration_handler_id = choke_configuration_handler_id
         self._simulator = simulator
         self._root_finding_strategy = root_finding_strategy
+        self._anti_surge_strategy = anti_surge_strategy
+        self._last_anti_surge_configurations: Sequence[Configuration[RecirculationConfiguration]] = []
 
     def apply(
         self,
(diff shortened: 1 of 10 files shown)


src/libecalc/process/process_solver/pressure_control/upstream_choke.py
```diff
@@ -52,23 +62,42 @@ def apply(
...
                 Configuration(configuration_handler_id=self._choke_configuration_handler_id, value=config)
             )
-            return self._simulator.run(
-                inlet_stream=inlet_stream,
-            )
+            # Clear recirculation loops before re-evaluating anti-surge. The ASV strategy
+            # computes the *additional* recirculation needed and writes it as the total,
+            # so leftover recirculation from a previous probe would cause the next probe
+            # to systematically under-recirculate.
+            for loop_id in self._recirculation_loop_ids:
```
jsolaas: Would a reset for anti_surge_strategy make sense? upstream choke wouldn't have to know about the specific recirculation_loops in that case

src/libecalc/process/process_solver/pressure_control/upstream_choke.py
```diff
@@ -52,23 +62,42 @@ def apply(
...
+                        value=RecirculationConfiguration(recirculation_rate=0.0),
+                    )
+                )
+            # Re-evaluate anti-surge with the choke applied so the recirculation rate
+            # reflects the actual (post-choke) compressor inlet conditions.
+            anti_surge_solution = self._anti_surge_strategy.apply(inlet_stream=inlet_stream)
+            for anti_surge_configuration in anti_surge_solution.configuration:
+                self._simulator.apply_configuration(anti_surge_configuration)
+            recirculation_at_last_probe.clear()
+            recirculation_at_last_probe.extend(anti_surge_solution.configuration)
```
jsolaas: Is this the same?
```suggestion
            recirculation_at_last_probe = anti_surge_solution.configuration
```

tests/libecalc/process/process_solver/pressure_control/test_upstream_choke_pressure_control_strategy.py
```diff
@@ -24,6 +25,7 @@ def test_upstream_choke_strategy_asserts_when_baseline_below_target(
         simulator=runner,
         choke_configuration_handler_id=upstream_choke_configuration_handler.get_id(),
         root_finding_strategy=root_finding_strategy,
+        anti_surge_strategy=NoASVAntiSurgeStrategy(),
```
jsolaas, line 28: We also have the `libecalc.testing` module for stuff shared in tests, `NoASVAntiSurgeStrategy` could be put there if it isn't used outside tests

jsolaas (approved):

src/libecalc/process/process_solver/pressure_control/upstream_choke.py
olelod: Tried using the reset_to() first, but that reset everything - and caused trouble. But resetting only the anti-surge strategy should work. Will try.

src/libecalc/process/process_solver/pressure_control/upstream_choke.py
olelod: then it would need to be `nonlocal`? when it is updated within `choke_func`. could make it an internal attribute on the class (`self._last_anti_surge_configurations`)?

PR #1591, opened 2026-05-28
refactor: replace yaml compressor stages with process units in pipeline

frodehk:
Replaces the `YamlCompressorStageProcessSystem` with explicit process units (compressor, pressure dropper, temperature setter, liquid remover) declared directly in the pipeline. This gives users more flexibility to compose process pipelines.

Refs: equinor/ecalc-internal#1831

---

## Type of Work


src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -66,6 +69,9 @@
 from libecalc.process.process_solver.process_runner import ProcessRunner
 from libecalc.process.process_solver.recirculation_loop import RecirculationLoop
 from libecalc.process.process_solver.search_strategies import ScipyRootFindingStrategy
+from libecalc.process.process_solver.temperature_setter_configuration_handler import (
+    TemperatureSetterConfigurationHandler,
```
tj098895: has been removed ...

frodehk: Yes, I reverted another change as well. Fixed now.

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -316,45 +302,67 @@ def map_process_simulation(
...
+                        process_unit_map[temperature_setter.get_id()] = temperature_setter
+                        current_segment_unit_ids.append(temperature_setter.get_id())
+
+                    case YamlLiquidRemover():
+                        liquid_remover = LiquidRemover(fluid_service=self._fluid_service)
+                        process_unit_map[liquid_remover.get_id()] = liquid_remover
+                        current_segment_unit_ids.append(liquid_remover.get_id())
+
+                    case _:
+                        assert_never(yaml_process_unit)
```
tj098895: In my opinion we can handle an invalid process unit here by ignoring it and informing user that the process unit is not allowed and ignored. OR a validation exception and informing user that that type is not allowed there; and list which types that are allowed. because, this will lead to an unhandled hard ot read error for user.

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -316,45 +302,67 @@ def map_process_simulation(
...
+
+            # Trailing units (units after the last compressor) are not supported: each segment
```
tj098895, line 364: We ahve decided this or you have decided? I agree that it is what we have had so far, but I didnt think much about it. I guess it makes sense :)

frodehk: I decided:-) But my intention was just to be clear on how the current implementation is - what it means - not that it should be like this in the end.

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
frodehk: Good point. I'll will change to raising EcalcValidationException with the list of allowed types.

tj098895 (approved):
nice .)

PR #1594, opened 2026-05-28
fix(libecalc): reset recirculation inside AntiSurgeStrategy.apply()

olelod:
## Type of Work

No changelog entry: the only existing caller affected (`UpstreamChokePressureControlStrategy`) was already resetting externally, so user-visible behaviour is unchanged.

## What is this PR all about?

Closes equinor/ecalc-internal#1860.

`AntiSurgeStrategy.apply()` walks the train using the simulator's current configuration state. Without an internal reset, a second call to `apply()` sees the recirculation rate set by the first call still active on the loop, so the boundary check is computed against the wrong compressor inlet stream and the strategy returns a state-dependent result.

`CommonASVAntiSurgeStrategy.apply()` and `IndividualASVAntiSurgeStrategy.apply()` now call `self.reset()` as their first step, zeroing the recirculation handlers they own before recomputing. Callers no longer need to perform an external reset.

The explicit `self._anti_surge_strategy.reset()` workaround in `UpstreamChokePressureControlStrategy` is removed. `OutletPressureSolver` continues to work unchanged since its outer `reset_to()` already cleared state.

Regression test `test_common_asv_apply_result_independent_of_prior_simulator_state` calls `apply()` twice with the same inlet on a single-stage train and asserts the recirculation rates match. The test fails on `main` with:

```
AssertionError: apply() result depends on prior simulator state:
  first call set recirculation rate 500000.000,
  second call set 500976.562.
```

and passes with the fix applied.

- Leaving the explicit `reset()` in `UpstreamChokePressureControlStrategy` and only documenting the precondition on `apply()`. Rejected: any future caller would have to know the dance, and OPS only works today by coincidence of outer ordering.
- Resetting inside `_increase_recirculation_to_minimum_feasible` instead of at the top of `apply()`. Rejected: `apply()` is the public seam, the helper is an implementation detail.

## Between the lines?

`CommonASVPressureControlStrategy.apply()` has a similar shape with a comparable inline workaround. It is a refactor rather than a fix (the workaround is correct and the result is independent of prior state today), so it is tracked separately and not included here.

### DIFF
--- a/src/libecalc/process/process_solver/anti_surge/common_asv.py
+++ b/src/libecalc/process/process_solver/anti_surge/common_asv.py
@@ -46,7 +46,7 @@ class CommonASVAntiSurgeStrategy(AntiSurgeStrategy):
         )
 
     def apply(self, inlet_stream: FluidStream) -> Solution[Sequence[Configuration[RecirculationConfiguration]]]:
-        # Increase recirculation to give minimum feasible flow and return outlet.
+        self.reset()
         recirculation_solution = self._increase_recirculation_to_minimum_feasible(inlet_stream)
         return Solution(
             success=recirculation_solution.success,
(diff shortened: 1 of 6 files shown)


tj098895 (approved):
should common_asv in pressure control have the same? others? s

olelod:
> should common_asv in pressure control have the same? others? s

`CommonASVPressureControlStrategy.apply() ` and `IndividualASVPressureControlStrategy.apply()` already reset (but closer to call site) - i will lift those to start of `apply()` as well.

IndividualASVRateControlStrategy needs some discussion - i will make an issue.

PR #1608, opened 2026-06-01
test: add tests for process simulation mapper using yaml builders

frodehk:
Adds tests for `ProcessSimulationMapper`.

Copilot was used as a sparring partner to evaluate test scope, test code and challenge weak/low-value tests.

---

## Type of Work

### DIFF
--- a/tests/libecalc/presentation/yaml/mappers/test_process_simulation_mapper.py
+++ b/tests/libecalc/presentation/yaml/mappers/test_process_simulation_mapper.py
@@ -1,8 +1,5 @@
 from datetime import datetime
 
-import pytest
-
-from libecalc.common.errors.ecalc_validation_error import EcalcValidationException
 from libecalc.common.time_utils import Period
 from libecalc.process.process_solver.pressure_control.pressure_control_strategy import PressureControlType
 from libecalc.process.process_units.choke import Choke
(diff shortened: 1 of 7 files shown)


tests/libecalc/presentation/yaml/mappers/test_process_simulation_mapper.py
```diff
@@ -0,0 +1,213 @@
...
+    assert sim_individual.process_problems[0].get_anti_surge_strategy().type == "INDIVIDUAL_ASV"
+
+
+# ---------------------------------------------------------------------------
+# Tests: validation
+# ---------------------------------------------------------------------------
+
+
+def test_mapper_raises_when_pipeline_has_units_after_last_compressor(process_simulation_mapper):
+    """Trailing units (after the last compressor) are not allowed."""
```
jsolaas: Why is this rule needed?

jsolaas (approved):

tests/libecalc/presentation/yaml/mappers/test_process_simulation_mapper.py
frodehk: I think this is the only way we can avoid asking the user to specify segments explicitly. And for now, the compressor can work as a boundary to split in segments, as it is the only process unit that needs an ASV recirculation loop.

frodehk: But I realize that it is probably a better solution to move trailing units out of the segment-wrapping and out of the asv loop for individual asv. For common asv it does not make a difference. I have updated the code and the test.

PR #1609, opened 2026-06-01
fix(libecalc): tighten Solution.combine semantics and clean up OutletPressureSolver

olelod:
## Type of Work

- [x] I have added tests (if not, comment why) — new `test_solution.py` covers the failure-propagation semantics

## What is this PR all about?

Closes equinor/ecalc-internal#1873.

`Solution.combine` was used in two incompatible ways: strict sequential aggregation (`MultiPressureSolver`) and pure configuration merging (`OutletPressureSolver`). It silently dropped `failure`, hiding the cause of combined failures.

This PR tightens `combine` to mean **strict sequential aggregation**:

- If the running solution has already failed, return it unchanged. Subsequent stages cannot meaningfully extend a failed pipeline, and the failure must keep referring to the configurations it was generated from.
- Otherwise: `success` follows other, `failure` is whatever other carries, configurations are merged with later entries winning per handler.

`OutletPressureSolver.find_solution` is rewritten to use an explicit configuration list + module-level `merge_configurations(...)` at the single point where deduplication actually matters (pressure control may re-emit anti-surge configurations). No more accumulating into a dict only to flatten it at every return.

## Between the lines?

`MultiPressureSolver` gets correct failure propagation for free — it previously silently lost `pressure_control_solution.failure`. No call-site changes needed there.

### DIFF
--- a/src/libecalc/process/process_solver/outlet_pressure_solver.py
+++ b/src/libecalc/process/process_solver/outlet_pressure_solver.py
@@ -11,6 +11,7 @@ from libecalc.process.process_solver.configuration import (
     ConfigurationHandlerId,
     RecirculationConfiguration,
     SpeedConfiguration,
+    merge_configurations,
 )
 from libecalc.process.process_solver.float_constraint import FloatConstraint
 from libecalc.process.process_solver.pressure_control.pressure_control_strategy import PressureControlStrategy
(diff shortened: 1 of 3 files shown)


jsolaas (approved):
Nice, currently we are using the same Solution class several places, but there are different layers to the solvers, i.e. OutletPressureSolver could provide anti_surge config and pressure control config etc (not mixed together). SpeedSolver has a simpler solution. Maybe we should create separate solutions?

PR #1612, opened 2026-06-05
refactor(libecalc): implement reset_handler() on individual configuration handlers

olelod:
The reset_to() method felt a bit off. This is an attempt to let the configuration handlers own it's own neutral/default state, and let the strategies reset then using reset_handler() and config_handler_id.

The shaft is intentionally never reset in the solver flow, the shaft speed is always explicitiy set. Shaft.reset() still exists because it is an abstract method - and maybe at some point we would like to reset it.

### DIFF
--- a/src/libecalc/process/process_solver/process_pipeline_runner.py
+++ b/src/libecalc/process/process_solver/process_pipeline_runner.py
@@ -18,7 +18,6 @@ class ProcessPipelineRunner(ProcessRunner):
     def __init__(self, configuration_handlers: Sequence[ConfigurationHandler], units: Sequence[ProcessUnit]):
         self._configuration_handlers = {handler.get_id(): handler for handler in configuration_handlers}
         self._units = {unit.get_id(): unit for unit in units}
-        self._configurations: dict[ConfigurationHandlerId, Configuration] = {}
 
     @staticmethod
     def _apply_config_for_unit(configuration_handler: ConfigurationHandler, configuration: Configuration):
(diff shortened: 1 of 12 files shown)


src/libecalc/process/process_solver/process_pipeline_runner.py
```diff
@@ -18,26 +18,17 @@ class ProcessPipelineRunner(ProcessRunner):
     def __init__(self, configuration_handlers: Sequence[ConfigurationHandler], units: Sequence[ProcessUnit]):
         self._configuration_handlers = {handler.get_id(): handler for handler in configuration_handlers}
         self._units = {unit.get_id(): unit for unit in units}
-        self._configurations: dict[ConfigurationHandlerId, Configuration] = {}
```
olelod, line 21: Not used for anything?

src/libecalc/process/process_solver/process_pipeline_runner.py
```diff
@@ -18,26 +18,17 @@ class ProcessPipelineRunner(ProcessRunner):
...
-    def reset_to(
-        self,
-        configurations: Sequence[Configuration] = (),
-    ) -> None:
-        self._configurations.clear()
-        for handler in self._configuration_handlers.values():
-            handler.reset()
-        if configurations:
-            self.apply_configurations(configurations)
+    def reset_handler(self, handler_id: ConfigurationHandlerId) -> None:
```
jsolaas:
```suggestion
    def reset_configuration_handler(self, handler_id: ConfigurationHandlerId) -> None:
```

I read `reset_handler` as something that handles reset, not reset a configuration handler.

jsolaas (approved):
Nice! Implementing reset for strategies makes sense to me. Having a reset/default value for choke and recirc makes sense, since they can "do nothing". I guess we could say the same for shaft.

tj098895:
NICE :)

PR #1622, opened 2026-06-10
feat: isenthalpic stream mixing with configurable outlet pressure

kjbrak:
## Type of Work

## Why this PR?
The stream mixer had some simplified assumptions (fast and no PH flash, and probably some rationale based on the old fluid domain existing at the time it was made) that would not give correct outlet temperature in cases where temperatures or compositions of the inlet gases differed meaningfully. I was not based on a isenthalpic process.

# What in this PR?

Replace mass-weighted-temperature mixing with adiabatic (isenthalpic) mixing: conserve total enthalpy and PH-flash to the outlet pressure for the correct outlet temperature and phase. Considered adding a pressure strategy, with both min of inlets and mass weighted average pressure of inlets (a simplified representation of an ejector type mixer where it would be possible to mix in a somewhat higher pressure stream probably adn recover some of the pressure dowstream. But this was dropped since there is no use case yet. Min of inlets is the default in process simulators, and its also what the mixer in neqsim does.

Rename SimplifiedStreamMixer to StreamMixer and move it to process/fluid_stream. Update tests with an enthalpy-balance check and a split-then-mix regression test.


src/libecalc/process/fluid_stream/stream_mixer.py
```diff
@@ -0,0 +1,143 @@
...
+        pressure_strategy: MixerPressureStrategy = MixerPressureStrategy.LOWEST_INLET,
+    ):
+        self._fluid_service = fluid_service
+        self._pressure_strategy = pressure_strategy
+
+    def _outlet_pressure(self, streams: list[FluidStream], total_mass_rate: float) -> float:
+        """Determine outlet pressure according to the configured strategy."""
+        if self._pressure_strategy == MixerPressureStrategy.MASS_WEIGHTED_AVERAGE:
+            return sum(s.mass_rate_kg_per_h * s.pressure_bara for s in streams) / total_mass_rate
+        return min(s.pressure_bara for s in streams)
```
jsolaas: Actually applying the strategy pattern here by passing this function in to StreamMixer instead would allow for expansion without changing the StreamMixer. That might not be the goal here though.

src/libecalc/process/fluid_stream/stream_mixer.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # All streams must share the same EoS model so their enthalpies are on a
+        # consistent reference state and may be summed.
+        reference_eos_model = streams[0].fluid_model.eos_model
+        for s in streams[1:]:
+            if s.fluid_model.eos_model != reference_eos_model:
+                raise IncompatibleEoSModelsException(reference_eos_model, s.fluid_model.eos_model)
+
+        # Outlet pressure from the configured strategy
+        outlet_pressure = self._outlet_pressure(streams, total_mass_rate)
```
jsolaas:
```suggestion
        outlet_pressure = self._pressure_strategy(streams, total_mass_rate)
```
In case of passing in a strategy

jsolaas (approved):

src/libecalc/process/fluid_stream/stream_mixer.py
kjbrak: Yes, a callable is the best. I actually think i will just drop the pressure strategy alltogether, since we dont have a use case yet. Go with min(inlets), its the default in other simulators, and what neqsim does in its mixer too.

PR #1627, opened 2026-06-12
refactor(libecalc): rename single-variable solvers to finders

olelod:
Renames the solver classes and files that implement single-variable search to use the `Finder`/`Finding` naming convention introduced in the `Finder` base class.

Findings should be used to make Configurations and a sequence of Configurations will be a Solution from the higher level solvers.

## Changes

- `SpeedSolver` → `ShaftSpeedFinder` (file: `solvers/speed_solver.py` → `finders/shaft_speed_finder.py`)
- `RecirculationRateFinder` → `RecirculationLoopRateFinder` (file: `solvers/recirculation_rate_finder.py` → `finders/recirculation_loop_rate_finder.py`)
- `UpstreamChokeSolver` and `DownstreamChokeSolver` → `ChokeDeltaPressureFinder` (file: `solvers/upstream_choke_solver.py` and `solvers/downstream_choke_solver.py` → `finders/choke_delta_pressure_finders.py`)

## Next up

Solvers (pipeline_section_solver, multi_pressure_solver etc.) should move to solvers


src/libecalc/process/process_solver/finder.py
```diff
@@ -0,0 +1,31 @@
...
+
+from libecalc.process.fluid_stream.fluid_stream import FluidStream
+from libecalc.process.process_solver.solver import SolverFailure
+
+
+@dataclass
+class Finding[TConfiguration]:
+    """Result of a Finder.find() call.
+
+    Always contains a configuration — the found one on success, the closest reached on failure.
```
olelod, line 15: Can be debated if we should return anything on failure? But I guess it has value to be able to see how close one got?

tests/libecalc/process/process_solver/finders/test_shaft_speed_finder.py
```diff
@@ -78,10 +77,8 @@ def speed_func(configuration: SpeedConfiguration):
         shaft.set_speed(configuration.speed)
         return propagate_stream_many(process_units=process_units, inlet_stream=inlet_stream)
 
-    solution = speed_solver.solve(speed_func)
+    result = speed_search.find(speed_func)
 
-    assert solution.success == solution_found
```
frodehk, line 83: `test_shaft_speed_finder` no longer asserts success for unreachable targets. Should there be a corresponding test to verify it catches these cases downstream?

frodehk (approved):
Nice. A little comment: The finder now only finds the best available speed — deciding whether that speed meets the target is the solver's responsibility. This is a behavior change, not just a rename. Worth to mention i PR maybe?

tests/libecalc/process/process_solver/finders/test_shaft_speed_finder.py
olelod: Changed to returning failures, and letting solver potentially recover from `max_above_target` with pressure control.

PR #1629, opened 2026-06-12
feat: multi-pressure target constraints

frodehk:
Adds optional name field to all process unit types and changes constraints from one-per-pipeline to a list of per-unit pressure targets. Any named process unit — compressor, mixer, splitter, temperature setter, pressure dropper, or liquid remover — can now be targeted by a constraint. This enables interstage pressure targets, e.g. setting different pressures at LP compressor outlet, mixer outlet, and pipeline outlet.

Also fixes ASV loop chunking so that interstage units between a compressor and a mixer/splitter are correctly placed outside the ASV loop.

**Future work**: When adding `PRESSURE_CONTROL` per constraint, ASV-based strategies (`COMMON_ASV`, `INDIVIDUAL_ASV`) should be validated to only apply to sections containing at least one compressor. Choke strategies can apply to any section.

Refs equinor/ecalc-internal#1898
---

## Type of Work

### DIFF
--- a/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
+++ b/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
@@ -339,9 +339,9 @@ class ProcessSimulationMapper:
             ] = {}
             unit_name_to_id_map: dict[str, ProcessUnitId] = {}  # reset per pipeline
 
-            # Pipeline chunking: compressor segments are wrapped in ASV loops,
+            # Pipeline segmenting: compressor segments are wrapped in ASV loops,
             # other units (mixers, splitters, interstage conditioning) are standalone.
-            pipeline_chunks: list[tuple[bool, list[ProcessUnitId]]] = []  # (wrap_in_asv, ids)
+            pipeline_segments: list[tuple[bool, list[ProcessUnitId]]] = []  # (wrap_in_asv, ids)
             current_segment_unit_ids: list[ProcessUnitId] = []
             for yaml_pipeline_item in item.items:
                 yaml_process_unit = yaml_pipeline_item.target
@@ -361,7 +361,7 @@ class ProcessSimulationMapper:
                         compressor_ids.append(compressor.get_id())
                         current_segment_unit_ids.append(compressor.get_id())
                         # Compressor should be wrapped in ASV-loop
-                        pipeline_chunks.append((True, current_segment_unit_ids))
+                        pipeline_segments.append((True, current_segment_unit_ids))
                         current_segment_unit_ids = []
 
                     case YamlPressureDropper():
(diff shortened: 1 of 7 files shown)


src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -298,6 +314,8 @@ def map_process_simulation(
         ] = {}
 
         process_pipeline_reference_to_id_map: dict[str, ProcessPipelineId] = {}
+        compressor_names_per_pipeline: dict[str, dict[str, ProcessUnitId]] = {}
```
jsolaas: Should the constraint be for compressors? Isn't it the mixer or splitter we want to constrain?

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -414,15 +443,29 @@ def map_process_simulation(
             process_pipeline_reference_to_id_map[item.name] = process_pipeline.get_id()
             process_pipelines.append(process_pipeline)
 
-        for process_pipeline_reference, constraint in yaml_process_simulation.constraints.items():
-            process_pipeline_id = process_pipeline_reference_to_id_map.get(process_pipeline_reference)
+        for constraint_key, constraint in yaml_process_simulation.constraints.items():
```
jsolaas: Should we introduce types of constraints in yaml? The generic dot-notation indicates to me that anything could be constrained, but that isn't the case.

Instead we could introduce a `PIPELINE_CONSTRAINT` that requires outlet_pressure and a pipeline reference, `COMPRESSOR_CONSTRAINT` that requires a pipeline and compressor reference + outlet_pressure. Or maybe it should be a mixer/splitter constraint instead?

frodehk: Edit: I see that typed constraints probably is better. Slightly more verbose - but worth it, both from a user perspective - and for setting up a correct pipeline that aligns with MultiPressureSolver

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
frodehk: Changed, applies to any process unit now. Next pr: allow pressure control for each constraint

olelod:
I am wondering about the relationship between `constraints` and `pipeline_sections`. The way i see it, each constraint defines one `PipelineSection`, meaning that all `ProcessUnits` between two consecutive constraints form one section, each with its own solver and pressure target. `MultiPressureSolver` is a chain of `PipelineSectionSolvers`, one per constraint. 

Do we need some kind of validation that there is one constraint per pipeline section?

frodehk:
> I am wondering about the relationship between `constraints` and `pipeline_sections`. The way i see it, each constraint defines one `PipelineSection`, meaning that all `ProcessUnits` between two consecutive constraints form one section, each with its own solver and pressure target. `MultiPressureSolver` is a chain of `PipelineSectionSolvers`, one per constraint.
> 
> Do we need some kind of validation that there is one constraint per pipeline section?

Yes, each constraint maps to one section in the solver. But `PipelineSection` as defined today only fits compressor sections (shaft, speed boundary, anti-surge). When we include constraints on non-compressor units (e.g. mixer with choke), it does not fit this structure. I guess PipelineSection needs to be generalized, or we need different section types?

frodehk:
Closing for split in smaller PRs. Work is preserved here for reference.

PR #1630, opened 2026-06-12
refactor(libecalc): introduce BisectStrategy base with BisectLowest and BisectHighest

olelod:
## What

Extracts/re-writes the bisection loop from `BinarySearchStrategy` into a `BisectStrategy` base class with a single `_bisect()` method. Two new subclasses are added:

- **`BisectLowest`** — finds the lowest `x` in a boundary where a boolean predicate is `True`
- **`BisectHighest`** — finds the highest `x` in a boundary where a boolean predicate is `True`

`BinarySearchStrategy`, which returns a tuple of booleans, becomes a subclass, delegating its loop to `_bisect()`.

The function return type is made explicit with a `BisectResult` named tuple (`higher`, `accepted`), replacing anonymous `tuple[bool, bool]` throughout.

## Bug fixed

Edit: this bug fix has been extracted into a separate PR: https://github.com/equinor/ecalc/pull/1632

~~The original `BinarySearchStrategy` only updated its convergence criterion (`rel_diff`) when a probe was `accepted`. In cases where several consecutive probes are rejected (e.g. the compressor hits stonewall for many consecutive candidate speeds), the convergence check never occured — the interval kept halving but the loop didn't notice, burned through all 20 iterations, and raised `DidNotConvergeError`.~~

~~The new shared `_bisect` loop checks interval width on every step regardless of acceptance. Bisection halves the interval geometrically every step — we just needed to check it every time, not only when lucky enough to get an accepted point.~~

~~This fixes the R1 and R6 path-matrix xfails, which both failed with `DidNotConvergeError` for this reason.~~

## What's next

The `BisectLowest`/`BisectHighest` primitives enable a follow-up fix where `SpeedSolver` narrows its own speed boundary before solving, 
by searching upward from minimum speed when `RateTooHighError` (stonewall) or `CompressorThermodynamicCalculationError` (EOS failure) is encountered, 
and searching downward from maximum speed when `CompressorThermodynamicCalculationError` is encountered.

### DIFF
--- a/src/libecalc/process/process_solver/search_strategies.py
+++ b/src/libecalc/process/process_solver/search_strategies.py
@@ -1,5 +1,6 @@
 import abc
 from collections.abc import Callable
+from typing import NamedTuple
 
 from scipy.optimize import root_scalar
 
@@ -7,6 +8,11 @@ from libecalc.common.errors.exceptions import EcalcError
 from libecalc.process.process_solver.boundary import Boundary
 
 CONVERGENCE_TOLERANCE = 1e-5
+
+
+class BisectResult(NamedTuple):
+    higher: bool
+    accepted: bool
 
 
 class DidNotConvergeError(EcalcError):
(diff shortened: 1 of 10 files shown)


kjbrak:
Are there too many abstractions when the only "Search strategy" we have and are planning to have per now is the bisection, should we simplify? Or maybe not, would be nice to swap in another search strategy I guess..

The `BisectLowest/BisectHighest` "primitives" are evaluating the probe function 2x more than it should. Could add a probe() function in the search method that evaluated once and set the two booleans based on that.

I will continue the review next week.

src/libecalc/process/process_solver/search_strategies.py
```diff
@@ -39,7 +42,34 @@ def __init__(self, tolerance: float = CONVERGENCE_TOLERANCE, max_iterations: int
...
+            else:
+                x1 = x2
+            rel_diff = (x1 - x0) / max(abs(x0), abs(x1), 1.0)
+            if rel_diff < self._tolerance:
+                break
+        return last_accepted
+
+
+class BinarySearchStrategy(BisectStrategy):
+    def __init__(self, tolerance: float = CONVERGENCE_TOLERANCE, max_iterations: int = 20):
```
kjbrak: This init is redundant since the parent class init has the same singature (but maybe it gives clarity? idk)

src/libecalc/process/process_solver/solvers/recirculation_solver.py
```diff
@@ -118,16 +118,16 @@ def _bool_func(
         func: Callable[[RecirculationConfiguration], FluidStream],
         x: float,
         mode: Literal["minimize", "maximize"],
-    ) -> tuple[bool, bool]:
+    ) -> BisectResult:
         """Probe a candidate rate for the search strategy.
 
         Returns ``(is_higher, is_accepted)``. The two booleans are decoupled so an
```
kjbrak: It returns `BisectResult`

kjbrak:
### Consolidation suggestion
Can consider consolidating the bisection classes to one bisection helper class:

~~~
 class Bisect:
     def __init__(self, tolerance=CONVERGENCE_TOLERANCE, max_iterations=20): ...
     def _bisect(self, boundary, probe) -> float | None: ...        # shared core
     def search(self, boundary, func) -> float: ...                 # was BinarySearchStrategy.search
     def lowest_true(self, boundary, predicate) -> float | None: ...   # was BisectLowest
     def highest_true(self, boundary, predicate) -> float | None: ...  # was BisectHighest
     def _first_true(self, boundary, predicate, *, lowest): ...     # shared direction-parameterised body
~~~

It would remove some duplicated code and the two search methods with different signatures.

### Suggestion: Removing strategy injection?
Now that you removed the SearchStrategy abc class, if there is only one "strategy" planned for this in e.g. SpeedSolver and RecirculationSolver then maybe we could consider just importing the bisection helper in those modules and using it directly instead of plumbing it through as a strategy which implies there are multiple implementations? Maybe we want this strategy injection everywhere, but not sure what good it does before we actually have multiple interchangeable strategies planned. 

note: If the caller of the solver classes needs to set tolerance then maybe it would need to be an init arg on the solver classes, but if not and the tolerance is always the same for each solver it can be hardcoded as a constant in the solver or when instantiating the bisection class.

### Full suggestion for the consolidation
~~~
class Bisect:
    """Bisection over a numeric range to locate a boolean threshold.

    The condition is assumed monotonic (a single True/False crossing). Probes are evaluated
    at the midpoint of a shrinking bracket until it is narrower than ``tolerance``.
    """

    def __init__(self, tolerance: float = CONVERGENCE_TOLERANCE, max_iterations: int = 20):
        """

        Args:
            tolerance: The relative bracket width at which the search stops.
            max_iterations: The maximum number of bisection steps.
        """
        self._tolerance = tolerance
        self._max_iterations = max_iterations

    def _bisect(
        self,
        boundary: Boundary,
        probe: Callable[[float], BisectResult],
    ) -> float | None:
        """Bisect [boundary.min, boundary.max]. ``probe(x)`` returns a ``BisectResult``.

        Returns the last accepted x, or ``None`` if no probe was ever accepted.
        """
        x0, x1 = boundary.min, boundary.max
        last_accepted: float | None = None
        for _ in range(self._max_iterations):
            x2 = (x0 + x1) / 2
            higher, accepted = probe(x2)
            if accepted:
                last_accepted = x2
            if higher:
                x0 = x2
            else:
                x1 = x2
            rel_diff = (x1 - x0) / max(abs(x0), abs(x1), 1.0)
            if rel_diff < self._tolerance:
                break
        return last_accepted

    def search(self, boundary: Boundary, func: Callable[[float], BisectResult]) -> float:
        """Highest x in ``boundary`` where the indicator holds (``max(x) given f(x) == True``).

        ``func`` returns a ``BisectResult`` so steering (``higher``) and validity (``accepted``)
        are decoupled: a probe can steer the search without being accepted as a solution.

        We assume f(x) to be a binary (Heaviside step) function where f(x) is 1 for x <= n and 0 for x > n.
        n is the target value in this optimization. x == n is the highest possible value of x before f(x) turns to 0.

        Raises ``DidNotConvergeError`` if no probe was ever accepted.
        """
        result = self._bisect(boundary=boundary, probe=func)
        if result is None:
            raise DidNotConvergeError(
                boundary=boundary,
                tolerance=self._tolerance,
                iterations=self._max_iterations,
            )
        return result

    def lowest_true(self, boundary: Boundary, predicate: Callable[[float], bool]) -> float | None:
        """Return the lowest x in ``boundary`` where ``predicate(x)`` is True, or ``None``."""
        return self._first_true(boundary, predicate, lowest=True)

    def highest_true(self, boundary: Boundary, predicate: Callable[[float], bool]) -> float | None:
        """Return the highest x in ``boundary`` where ``predicate(x)`` is True, or ``None``."""
        return self._first_true(boundary, predicate, lowest=False)

    def _first_true(
        self,
        boundary: Boundary,
        predicate: Callable[[float], bool],
        *,
        lowest: bool,
    ) -> float | None:
        """Locate the boundary x where ``predicate`` flips, searching from the ``near`` end.

        ``lowest`` selects the direction: True searches up from ``boundary.min`` for the lowest
        passing x; False searches down from ``boundary.max`` for the highest passing x.
        """
        near, far = (boundary.min, boundary.max) if lowest else (boundary.max, boundary.min)
        if predicate(near):
            return near
        if not predicate(far):
            return None

        def probe(x: float) -> BisectResult:
            ok = predicate(x)
            return BisectResult(higher=ok != lowest, accepted=ok)

        result = self._bisect(boundary=boundary, probe=probe)
        return result if result is not None else far
~~~

kjbrak (approved):

PR #1636, opened 2026-06-15
refactor: per-constraint pressure control and pipeline sectioning by constraint

frodehk:
Restructure the process-simulation YAML and mapping to support multiple pressure targets along one pipeline.

To model trains where different stages meet different pressures (e.g. an interstage export/injection target plus a final discharge), pressure control and anti-surge must be set **per constraint**, not once per pipeline.

Main changes:
- `PRESSURE_CONTROL` and `ANTI_SURGE` move into each constraint; `ANTI_SURGE` is optional (omitted for compressor-free sections).
- A constraint targets a named `UNIT`, or the pipeline outlet when `UNIT` is omitted.
- The mapper splits a pipeline into one section per constraint and resolves each section's topology and validation independently.

Refs equinor/ecalc-internal#1898

<details>
<summary>YAML before/after</summary>

**Before:**
```yaml
PROCESS_SIMULATIONS:
  - NAME: export
    TARGETS:
      - TARGET: export_train

    # One pressure-control strategy per process
    PRESSURE_CONTROL:
      export_train: DOWNSTREAM_CHOKE
      # anti-surge strategy was implicitly derived from pressure control

    # One outlet pressure constraint per process
    CONSTRAINTS:
      export_train:
        OUTLET_PRESSURE: 180

    STREAM_DISTRIBUTION:
      ...
```

**After:**
```yaml
PROCESS_SIMULATIONS:
  - NAME: export
    TARGETS:
      - TARGET: export_train

    CONSTRAINTS:
      export_train:
        - PROCESS_UNIT: lp_compressor          # section 1: ...→ lp_compressor
          OUTLET_PRESSURE: 30
          PRESSURE_CONTROL: INDIVIDUAL_ASV_PRESSURE
          ANTI_SURGE: INDIVIDUAL_ASV
        - PROCESS_UNIT: hp_compressor          # section 2: ...→ hp_compressor
          OUTLET_PRESSURE: 90
          PRESSURE_CONTROL: DOWNSTREAM_CHOKE
          ANTI_SURGE: COMMON_ASV
        - OUTLET_PRESSURE: 180         # section 3: fuel_offtake (no compressor)
          PRESSURE_CONTROL: DOWNSTREAM_CHOKE
          ANTI_SURGE: NO_ASV — section has no compressor

    STREAM_DISTRIBUTION:
      ...
```

The train (`export_train`), showing why the final section contains no compressor — a splitter
follows the last compressor:
```yaml
PROCESS_PIPELINES:
  export_train:
    NAME: my_export_train
    TYPE: SERIAL
    ITEMS:
      - NAME: lp_compressor
        TARGET:
          TYPE: COMPRESSOR
          COMPRESSOR_MODEL: ...
      - NAME: hp_compressor
        TARGET:
          TYPE: COMPRESSOR
          COMPRESSOR_MODEL: ...
      - NAME: fuel_offtake
        TARGET:
          TYPE: SPLITTER
          NAME: fuel_offtake             # The final constraint has no PROCESS_UNIT, so it applies to the process outlet and is mapped to the last process unit in the section
          OFFTAKE_RATE: ...
```

</details>

---

## Type of Work


src/libecalc/ecalc_model/process_simulation.py
```diff
@@ -63,14 +63,12 @@ class ProcessProblem(Entity[ProcessProblemId]):  # TODO: Rename to subproblem?
 
     def __init__(
         self,
-        pressure_control_strategy: PressureControlConfig,
```
jsolaas, line 66: We need this information still, right? Should Constraint change?

frodehk: Nice catch:-) Will update.

olelod:
Are we in a situation where the user is allowed to define conflicting anti surge and pressure control strategies now? Should we validate somehow that e.g. pressure control can not be COMMON_ASV if the anti surge topology is INDIVIDUAL_ASV? Or will it be handled in mapping later?

frodehk:
> Are we in a situation where the user is allowed to define conflicting anti surge and pressure control strategies now? Should we validate somehow that e.g. pressure control can not be COMMON_ASV if the anti surge topology is INDIVIDUAL_ASV? Or will it be handled in mapping later?

Yes we can. My thought was to validate in mapping later. But it may be better to take it in YamlProcessSimulation right away?

olelod:
> > Are we in a situation where the user is allowed to define conflicting anti surge and pressure control strategies now? Should we validate somehow that e.g. pressure control can not be COMMON_ASV if the anti surge topology is INDIVIDUAL_ASV? Or will it be handled in mapping later?
> 
> Yes we can. My thought was to validate in mapping later. But it may be better to take it in YamlProcessSimulation right away

I dont have a clear opinion on this. I guess both work. But conceptually - should the mapper take a valid YAML and map it, or is the job also to make sure the YAML is valid?

frodehk:
> > > Are we in a situation where the user is allowed to define conflicting anti surge and pressure control strategies now? Should we validate somehow that e.g. pressure control can not be COMMON_ASV if the anti surge topology is INDIVIDUAL_ASV? Or will it be handled in mapping later?
> > 
> > 
> > Yes we can. My thought was to validate in mapping later. But it may be better to take it in YamlProcessSimulation right away
> 
> I dont have a clear opinion on this. I guess both work. But conceptually - should the mapper take a valid YAML and map it, or is the job also to make sure the YAML is valid?

The problem is that we currently support both inline definitions (ok to validate in Yaml-classes) and references (must be resolved). In the mapper we have resolved the references, så for now I think we need to have it in the mapper

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -288,12 +292,29 @@ def map_anti_surge_strategy(
...
+        """Validate that ANTI_SURGE and PRESSURE_CONTROL strategies are compatible."""
+        for yaml_target_item in yaml_process_simulation.targets:
+            item = self._resolve_train_reference(yaml_target_item.target)
+            pipeline_constraints = yaml_process_simulation.constraints.get(item.name)
+
+            if not pipeline_constraints:
+                raise EcalcValidationException(f"Missing constraint for pipeline '{item.name}'")
+
+            for i, constraint in enumerate(pipeline_constraints):
+                if (item.anti_surge, constraint.pressure_control) in INCOMPATIBLE_STRATEGIES:
```
jsolaas: What about the other way around?

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -92,6 +92,12 @@
     HasValidity,
 )
 
+INCOMPATIBLE_STRATEGIES = [
```
tj098895: I prefer to list ALLOWED_COMBINED_STRATEGIES instead, and possibly include the disallowed ones, commented out, with a comment why it is not allowed - and then also use that to define which (edge) cases to test. -and expected outcome

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -266,37 +236,11 @@ def _map_fluid_model(self, yaml_fluid_model: YamlFluidModel) -> FluidModel:
             case _:
                 assert_never(yaml_fluid_model)
```
frodehk, line 236: `map_anti_surge_strategy` was not used.

src/libecalc/ecalc_model/process_simulation.py
```diff
@@ -38,19 +39,21 @@ class IndividualStreamDistributionConfig:
...
-
 @value_object
 class PressureControlConfig:  # Spec
     type: PressureControlType
 
 
 @value_object
 class AntiSurgeConfig:
-    type: Literal["INDIVIDUAL_ASV", "COMMON_ASV"]
+    type: AntiSurgeType | None = None
```
tj098895: allow None? Is that to also use anti_surge_config when it is not allowed, e.g. when there is no compressor? Would it not be better to have a specific enum for that?

src/libecalc/ecalc_model/process_simulation.py
```diff
@@ -38,19 +39,21 @@ class IndividualStreamDistributionConfig:
...
+
+
+@value_object
+class Constraint:  # should this instead be more flexible wrt. matching one or more stream conditions?
```
tj098895: remove comment, as it basically states what you have fixed :)

src/libecalc/presentation/yaml/mappers/process/build_sections.py
```diff
@@ -0,0 +1,118 @@
...
+from libecalc.process.process_units.choke import Choke
+from libecalc.process.process_units.compressor import Compressor
+from libecalc.process.process_units.direct_mixer import DirectMixer
+from libecalc.process.process_units.direct_splitter import DirectSplitter
+from libecalc.process.process_units.mixer import Mixer
+from libecalc.process.process_units.splitter import Splitter
+
+
+@value_object
+class WrappedSection:
```
tj098895: foretrekker at vi skriver navnene fullt ut isf. å fjerne kontekst-prefikset, som her og i `Unit` isf `ProcessUnit`. Det koster ikke nok ekstra til at det er verdt det :) Hodet mitt foretrekker å være konsekvent isf å spare bytes :)

Navnet her sier meg lite og ingenting, and strains my poor brain :(

src/libecalc/presentation/yaml/mappers/process/build_sections.py
```diff
@@ -0,0 +1,118 @@
...
+    index: int
```
tj098895: indekser er roten til alt vondt, bedre med dict el klasse som wrapper dette imo :)

src/libecalc/presentation/yaml/mappers/process/build_sections.py
```diff
@@ -0,0 +1,118 @@
...
+    process_units: list[ProcessUnit]
+    configuration_handlers: list[ConfigurationHandler]
+    pressure_control: str
```
tj098895: enum?

src/libecalc/presentation/yaml/mappers/process/build_sections.py
```diff
@@ -0,0 +1,118 @@
...
+    outlet_pressure: object
```
tj098895: Any? eller float eller liste av floats?

src/libecalc/ecalc_model/process_simulation.py
frodehk: Yes, the idea was that no anti surge should be specified for sections without compressors. But maybe this should be solved differently?

src/libecalc/presentation/yaml/mappers/process/build_sections.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def _wrap_compressor_in_recirculation_loop(group: list[ProcessUnit]) -> tuple[RecirculationLoop, list[ProcessUnit]]:
```
tj098895:
```suggestion
def _wrap_compressor_in_recirculation_loop(process_units: list[ProcessUnit]) -> tuple[RecirculationLoop, list[ProcessUnit]]:
```
el `pipeline`

src/libecalc/presentation/yaml/mappers/process/process_partitioner.py
```diff
@@ -0,0 +1,96 @@
...
+
+@value_object
+class ProcessSection:
+    """
+    A contiguous range of process units governed by one constraint.
+
+    Pure data: indices + units + the section's strategies/target. No wrapping.
+    """
+
+    index: int
```
tj098895: I would prefer that this is rewritten using `ProcessUnitID`s etc, instead of using indexes. This is so easily prone to errors (imo). And difficult to read and keep trakc of what is going on...

src/libecalc/presentation/yaml/mappers/process/process_section_validator.py
```diff
@@ -0,0 +1,61 @@
+from collections.abc import Sequence
+
+from libecalc.common.errors.ecalc_validation_error import EcalcValidationException
+from libecalc.presentation.yaml.mappers.process.process_partitioner import ProcessSection
+from libecalc.process.process_units.compressor import Compressor
+from libecalc.process.process_units.mixer import Mixer
+from libecalc.process.process_units.splitter import Splitter
+
+_ANTI_SURGE = ("COMMON_ASV", "INDIVIDUAL_ASV")
```
tj098895: I think I would prefer to also have `NO_ASV` - and that it might be the default. In that way it is easier to exhaustively gather all possible configurations and handle them similarily using `match` and be reassured that we have handle all cases.

src/libecalc/presentation/yaml/mappers/process/process_section_validator.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def validate(self, sections: Sequence[ProcessSection]) -> None:
+        for s in sections:
+            self._validate_strategy_compatibility(s)
+            self._validate_anti_surge_vs_compressors(s)
+            self._validate_pressure_control_vs_compressors(s)
+            self._validate_common_asv_units(s)
+
+    @staticmethod
+    def _validate_strategy_compatibility(s: ProcessSection) -> None:
```
tj098895: nice and important to validate and check combinations :)

we should though keep in mind, that this is only checks when using YAML, and if one uses the domain objects directly, we do not have these business rule checks there ...

src/libecalc/presentation/yaml/mappers/process/process_section_validator.py
```diff
@@ -0,0 +1,61 @@
...
+from libecalc.common.errors.ecalc_validation_error import EcalcValidationException
+from libecalc.presentation.yaml.mappers.process.process_partitioner import ProcessSection
+from libecalc.process.process_units.compressor import Compressor
+from libecalc.process.process_units.mixer import Mixer
+from libecalc.process.process_units.splitter import Splitter
+
+_ANTI_SURGE = ("COMMON_ASV", "INDIVIDUAL_ASV")
+_CHOKE = ("UPSTREAM_CHOKE", "DOWNSTREAM_CHOKE")
+
+_PRESSURE_CONTROL_REQUIRES_ANTI_SURGE = {
```
tj098895: should we consider using enums instead of literals? it is in general my preference, unless there are good arguments to not use. easier to make sure we have handled all cases etc

frodehk: Yes, I was thinking about it actually. Just continued to use the same as for PressureControlType. But I prefer enums myself.

src/libecalc/presentation/yaml/yaml_types/process/yaml_process_simulation.py
```diff
@@ -9,17 +9,41 @@
...
+from libecalc.process.process_solver.anti_surge.anti_surge_strategy import AntiSurgeType
+from libecalc.process.process_solver.pressure_control.pressure_control_strategy import PressureControlType
 
 
-class YamlProcessConstraints(YamlBase):
+class YamlProcessConstraint(YamlBase):
+    unit: Annotated[
+        ProcessUnitReference | None,
+        Field(
+            title="UNIT",
```
tj098895: why shorten to unit and make it less readable imo? :) Should we use the name `Reference` as well to emphasize that it is a reference? Just like the type/class? And, the reference leads of course to the ID/name of the process unit IN the pipeline - ie, the process unit is given a name in the pipeline, because that is when it is instantiated and used and becomes an entity. Do we check that the name that we give is unique?

src/libecalc/presentation/yaml/yaml_types/process/yaml_process_simulation.py
```diff
@@ -9,17 +9,41 @@
...
+        PressureControlType,
+        Field(
+            title="PRESSURE_CONTROL",
+            description="How to meet the target pressure at this constraint point.",
+        ),
+    ]
+    anti_surge: Annotated[
+        AntiSurgeType | None,
+        Field(
+            title="ANTI_SURGE",
```
tj098895, line 45: is it anti_surge, or `ANTI_SURGE_CONTROL`? to be consistent with pressure control?

src/libecalc/presentation/yaml/yaml_types/process/yaml_process_units.py
```diff
@@ -36,13 +37,15 @@ class YamlCompressor(YamlBase):
     """
 
     type: Literal["COMPRESSOR"]
+    name: ProcessUnitReference | None = None
```
tj098895: I think perhaps this is not what I had in mind. Here we given the process_unit_DEFINITION a name. This is just a recipe/building block which is a value object (same properties=same object), and can be used many times. The name here should then be a reference when choosing which building block to use, ie a predefined process unit. However, the name should be set when added to the pipeline, that is when it si given an ID ...right?

src/libecalc/process/process_solver/anti_surge/anti_surge_strategy.py
```diff
@@ -1,10 +1,13 @@
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
+from typing import Literal
 
 from libecalc.process.fluid_stream.fluid_stream import FluidStream
 from libecalc.process.process_solver.configuration import Configuration, RecirculationConfiguration
 from libecalc.process.process_solver.solver import Solution
 
+AntiSurgeType = Literal["INDIVIDUAL_ASV", "COMMON_ASV"]
```
tj098895:
```suggestion
AntiSurgeControlMechanism = Literal["INDIVIDUAL_ASV", "COMMON_ASV"]
```
?

src/libecalc/testing/process_builders.py
```diff
@@ -143,11 +143,16 @@ class YamlCompressorBuilder(Builder[YamlCompressor]):
     def __init__(self):
         self.type = "COMPRESSOR"
         self.compressor_model = None
+        self.name = "default_compressor"
```
tj098895: so here we will easily have duplicate names, if this is used...right?

src/libecalc/presentation/yaml/yaml_types/process/yaml_process_units.py
frodehk: I think I was a bit quick here. It should just be a str, right? ProcessUnitReference is when referring to a specific process unit, e.g in a train setup.

src/libecalc/presentation/yaml/mappers/process/build_sections.py
frodehk: Before I rename: Is it the name itself that should change, or the role of the class that is a bit unclear?

tj098895: both perhaps? but we can talk about it :)

src/libecalc/testing/process_builders.py
frodehk: Yes, moved from init now.

src/libecalc/presentation/yaml/yaml_types/process/yaml_process_simulation.py
frodehk: I was thinking ANTI_SURGE (control is implicit in anti surge/avoid surge), consistent with AntiSurgeStrategy vs PressureControlStrategy. But I am not sure what is best. 

If we should introduce AntiSurgeControl, I think we should consider to rename to AntiSurgeControlStrategy as well.

src/libecalc/presentation/yaml/yaml_types/process/yaml_process_simulation.py
frodehk: Changed to PROCESS_UNIT

src/libecalc/presentation/yaml/mappers/process/process_section_validator.py
frodehk: Created enum for anti surge. Waiting with pressure control, as it is more widely used.

src/libecalc/presentation/yaml/mappers/process/process_section_validator.py
frodehk: Created enum including NO_ASV

src/libecalc/presentation/yaml/mappers/process/process_partitioner.py
frodehk: Indexes removed

src/libecalc/presentation/yaml/mappers/process/build_sections.py
frodehk: Not relevant anymore, class is changed to have only have process units and configuration handlers.

src/libecalc/presentation/yaml/mappers/process/build_sections.py
frodehk: Yes, but I think it should be in a separate PR

src/libecalc/presentation/yaml/mappers/process/build_sections.py
frodehk: Removed, index not used.

src/libecalc/presentation/yaml/mappers/process/build_sections.py
frodehk: Edit: Class is renamed to AssembledSection - including the solver ready process units + handlers, for a given section.

src/libecalc/ecalc_model/process_simulation.py
frodehk: NO_ASV added - None not allowed

src/libecalc/presentation/yaml/yaml_types/process/yaml_process_units.py
```diff
@@ -36,13 +36,15 @@ class YamlCompressor(YamlBase):
     """
 
     type: Literal["COMPRESSOR"]
+    name: str | None = None
```
tj098895: imo it is not the process_units that should have name. they are process_unit_definitions - the recipe or spec for a process unit, that many instances can share.

where we need a name, is in the process_pipeline, where we specific which spec/definition to use, and if the process_unit has a constraint, then it will need a name that we can refer to in constraints ...

agree?

src/libecalc/ecalc_model/process_simulation.py
```diff
@@ -38,19 +39,22 @@ class IndividualStreamDistributionConfig:
...
 @value_object
 class AntiSurgeConfig:
-    type: Literal["INDIVIDUAL_ASV", "COMMON_ASV"]
+    type: AntiSurgeType
+
+
+@value_object
+class Constraint:
+    outlet_pressure: TimeSeriesExpression
+    pressure_control: PressureControlConfig
```
tj098895, line 55: just for the matter of asking; do we think it can be relevant to change pressure control and/or anti surge control mechanism over time; or should they be fixed?

src/libecalc/presentation/yaml/mappers/process/build_sections.py
```diff
@@ -0,0 +1,109 @@
...
+from libecalc.process.process_solver.recirculation_loop import RecirculationLoop
+from libecalc.process.process_units.choke import Choke
+from libecalc.process.process_units.compressor import Compressor
+from libecalc.process.process_units.direct_mixer import DirectMixer
+from libecalc.process.process_units.direct_splitter import DirectSplitter
+from libecalc.process.process_units.mixer import Mixer
+from libecalc.process.process_units.splitter import Splitter
+
+
+@value_object
```
tj098895, line 23: It seems that this is sth that we could move to domain, as an entry-point when creating domain objects, to make sure that the pipeline and problem is correctly set up. It is convenient to have a builder there (factory) that will create the proper domain objects if no domain policies/rules are violated :) what do you think?

src/libecalc/testing/process_builders.py
```diff
@@ -143,12 +143,18 @@ class YamlCompressorBuilder(Builder[YamlCompressor]):
     def __init__(self):
         self.type = "COMPRESSOR"
         self.compressor_model = None
+        self.name = None
```
tj098895: Hmm. these names here are incorrect now, right?

tj098895 (approved):
see my last comments :)

src/libecalc/presentation/yaml/mappers/process/build_sections.py
frodehk: Yes probably - I leave it here for now, but let us keep it in mind.

PR #1641, opened 2026-06-16
fix(libecalc): handle EOS failures in ShaftSpeedFinder speed boundary

olelod:
`CompressorThermodynamicCalculationError` (PH-flash failure) at a boundary speed is now handled by bisecting to the nearest thermodynamically feasible operating point. The feasible speed range is tightened from both ends — max-side via `highest_true`, min-side via the existing stonewall bisect — before the root-finding proceeds.

### DIFF
--- a/src/libecalc/process/process_solver/finders/shaft_speed_finder.py
+++ b/src/libecalc/process/process_solver/finders/shaft_speed_finder.py
@@ -1,6 +1,7 @@
 import logging
 from collections.abc import Callable
 
+from libecalc.domain.process.compressor.core.exceptions import CompressorThermodynamicCalculationError
 from libecalc.process.fluid_stream.fluid_stream import FluidStream
 from libecalc.process.process_pipeline.process_error import RateTooHighError, RateTooLowError
 from libecalc.process.process_solver.boundary import Boundary
@@ -8,6 +9,7 @@ from libecalc.process.process_solver.configuration import SpeedConfiguration
 from libecalc.process.process_solver.finder import Finder, Finding
 from libecalc.process.process_solver.search_strategies import Bisect, BisectResult, RootFindingStrategy
 from libecalc.process.process_solver.solver import (
+    EosFailure,
     RateTooHighFailure,
     RateTooLowFailure,
     TargetDirection,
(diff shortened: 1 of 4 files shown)


src/libecalc/process/process_solver/finders/shaft_speed_finder.py
```diff
@@ -31,15 +32,29 @@ def __init__(
...
         except RateTooLowError as e:
             logger.debug(f"No solution found for maximum speed: {max_speed_configuration}")
             return Finding(configuration=max_speed_configuration, failure=RateTooLowFailure.from_error(e))
+        except CompressorThermodynamicCalculationError:
+            logger.debug(
+                "EOS failure at max speed %.1f rpm; searching for highest EOS-valid speed.", self._boundary.max
+            )
+            valid_max = self._search_strategy.highest_true(self._boundary, lambda speed: self._eos_ok(func, speed))
+            if valid_max is None:
+                raise
```
olelod: If EOS fails at every candidate speed (`highest_true` returns `None`), we re-raise rather than returning a structured failure. This would indicate a fundamentally broken fluid model — all speeds thermodynamically invalid — which shouldn't be silently absorbed into a `Finding`.

jsolaas: What's the pattern for finders here? Finding if we have a configuration that makes sense to return, exception if not? Since failure is an option in finding I mean.

jsolaas (approved):

src/libecalc/process/process_solver/finders/shaft_speed_finder.py
olelod: Valid question. Looking at it more thoroughly, it seems CompressorThermodynamicCalculationError will end up unhandled, and result in a crash. I'll catch it in an EosFailure, but I never expect to see this one happen.

src/libecalc/process/process_solver/finders/shaft_speed_finder.py
```diff
@@ -31,15 +33,33 @@ def __init__(
...
             logger.debug(f"No solution found for maximum speed: {max_speed_configuration}")
             return Finding(configuration=max_speed_configuration, failure=RateTooLowFailure.from_error(e))
+        except CompressorThermodynamicCalculationError:
+            logger.debug(
+                "EOS failure at max speed %.1f rpm; searching for highest EOS-valid speed.", self._boundary.max
+            )
+            valid_max = self._search_strategy.highest_true(self._boundary, lambda speed: self._eos_ok(func, speed))
+            if valid_max is None:
+                return Finding(
+                    configuration=max_speed_configuration,
```
olelod, line 52: I guess you can't say that there is a "closest" configuration here. It's like a filler or placeholder here, but i don't think it is enough to change to `configuration: TConfiguration | None`.

PR #1644, opened 2026-06-17
fix: deprecation warning parameterize python

Aleksander-Karlsson:
Expanded the itertools combinations so we can update pytest


tj098895 (approved):
Wee, much cleaner :)

PR #1645, opened 2026-06-18
refactor: remove anti-surge config from PipelineSectionSolver

olelod:
The anti-surge configurations are already merged into Solution.configuration, so rates are readable directly from the returned solution.

After potential pressure control, the recirculation rates will change (also the contribution from flow/protection/anti-surge), so it does not make any sense to store them to use as a "floor" for the lowest recirculation either. 

If the goal is to track the contribution from anti-surge and from pressure control on the total recirculation it needs to be done another way.

### DIFF
--- a/tests/libecalc/process/process_solver/compressor_solver_path_matrix/single_stage_solver_path_matrix/test_process_solver.py
+++ b/tests/libecalc/process/process_solver/compressor_solver_path_matrix/single_stage_solver_path_matrix/test_process_solver.py
@@ -174,12 +174,14 @@ def test_process_solver_path(
         (c.value.delta_pressure for c in solution.configuration if isinstance(c.value, ChokeConfiguration)),
         None,
     )
-    anti_surge_solution = system.solver.get_anti_surge_solution()
-    anti_surge_recirculation_rates = (
-        tuple(c.value.recirculation_rate for c in anti_surge_solution.configuration)
-        if anti_surge_solution is not None
-        else ()
+
+    # Compute the anti-surge floor at the solution speed, without pressure control.
+    system.runner.apply_configurations(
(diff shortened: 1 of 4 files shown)


tests/libecalc/process/process_solver/compressor_solver_path_matrix/single_stage_solver_path_matrix/test_process_solver.py
```diff
@@ -174,12 +174,14 @@ def test_process_solver_path(
...
         None,
     )
-    anti_surge_solution = system.solver.get_anti_surge_solution()
-    anti_surge_recirculation_rates = (
-        tuple(c.value.recirculation_rate for c in anti_surge_solution.configuration)
-        if anti_surge_solution is not None
-        else ()
+
+    # Compute the anti-surge floor at the solution speed, without pressure control.
+    system.runner.apply_configurations(
```
olelod, line 179: This only makes sense because there is only one compressor

tests/libecalc/process/process_solver/test_pipeline_section_solver_with_common_asv.py
```diff
@@ -72,19 +72,21 @@ def test_pipeline_section_solver_with_common_asv(
...
 
-    recirculation_rate_at_capacity = recirculation_at_capacity_solution.configuration[0].value.recirculation_rate
-
     recirculation_configuration = [
         config for config in solution.configuration if isinstance(config.value, RecirculationConfiguration)
     ][0]
     recirculation_rate_after_pressure_control = recirculation_configuration.value.recirculation_rate
 
+    # Get the anti-surge floor at the solution speed, without pressure control.
+    runner.apply_configurations(
```
olelod, line 84: Same here. It only makes sense because there is only one recirculation loop.

jsolaas (approved):

PR #1650, opened 2026-06-19
refactor: typed exception boundary — process units raise, finders steer, solvers catch

olelod:


tests/libecalc/process/process_solver/test_find_valid_speed_boundary.py
```diff
@@ -168,8 +168,8 @@ def test_eos_failure_at_max_speed_with_real_fluid(fluid_service, search_strategy
     inlet = fluid_service.create_stream_from_standard_rate(
         fluid_model=fluid_model,
         pressure_bara=10.0,
-        standard_rate_m3_per_day=500_000.0,
-        temperature_kelvin=270.0,
+        standard_rate_m3_per_day=200_000.0,
```
olelod, line 171: This stream actually had liquid (raised LiquidAtInlet - and that's not what this test is supposed to test), so changed the rate and inlet temperature to avoid that.

src/libecalc/process/process_pipeline/process_error.py
```diff
@@ -8,34 +8,87 @@ def __init__(self, reason: str | None = None):
...
+        self,
+        process_unit_id: ProcessUnitId,
+        vapor_fraction: float,
+    ):
+        self.process_unit_id = process_unit_id
+        self.vapor_fraction = vapor_fraction
+        super().__init__(f"Inlet stream contains liquid (vapor fraction: {vapor_fraction:.3f})")
+
+
+class OfftakeExceedsInletError(ProcessError):
```
tj098895, line 66: So, it is officially called "offtake"?

src/libecalc/process/process_pipeline/process_error.py
```diff
@@ -8,34 +8,87 @@ def __init__(self, reason: str | None = None):
...
+            f"Compressor {process_unit_id} is below surge line"
+            + (
+                f": actual {actual_rate:.3f}, minimum {boundary_rate:.3f} m³/h."
+                if actual_rate and boundary_rate
+                else "."
+            )
+        )
+
 
+class CompressorStonewallError(ProcessError):
```
tj098895, line 33: I really like to make these errors even more domain specific :)

src/libecalc/process/process_solver/pipeline_section_solver.py
```diff
@@ -75,14 +77,35 @@ def find_solution(
         """
         Finds the speed and recirculation rates for each compressor to meet the pressure constraint.
         """
+        shaft_config = Configuration(
```
tj098895, line 86: is this shaft_config set implicitly?

src/libecalc/process/process_solver/solver.py
```diff
@@ -61,10 +70,81 @@ def from_error(cls, e: RateTooLowError) -> Self:
...
 
 
+@dataclass
+class ConvergenceFailure(SolverFailure):
+    reason: str = ""
+    source_id: ProcessPipelineId | None = None
+    lower_bound: float | None = None
+    upper_bound: float | None = None
+    tolerance: float | None = None
+    iterations: int | None = None
```
tj098895, line 84: Include how "far away" we were from a solution?

src/libecalc/process/process_solver/solver.py
```diff
@@ -61,10 +70,81 @@ def from_error(cls, e: RateTooLowError) -> Self:
...
 @dataclass
-class EosFailure(SolverFailure):
+class ThermodynamicCalculationFailure(SolverFailure):
     reason: str = ""
 
 
+@dataclass
+class ConvergenceFailure(SolverFailure):
+    reason: str = ""
+    source_id: ProcessPipelineId | None = None
```
tj098895, line 80: If a pipeline is split into sections, would we want to know "which section" that failed? Can we include a process unit id for now to specify "nearest known location where error occured"?

src/libecalc/process/process_units/liquid_remover.py
```diff
@@ -35,10 +34,10 @@ def propagate_stream(self, inlet_stream: FluidStream) -> FluidStream:
         if inlet_stream.vapor_fraction_molar < ThermodynamicConstants.PURE_VAPOR_THRESHOLD:
             new_fluid = self._fluid_service.remove_liquid(inlet_stream.fluid)
             inlet_molar_mass = inlet_stream.fluid.molar_mass
-            if inlet_molar_mass <= 0.0:
-                raise InvalidStreamException(
-                    f"Cannot remove liquid from a degenerate stream with non-positive molar mass: {inlet_molar_mass}"
-                )
+            assert inlet_molar_mass > 0.0, (
```
tj098895, line 37: The problem with this assert,is that it will never reach user. It will crash before that, unless we do actually gather those AssertionErrors and change to ProgrammingError and basically still give an error message, saying "It's not you, its us" etc :)

src/libecalc/process/process_units/splitter.py
```diff
@@ -29,6 +30,12 @@ def get_rate(self) -> float:
 
     def propagate_stream(self, inlet_stream: FluidStream) -> FluidStream:
         remaining_rate = inlet_stream.standard_rate_sm3_per_day - self._rate
+        if remaining_rate < 0.0:
```
tj098895, line 33: nice adding thse business rules :D

tj098895 (approved):
Nice. Just want to discuss the speed config that is set implicitly, and makes it very hard to read and understand how this is set. It was just a symptom here I guess, since you needed to return it :)

PR #1656, opened 2026-06-23
fix(libecalc): add zero-rate guard in PipelineSectionSolver

kjbrak:
## Type of Work

## What is this PR all about?

Adds a zero-rate guard at the top of `PipelineSectionSolver.find_solution()` - when inlet mass rate is zero, return success immediately with speed=0 instead of trying to solve. Matches the legacy solver's `_validate_positive_ingoing_rates` + `create_empty` pattern.

Resolves 10 matrix test xfails (R8 x5 single-stage, M6 x5 two-stage). Matrix results: 103 passed, 4 xfailed (was 93/14).

Two alternative zero-rate conventions were considered:
- **Return pressure=inlet_pressure** (no compression, stream passes through) - rejected because it breaks golden snapshot compatibility and the physics of "compressor off" vs "compressor as pass-through" is debatable
- **Return failure (NOT_CALCULATED)** - rejected because zero rate is not an error in production use; the compressor is simply off

## Between the lines?

The guard uses exact `== 0.0` comparison (no tolerance), matching legacy's strict `rate > 0` check. If a near-zero rate (e.g. 1e-15) arrives, it will still attempt to solve. This is consistent with legacy behavior.

### DIFF
--- a/src/libecalc/process/process_solver/pipeline_section_solver.py
+++ b/src/libecalc/process/process_solver/pipeline_section_solver.py
@@ -75,6 +75,16 @@ class PipelineSectionSolver:
         """
         Finds the speed and recirculation rates for each compressor to meet the pressure constraint.
         """
+        if inlet_stream.mass_rate_kg_per_h == 0.0:
+            return Solution(
+                configuration=[
+                    Configuration(
+                        configuration_handler_id=self._pipeline_section.shaft_id,
+                        value=SpeedConfiguration(speed=0.0),
+                    )
+                ]
+            )
+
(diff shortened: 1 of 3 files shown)


kjbrak:
PR closed. We will accept xfails on these 0 rate inlet cases, since running on only recycle is wanted behavior per now. There might be need for a thershold value spec on compressors for user to determine if compressor should shut down under a certain rate (or above a certain recycle/total_flow ratio). 

Also, for parallel trains in "compressor systems" (legacy name), a special handling of idle trains must be implemented where idle trains without inlet flow can stand idle without any recycle, perhaps still having success/is_valid true.

PR #1663, opened 2026-06-29
chore: upgrade dependencies for week 27. Auto-generated by devCalc

Aleksander-Karlsson:
Created issue to update Numpy to 2.5:
https://github.com/equinor/ecalc-internal/issues/1996


uv.lock
```diff
@@ -603,75 +603,65 @@ wheels = [
 
 [[package]]
 name = "numpy"
-version = "2.4.6"
+version = "2.5.0"
```
tj098895: Numpy 2.5.0 is a transitional release. It drops support for Python 3.11, marking the end of distutils, and expires a large number of deprecations made in the 2.0.x release. It also improves free threading and brings sorting into compliance with the array-api standard with the addition of descending sorts. There is also a fair amount of preparation for Python 3.15, which will be supported starting with the first rc.

This release supports Python versions 3.12-3.14.

tj098895: Should be ok though. Komodo is == 3.12 :)

tj098895 (approved):

tj098895 (changes requested):
Change numpy to < 2.5 for now since it is breaking, and create an issue to fix :)

tj098895 (approved):

PR #1665, opened 2026-07-02
chore: store process section info on ProcessProblem

frodehk:
This preserves the process section information produced during mapping, enabling support for multi-constraint support without rebuilding the section structure later.

Refs: equinor/ecalc-internal#2023

---

## Type of Work

### DIFF
--- a/src/libecalc/ecalc_model/process_simulation.py
+++ b/src/libecalc/ecalc_model/process_simulation.py
@@ -15,7 +15,7 @@ from libecalc.ecalc_model.time_series_configuration import (
 from libecalc.ecalc_model.time_series_stream import TimeSeriesStream
 from libecalc.presentation.yaml.domain.time_series_expression import TimeSeriesExpression
 from libecalc.process.process_pipeline.process_pipeline import ProcessPipelineId
-from libecalc.process.process_pipeline.process_unit import ProcessUnitId
+from libecalc.process.process_pipeline.process_unit import ProcessUnit, ProcessUnitId
 from libecalc.process.process_solver.anti_surge.anti_surge_strategy import AntiSurgeType
 from libecalc.process.process_solver.configuration_handler import ConfigurationHandler
 from libecalc.process.process_solver.pressure_control.pressure_control_strategy import PressureControlType
@@ -60,6 +60,15 @@ class Constraint:
 ProcessProblemId = NewType("ProcessProblemId", UUID)
 
 
+@value_object
+class ProcessProblemSection:
+    """A process section assembled for solver execution."""
+
+    process_units: list[ProcessUnit]
+    configuration_handlers: Sequence[ConfigurationHandler]
+    constraint: Constraint
+
+
 class ProcessProblem(Entity[ProcessProblemId]):  # TODO: Rename to subproblem?
     # can a problem exist wo. a simulation? yes, e.g. get max rate ...
     # given a physical pipeline (a contained problem, such as a compressor train), the user needs to define strategies to find a solution for the sub problem
@@ -67,14 +76,14 @@ class ProcessProblem(Entity[ProcessProblemId]):  # TODO: Rename to subproblem?
 
     def __init__(
         self,
-        constraints: list[Constraint],
+        sections: Sequence[ProcessProblemSection],
         configuration_handlers: Sequence[ConfigurationHandler],
         process_pipeline_id: ProcessPipelineId,
         process_problem_id: ProcessProblemId | None = None,
     ):
-        self.constraints = constraints
-        self.process_pipeline_id = process_pipeline_id
+        self.sections = sections
         self.configuration_handlers = configuration_handlers
+        self.process_pipeline_id = process_pipeline_id
         self._id: Final[ProcessProblemId] = process_problem_id or ProcessProblem._create_id()
 
     def get_id(self) -> ProcessProblemId:
@@ -85,7 +94,7 @@ class ProcessProblem(Entity[ProcessProblemId]):  # TODO: Rename to subproblem?
         return ProcessProblemId(ecalc_id_generator())
 
     def get_constraints(self) -> list[Constraint]:
-        return self.constraints
+        return [section.constraint for section in self.sections]
 
 
 ProcessSimulationId = NewType("ProcessSimulationId", UUID)
--- a/src/libecalc/ecalc_model/process_simulation.py
+++ b/src/libecalc/ecalc_model/process_simulation.py
@@ -76,12 +76,12 @@ class ProcessProblem(Entity[ProcessProblemId]):  # TODO: Rename to subproblem?
 
     def __init__(
         self,
-        sections: Sequence[ProcessProblemSection],
+        process_problem_sections: Sequence[ProcessProblemSection],
         configuration_handlers: Sequence[ConfigurationHandler],
         process_pipeline_id: ProcessPipelineId,
         process_problem_id: ProcessProblemId | None = None,
     ):
-        self.sections = sections
+        self.process_problem_sections = process_problem_sections
         self.configuration_handlers = configuration_handlers
         self.process_pipeline_id = process_pipeline_id
         self._id: Final[ProcessProblemId] = process_problem_id or ProcessProblem._create_id()
@@ -94,7 +94,7 @@ class ProcessProblem(Entity[ProcessProblemId]):  # TODO: Rename to subproblem?
         return ProcessProblemId(ecalc_id_generator())
 
     def get_constraints(self) -> list[Constraint]:
-        return [section.constraint for section in self.sections]
+        return [section.constraint for section in self.process_problem_sections]
 
 
 ProcessSimulationId = NewType("ProcessSimulationId", UUID)
--- a/src/libecalc/ecalc_model/process_simulation.py
+++ b/src/libecalc/ecalc_model/process_simulation.py
@@ -15,7 +15,7 @@ from libecalc.ecalc_model.time_series_configuration import (
 from libecalc.ecalc_model.time_series_stream import TimeSeriesStream
 from libecalc.presentation.yaml.domain.time_series_expression import TimeSeriesExpression
 from libecalc.process.process_pipeline.process_pipeline import ProcessPipelineId
-from libecalc.process.process_pipeline.process_unit import ProcessUnit, ProcessUnitId
+from libecalc.process.process_pipeline.process_unit import ProcessUnitId
 from libecalc.process.process_solver.anti_surge.anti_surge_strategy import AntiSurgeType
 from libecalc.process.process_solver.configuration_handler import ConfigurationHandler
 from libecalc.process.process_solver.pressure_control.pressure_control_strategy import PressureControlType
@@ -64,7 +64,7 @@ ProcessProblemId = NewType("ProcessProblemId", UUID)
 class ProcessProblemSection:
     """A process section assembled for solver execution."""
 
-    process_units: list[ProcessUnit]
+    process_unit_ids: list[ProcessUnitId]
     configuration_handlers: Sequence[ConfigurationHandler]
     constraint: Constraint
 
--- a/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
+++ b/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
@@ -366,7 +366,7 @@ class ProcessSimulationMapper:
                 )
                 process_problem_sections.append(
                     ProcessProblemSection(
-                        process_units=assembled_section.process_units,
+                        process_unit_ids=[u.get_id() for u in assembled_section.process_units],
                         configuration_handlers=assembled_section.configuration_handlers,
                         constraint=constraint,
                     )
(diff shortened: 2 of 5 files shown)


src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -353,36 +353,43 @@ def map_process_simulation(
...
+                        expression=mapped_section.constraint.outlet_pressure,
+                        expression_evaluator=self._expression_evaluator,
+                    ),
+                    pressure_control=PressureControlConfig(type=mapped_section.constraint.pressure_control),
+                    anti_surge=AntiSurgeConfig(mapped_section.constraint.anti_surge),
+                    target_process_unit_id=mapped_section.target_process_unit_id,
+                )
+                process_problem_sections.append(
+                    ProcessProblemSection(
+                        process_units=assembled_section.process_units,
```
tj098895: Why do we need process units in a section?

Is it a symptom that we need both a "physical" section, e.g. wrt solving it separately in core, but in sequence, as the core code suggests - and a "logical" section that says how it should be solved (like the problem we have now). However, a section is only needed wrt. solving - but you can say the same about a pipeline, right?

Lets think aloud tmrw :)

frodehk: We can store metadata only of course (names etc). This is basically why I suggested to introduce a new solver assembly layer - with the responsibility for sectioning, validating and assemble real sections for the solver - and from that build PipelineSections for the solvers, and select between PipelineSectionSolver and MultiPressureSolver.

frodehk: But it is a good question. Now we have process units two places. Reason I kept it was to avoid rebuilding. We have several options: solver assembly layer, or just say that ProcessProblemSection contains constraint, and reference to start and end process unit.

tj098895: I think a good trade-off and solution now is:
- Refer to the process units in a section with IDs - process unit IDs (the process units are still owned by the pipeline, they are physical and the pipeline is physical. Makes sense. We cannot have "2 thing own the same thing", then that is solved.
- If we for now keep the section logical and as part of the problem, it makes sense to associate it with ALL the process units in that section (if we decided that section is a physical thing (as well), then it would make more sense to say that a pipeline consists of sections rather than process units, and the process units are part of sections - to still fulfill the requirement of having "1 owner" that is aware of the internals and can manipulate them (the process units).

frodehk: I have stored the IDs instead, on the ProcessProblemSection - now we have physical units only in the pipeline.

src/libecalc/ecalc_model/process_simulation.py
```diff
@@ -60,21 +60,30 @@ class Constraint:
...
+
 class ProcessProblem(Entity[ProcessProblemId]):  # TODO: Rename to subproblem?
     # can a problem exist wo. a simulation? yes, e.g. get max rate ...
     # given a physical pipeline (a contained problem, such as a compressor train), the user needs to define strategies to find a solution for the sub problem
     # TODO: might have subproblems, or dependencies, but we may want to add those as problems that depend on each other and needs to be evaluated in a given order
 
     def __init__(
         self,
-        constraints: list[Constraint],
+        sections: Sequence[ProcessProblemSection],
```
tj098895: nice that you emphasize the importance of order - ie they need to be resolved in the given order

src/libecalc/ecalc_model/process_simulation.py
```diff
@@ -85,7 +94,7 @@ def _create_id(cls: type[Self]) -> ProcessProblemId:
         return ProcessProblemId(ecalc_id_generator())
 
     def get_constraints(self) -> list[Constraint]:
-        return self.constraints
+        return [section.constraint for section in self.sections]
```
tj098895: not sure if this is needed or should be exposed any more? a constraint is only relevant with a section, not "outside that context"?

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -353,36 +353,43 @@ def map_process_simulation(
...
-                )
 
-            configuration_handlers[process_pipeline.get_id()] = problem_configuration_handlers
             predefined_configurations[process_pipeline.get_id()] = problem_time_series_configurations
 
             process_pipeline_reference_to_id_map[item.name] = process_pipeline.get_id()
             process_pipelines.append(process_pipeline)
+            process_problems.append(
+                ProcessProblem(
+                    sections=process_problem_sections,
```
tj098895: I think we should aim to be consistent wrt nameing, instead of changing or arbitrarily skip the prefix, e.g. section, problem_section, process_problem_section. I easily forget or do the same mistake myself. Not sure if this was intentional?

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -353,36 +353,43 @@ def map_process_simulation(
...
+                        expression_evaluator=self._expression_evaluator,
+                    ),
+                    pressure_control=PressureControlConfig(type=mapped_section.constraint.pressure_control),
+                    anti_surge=AntiSurgeConfig(mapped_section.constraint.anti_surge),
+                    target_process_unit_id=mapped_section.target_process_unit_id,
+                )
+                process_problem_sections.append(
+                    ProcessProblemSection(
+                        process_unit_ids=[u.get_id() for u in assembled_section.process_units],
+                        configuration_handlers=assembled_section.configuration_handlers,
```
tj098895, line 370: I dont think we want or need to set config handlers on the section? It is atm. "a bit" independent, and either set it on the problem or the section, not both ...

tj098895: also ref 1 owner

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -353,36 +353,43 @@ def map_process_simulation(
...
+                    outlet_pressure=TimeSeriesExpression(
+                        expression=mapped_section.constraint.outlet_pressure,
+                        expression_evaluator=self._expression_evaluator,
+                    ),
+                    pressure_control=PressureControlConfig(type=mapped_section.constraint.pressure_control),
+                    anti_surge=AntiSurgeConfig(mapped_section.constraint.anti_surge),
+                    target_process_unit_id=mapped_section.target_process_unit_id,
+                )
+                process_problem_sections.append(
+                    ProcessProblemSection(
```
tj098895, line 368: A section would be an entity, and would require an ID then.

src/libecalc/process/process_solver/section_assembly.py
```diff
@@ -0,0 +1,76 @@
...
+from libecalc.process.process_solver.recirculation_loop import RecirculationLoop
+from libecalc.process.process_units.choke import Choke
+from libecalc.process.process_units.compressor import Compressor
+from libecalc.process.process_units.direct_mixer import DirectMixer
+from libecalc.process.process_units.direct_splitter import DirectSplitter
+from libecalc.process.process_units.mixer import Mixer
+from libecalc.process.process_units.splitter import Splitter
+
+
+@value_object
```
tj098895, line 17: Entity

src/libecalc/process/process_solver/section_assembly.py
```diff
@@ -0,0 +1,76 @@
+from libecalc.common.ddd import value_object
```
tj098895, line 1: nice, it makes sense to move the builders here and separete from mapper logic. In particular when we in a future we may want to set up wo. YAML.

src/libecalc/process/process_solver/section_assembly.py
```diff
@@ -0,0 +1,76 @@
...
+
+@value_object
+class AssembledSection:
+    """Solver-ready process units + handlers for a given process section."""
+
+    process_units: list[ProcessUnit]
+    configuration_handlers: list[ConfigurationHandler]
+
+
+def _wrap_compressor_in_recirculation_loop(
```
tj098895:
```suggestion
def recirculation_loop(
```

also, since it doesnt in particularily wrap a compressor but a list of process units.

if it is required that at least one of them is a compressor - we should check for that! :)

src/libecalc/ecalc_model/process_simulation.py
frodehk: Probably not. Remove?

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
frodehk: Agree that handlers should out of section, but maybe next step? Problem handlers (currently only shaft) ar different from section handlers (RecirculationLoops, Choke etc.).

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
frodehk: Agree, not intentional

src/libecalc/process/process_solver/section_assembly.py
frodehk: Nice, agree!

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
frodehk: Good point, but are we sure that a section should be an entity yet? My intention was that ProcessProblemSection is section metadata owned by ProcessProblem - just describing the sectioning result. But maybe I dont see the full picture, we may need to reference the section itself?

src/libecalc/process/process_solver/section_assembly.py
frodehk: See answer in mapper.

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
frodehk: In my view, configuration handlers are part of solver preparation rather than the problem definition(problem: what to solve, handlers: how to solve). That’s why I think it could be a natural next step?

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
tj098895: Currently a section is a boundary of solving that needs to be consistent. Previously we thought that was the pipeline/problem, but it turns out now that in some cases at least, we are able to solve a subproblem in isolation (=process problem section). That means that we treat it as an aggregate, and an aggregate needs to have a main entity - which can be itself. Also, the way Ole-Petter has written the code for MultiShaft and MultiPressure solvers - he requires such an aggregate, but he wrote it to accept what we currently had - pipelines. The new boundary is a pipeline section for independently solvable problems, which is what MultiShaft and MultiPressure solvers  - and SectionSolver should take in for now.

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
tj098895: configuration handlers are also some sort of aggregates where we need to keep a boundary and be reassured that the entities they encapsulate/govern are in sync. Whether it is being solved or just applied. It is therefore a part of the problem domain imo, and a clear aggregate candidate, and also being used as one atm.

src/libecalc/process/process_solver/section_assembly.py
```diff
@@ -0,0 +1,76 @@
...
+from libecalc.process.process_units.mixer import Mixer
+from libecalc.process.process_units.splitter import Splitter
+
+
+@value_object
+class AssembledSection:
+    """Solver-ready process units + handlers for a given process section."""
+
+    process_units: list[ProcessUnit]
+    configuration_handlers: list[ConfigurationHandler]
```
tj098895, line 22: We are raising a few interesting questions here, but as far as I can see you have pragmatically just handled those, which is good and ok, but it is also important to be aware and somehow document those. I am talking about;
- a common ASV is currently implemented as governing a process section. Is that correct? Idk, but it makes sense to have this limitation now. It makes it easier, and whether it should be more flexible or different we can postpone.
- we also assume chokes before or after a section (first thing or last thing). Is that ok? I assume yes, and at least for now. However, it seems to me that it makes it harder and harder to hide the section concept for users, since it is so dominant wrt. business logic that we apply - and automagically insert process units here and there. But again, it can wait :)

So, currently it makes sense that (some) configuration handlers are within a section, because they are only applicable and limited to control process units within a section - except from `Shaft`, which can be across sections. It makes sense to keep it like this for now, yes.

src/libecalc/process/process_solver/section_assembly.py
```diff
@@ -0,0 +1,76 @@
...
+
+def recirculation_loop(
+    process_units: list[ProcessUnit],
+) -> tuple[RecirculationLoop, list[ProcessUnit]]:
+    mixer, splitter = DirectMixer(), DirectSplitter()
+    loop = RecirculationLoop(mixer=mixer, splitter=splitter)
+    return loop, [mixer, *process_units, splitter]
+
+
+def assemble_process_section(
```
tj098895, line 33: at some point it is important that we have good tests constructing these sections, to make sure we understand all (edge) cases. Can wait, as it is WIP.

tj098895 (approved):
nice, can you look into handling sections rather than pipelines in core code?

PR #1666, opened 2026-07-03
feat(libecalc): add pump for incompressible liquid to process domain

kjbrak:
**Note:** this approach (pump as a `ProcessUnit`) is parked as a draft. The proposed implementation is now the isolated liquid subdomain in #1670.

## What is this PR all about?

Adds a pump for a simplified liquid stream to the new process domain.

- `SimplifiedLiquidStream`: a minimal value object with pressure, density and mass rate (no equation of state, composition or temperature).
- `Pump`: a `ProcessUnit` that takes an inlet liquid stream and a set discharge pressure and returns the outlet stream at that pressure (density and rate unchanged). Discharge pressure is a setpoint on the unit (`set_discharge_pressure`).

To let the pump be a `ProcessUnit` without coupling the liquid to the gas `FluidStream`, `StreamPropagator` and `ProcessUnit` are generalized over the stream type: gas units are `ProcessUnit[FluidStream]` and the pump is `ProcessUnit[SimplifiedLiquidStream]`. `ProcessPipeline` and the sequential `propagate_stream_many` are likewise generic, so a future liquid pipeline can reuse them. The generic type parameter keeps a pipeline homogeneous (all gas or all liquid). Gas behaviour is unchanged.

**Scope note:** only the `ProcessPipeline` container and the sequential `propagate_stream_many` helper are generalized. The stateful `ProcessPipelineRunner` / `ProcessRunner` are left gas-typed and are *not* reused for the pump - they carry configuration-handler and pressure-control coupling that a single pump does not need. A liquid pipeline would propagate via `propagate_stream_many`, not the solver runner.

The unit is additive and not yet wired into YAML or any consumer path. The existing (legacy) pump model is untouched.

Chart lookup, efficiency and power, minimum-flow recirculation, and feasibility handling are intentionally left out of this step; power and the result are expected to be owned by the coming energy domain.

- Initially made a functional pump with min-flow recirculation and minimum flow input. Similar to legacy pump but simpler, and some additional result properties related to recirculation. After feedback, reverted this and made a pass-through liquid stream propagator setting the specified outlet pressure. Pump specific logic presumably deferred to energy domain later.
- Keeping the pump as a separate liquid subdomain with its own propagator interface, leaving the gas code untouched. That is the approach we went with - now the proposal in #1670.

### DIFF
--- a/src/libecalc/process/process_pipeline/process_pipeline.py
+++ b/src/libecalc/process/process_pipeline/process_pipeline.py
@@ -9,9 +9,12 @@ from libecalc.process.process_pipeline.process_unit import ProcessUnit
 ProcessPipelineId = NewType("ProcessPipelineId", UUID)
 
 
-class ProcessPipeline(Entity[ProcessPipelineId]):
+class ProcessPipeline[TStream](Entity[ProcessPipelineId]):
     def __init__(
-        self, name: str, stream_propagators: Sequence[ProcessUnit], process_pipeline_id: ProcessPipelineId | None = None
+        self,
+        name: str,
+        stream_propagators: Sequence[ProcessUnit[TStream]],
+        process_pipeline_id: ProcessPipelineId | None = None,
     ):
         self._name = name
         self._stream_propagators = stream_propagators
@@ -23,7 +26,7 @@ class ProcessPipeline(Entity[ProcessPipelineId]):
     def get_name(self) -> str:
         return self._name
 
-    def get_process_units(self) -> list[ProcessUnit]:
+    def get_process_units(self) -> list[ProcessUnit[TStream]]:
         return list(self._stream_propagators)
 
     @classmethod
--- a/src/libecalc/process/pump/exceptions.py
+++ b/src/libecalc/process/pump/exceptions.py
@@ -1,0 +1,32 @@
+"""Exceptions for invalid liquid streams and pump configuration."""
+
+from libecalc.common.errors.exceptions import EcalcError
+
+
+class InvalidLiquidStreamException(EcalcError):
+    def __init__(self, title: str | None = None, reason: str | None = None):
+        super().__init__(title=f"Invalid liquid stream: {title or ''}", message=reason or "")
+
+
+class NonPositiveDensityException(InvalidLiquidStreamException):
+    """Raised when a liquid density is not strictly positive."""
+
+    def __init__(self, density_kg_per_m3: float):
+        self.density_kg_per_m3 = density_kg_per_m3
+        super().__init__(reason="Liquid density must be positive.")
+
+
+class NegativeMassRateException(InvalidLiquidStreamException):
+    """Raised when a liquid stream mass rate is negative."""
+
+    def __init__(self, mass_rate_kg_per_h: float):
+        self.mass_rate_kg_per_h = mass_rate_kg_per_h
+        super().__init__(reason="Mass rate cannot be negative.")
+
+
+class NonPositivePressureException(InvalidLiquidStreamException):
+    """Raised when a liquid pressure is not strictly positive."""
+
+    def __init__(self, pressure_bara: float):
+        self.pressure_bara = pressure_bara
+        super().__init__(reason="Pressure must be positive.")
--- a/src/libecalc/process/pump/liquid_stream.py
+++ b/src/libecalc/process/pump/liquid_stream.py
@@ -1,0 +1,73 @@
+"""An incompressible-liquid stream defined by pressure, density and mass flow rate.
+
+It carries no equation of state, composition or temperature. Because the liquid is
+incompressible, standard and actual volumetric rates coincide.
+"""
+
+from __future__ import annotations
+
+import dataclasses
+
+from libecalc.common.ddd import value_object
+from libecalc.common.units import UnitConstants
+from libecalc.process.pump.exceptions import (
+    NegativeMassRateException,
+    NonPositiveDensityException,
+    NonPositivePressureException,
+)
+
+
+@value_object
+class SimplifiedLiquidStream:
+    """An incompressible liquid stream.
+
+    Attributes:
+        pressure_bara: Stream pressure [bara].
+        density_kg_per_m3: Liquid density [kg/m3].
+        mass_rate_kg_per_h: Mass flow rate [kg/h].
+    """
+
+    pressure_bara: float
+    density_kg_per_m3: float
+    mass_rate_kg_per_h: float
+
+    def __post_init__(self):
+        if self.density_kg_per_m3 <= 0:
+            raise NonPositiveDensityException(self.density_kg_per_m3)
+        if self.mass_rate_kg_per_h < 0:
+            raise NegativeMassRateException(self.mass_rate_kg_per_h)
+        if self.pressure_bara <= 0:
+            raise NonPositivePressureException(self.pressure_bara)
+
+    @property
+    def volumetric_rate_m3_per_hour(self) -> float:
+        """Volumetric flow rate [m3/h]"""
+        return self.mass_rate_kg_per_h / self.density_kg_per_m3
+
+    @property
+    def volumetric_rate_m3_per_day(self) -> float:
+        """Volumetric flow rate [m3/day]"""
+        return self.volumetric_rate_m3_per_hour * UnitConstants.HOURS_PER_DAY
+
+    def with_mass_rate(self, mass_rate_kg_per_h: float) -> SimplifiedLiquidStream:
+        """Return a new stream with the same state but a different mass rate."""
+        return dataclasses.replace(self, mass_rate_kg_per_h=mass_rate_kg_per_h)
+
+    def with_pressure(self, pressure_bara: float) -> SimplifiedLiquidStream:
+        """Return a new stream at a different pressure (density unchanged, incompressible)."""
+        return dataclasses.replace(self, pressure_bara=pressure_bara)
+
+    @classmethod
+    def from_volumetric_rate(
+        cls,
+        volumetric_rate_m3_per_day: float,
+        pressure_bara: float,
+        density_kg_per_m3: float,
+    ) -> SimplifiedLiquidStream:
+        """Create a stream from an actual volumetric flow rate [m3/day]."""
+        mass_rate_kg_per_h = volumetric_rate_m3_per_day * density_kg_per_m3 / UnitConstants.HOURS_PER_DAY
+        return cls(
+            pressure_bara=pressure_bara,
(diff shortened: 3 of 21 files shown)


tj098895:
> ## What is this PR all about?
> 
> Adds a pump for an incompressible liquid stream to the new process domain (`src/libecalc/process/pump/`):
> 
>     * `LiquidStream`: a minimal value object with pressure, density and mass rate (no equation of state, composition or temperature).
> 
>     * `Pump`: a single-point unit that, given an inlet `LiquidStream` and a required discharge pressure, computes shaft power and feasibility from a pump chart (single- and variable-speed). Minimum continuous flow is a required input; min-head choking is optional; a head margin gives tolerance on the maximum head.
> 
>     * `PumpResult`: carries the inlet stream, power, required and operational head, required and operational discharge pressure, operating and recirculation rate, and a feasibility status.
> 
> 
> The unit is additive and not yet wired into YAML or any consumer path. The existing (legacy) pump model is untouched.
> ## What else did you consider?
> 
>     * Making the pump a `StreamPropagator` node in the gas process pipeline, or reusing the gas `FluidStream`. Rejected: a pump models a special simplified liquid, will not be a pipeline node, and its evaluation is a single forward calculation with no solver, so a small isolated `LiquidStream`-based unit is a better fit.
> 
>     * Deriving the minimum flow from the chart (legacy behaviour) instead of requiring it as input. Chose a required, user-provided minimum flow, validated to be at least the chart minimum rate.

nice summary. How can we compare/set up new YAML based on existing YAML, if we do not know the minimum flow?

src/libecalc/process/pump/exceptions.py
```diff
@@ -0,0 +1,22 @@
...
+
+class InvalidLiquidStreamException(EcalcError):
+    def __init__(self, title: str | None = None, reason: str | None = None):
+        super().__init__(title=f"Invalid liquid stream: {title or ''}", message=reason or "")
+
+
+class NonPositiveDensityException(InvalidLiquidStreamException):
+    """Raised when a liquid density is not strictly positive."""
+
+    def __init__(self, density_kg_per_m3: float):
```
tj098895, line 14: please set the `density_kg_per_m3` as a member variable, so that we can extract this information separately. Keep the message separate and without that information possibly, as with compressors? The reason can be potentially set as a default value based on what type of exception it is.

src/libecalc/process/pump/liquid_stream.py
```diff
@@ -0,0 +1,67 @@
...
+import dataclasses
+
+from libecalc.common.ddd import value_object
+from libecalc.common.units import UnitConstants
+from libecalc.process.pump.exceptions import NegativeMassRateException, NonPositiveDensityException
+
+
+@value_object
+class LiquidStream:
+    """An incompressible liquid stream.
```
tj098895, line 22: Limited to water as oil would require other properties?

src/libecalc/process/pump/liquid_stream.py
```diff
@@ -0,0 +1,67 @@
...
+        pressure_bara: Stream pressure [bara].
+        density_kg_per_m3: Liquid density [kg/m3].
+        mass_rate_kg_per_h: Mass flow rate [kg/h].
+    """
+
+    pressure_bara: float
+    density_kg_per_m3: float
+    mass_rate_kg_per_h: float
+
+    def __post_init__(self):
```
tj098895, line 34: nice and clear :)

src/libecalc/process/pump/liquid_stream.py
```diff
@@ -0,0 +1,67 @@
...
+        if self.density_kg_per_m3 <= 0:
+            raise NonPositiveDensityException(self.density_kg_per_m3)
+        if self.mass_rate_kg_per_h < 0:
```
tj098895, line 37: so mass rate can be 0? what does that mean/indicate?

src/libecalc/process/pump/liquid_stream.py
```diff
@@ -0,0 +1,67 @@
...
+            raise NegativeMassRateException(self.mass_rate_kg_per_h)
+
+    @property
+    def volumetric_rate_m3_per_hour(self) -> float:
+        """Volumetric flow rate [m3/h] (=Sm3/h for an incompressible liquid)"""
+        return self.mass_rate_kg_per_h / self.density_kg_per_m3
```
tj098895, line 45: so, here we may get 0 / n ...? Is that ok?

src/libecalc/process/pump/liquid_stream.py
kjbrak: It can be oil too, as long as we assume its incompressible, which we do if we are to use the liquid stream for an oil export pump also, which seems to be some use cases? or is it only water injectors? 

In the use case of oils with potentially higher viscosity than water, then the pump chart/curves must reflect the fluid that is used (i.e. they ust be made with the viscosity of the fluid that is used (viscosity is also quite temperature dependent))

src/libecalc/process/pump/liquid_stream.py
kjbrak: Yes the stream can have no rate, its standing still basically. But for the pump getting an inlet with no rate it means its not running (same as legacy).

src/libecalc/process/pump/liquid_stream.py
kjbrak: If mass rate is 0 then volumetric rate is 0 yes. I think thats ok.

src/libecalc/process/pump/liquid_stream.py
kjbrak: Same for water really, water at temperature of 20 degC will have different curves than water at temperature 100degC, they have quite different viscosities. This relationahip is known and there are probably corrections etc that can be used if needed.

Now its basically up to the user to specify a pump chart that corresponds to the case in question.

src/libecalc/process/pump/exceptions.py
kjbrak: Yes, I will sort it out.

src/libecalc/process/pump/pump.py
```diff
@@ -0,0 +1,301 @@
...
+    required_discharge_pressure_bara: float
+    operational_discharge_pressure_bara: float
+    failure_status: PumpFailureStatus
+
+    @property
+    def is_valid(self) -> bool:
+        return self.failure_status == PumpFailureStatus.NO_FAILURE
+
+
+class Pump(Entity[ProcessUnitId]):
```
tj098895: I am a bit disappointed to see that this goes against what we already discussed (and agreed upon?)

1. We want to break it down in smaller pieces
2. Even though "a process pipeline with solvers is overkill", we should aim for a similar (but simpler?) pattern
3. Here imo we mix potentially chart related functionality with pump. Check compressor and see how it is done there. The difference between a "compressor as a process unit" - where the focus and the context is just "what happens when the stream flows through it" - it makes it very easy to reason about and understand!
4. We only deal with data in process (that we care about for now); please do not include energy and details. We will see if we we want them later, and how to get them (event pattern or state pattern). So please remove PumpResult, and only return the resulting stream :)

kjbrak: Ok. I thought we agreed in the last meeting (not the first where things were a bit unclear) to implement a simplified subdomain for pump and not e.g. thread a generic stream type through all the gas/compressor domain logic or add solver logic to the pump.

I can make a draft of how it would look like if using stream propagator and removing results (including power).

but there are a few open questions and inconsistencies arising from this:
- If no pump solver (makes sense to not have a pump solver) then who will control the recirculation if not the pump itself, it would have no owner (compressor delegates recycle to the solver because recycle is iteratively found for pressure control i think)
- And if no power in the pump then why any recirculation, the only thing recirc affects is the power, not the outlet stream, the outlet stream is specified by the outlet pressure.

So power and recirc belong together and there is no pump solver for a single pump, so it seems a bit inconsistent.

If we go stream-only, the pump is basically a pass-through that sets the outlet pressure and has no pump logic at all for now - no chart, no recirc, no power.

We can talk about it, and I will think about it some more. No problem making it a thin pass-through, but then the implications should be clear.

kjbrak:
> > ## What is this PR all about?
> > Adds a pump for an incompressible liquid stream to the new process domain (`src/libecalc/process/pump/`):
> > ```
> > * `LiquidStream`: a minimal value object with pressure, density and mass rate (no equation of state, composition or temperature).
> > 
> > * `Pump`: a single-point unit that, given an inlet `LiquidStream` and a required discharge pressure, computes shaft power and feasibility from a pump chart (single- and variable-speed). Minimum continuous flow is a required input; min-head choking is optional; a head margin gives tolerance on the maximum head.
> > 
> > * `PumpResult`: carries the inlet stream, power, required and operational head, required and operational discharge pressure, operating and recirculation rate, and a feasibility status.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The unit is additive and not yet wired into YAML or any consumer path. The existing (legacy) pump model is untouched.
> > ## What else did you consider?
> > ```
> > * Making the pump a `StreamPropagator` node in the gas process pipeline, or reusing the gas `FluidStream`. Rejected: a pump models a special simplified liquid, will not be a pipeline node, and its evaluation is a single forward calculation with no solver, so a small isolated `LiquidStream`-based unit is a better fit.
> > 
> > * Deriving the minimum flow from the chart (legacy behaviour) instead of requiring it as input. Chose a required, user-provided minimum flow, validated to be at least the chart minimum rate.
> > ```
> 
> nice summary. How can we compare/set up new YAML based on existing YAML, if we do not know the minimum flow?

Later when we implement the logic i had in the first PR draft, where rate is clamped to max of minimum flow or leftmost edge of the pump chart, then setting minimum flow to the leftmost point of the min speed curve in the pump chart will give the same behavior and results as legacy (ie. the flow will get clamped to the chart min).

src/libecalc/process/process_pipeline/process_pipeline.py
```diff
@@ -9,9 +9,12 @@
 ProcessPipelineId = NewType("ProcessPipelineId", UUID)
 
 
-class ProcessPipeline(Entity[ProcessPipelineId]):
+class ProcessPipeline[TStream](Entity[ProcessPipelineId]):
```
tj098895, line 12: I agree with you, adding "generics" here and "forcing" a completely different stream to fit into this, makes it harder to understand and reason about, and potentially adds complexity wrt "checking that it is a compatible stream etc". Also, different types of complexity so worth keeping them separate as you say. :)

kjbrak:
Superseded by #1670

PR #1669, opened 2026-07-07
refactor: build pipeline sections

frodehk:
A few refactoring going on here:

1. We current set a hard limit that each pipeline section MUST have at least 1 compressor. This is most likely temporary, but required for now, as we do not support processing a section without.
2. A common `PipelineSolver` interface to gather the Solvers to a common interface, to make it easier to relate to and possible change. Early phase, and we are mostly dealing with separate solvers in different scenarios, but we need to prepare combining them etc. A trade-off is that a few solvers requires >1 pressure constraints/targets, while a few require ==1. Changed signature to use a list, which supports both, and it is the solver itself that is responsible to check that input fulfills its requirement.
3. Changed `PipelineSection` to an entity, as it does represent more or less an entity in DDD, it is also clearly a boundary that we currently solve almost independently (but at the same time as a part of the new `Pipeline` entity I introduced, which is currently the aggregate that needs the sections to be consistent and compatible), and only when they are - we can store the result in "1 transaction". Additionally, I make it clear that all the properties of a section needs to be fetched with `get_`. This is one step towards saying that those properties are read-only and should not be manipulated directly from external code, which is the current case (as the entity or aggregate, cannot control that the change is compatible, doesnt have any side-effects or makes the entity consistent)

Refs: equinor/ecalc-internal#2028

---

## Type of Work


src/libecalc/process/process_solver/process_problem_input.py
```diff
@@ -0,0 +1,50 @@
+"""
```
tj098895: In general you are right, a solution to not directly import something from e.g. another layer or module, either because of restrictions wrt. dependencies in the module OR direction of dependencies.

We do NOT want any relation between the "ephemeral mapping layer" and the process domain. The best way imo. to achieve this is to have a completely set of different objects, but that are mostly in sync wrt attributes, but sometimes not.

We should improve our naming, to make sure we do not do this mistake, ie. prefix with the purpose or module. For the mapping objects, we should prefix those with `ephemeral_*` maybe, and keep the domain objects clean wo. any prefix.

I have been seeing this several places last couple of days working on the section-thing, and thought we need to clean up to make it easier to understand and make sure we respect the boundaries :)

tj098895: so, tldr; we should not have these protocols or interfaces, but a separate set of objects imo :)

frodehk: Separate set of objects has replaced the protocols now.

tj098895 (approved):

PR #1670, opened 2026-07-07
feat(libecalc): add closed-form pump for incompressible liquids to process domain

kjbrak:
## What is this PR all about?

Adds a closed-form pump model for incompressible liquids to the new process domain
(`src/libecalc/process/pump/`).

`Pump.evaluate(inlet, discharge_pressure) -> PumpResult`:

- calculates the required head from the pressure duty and liquid density;
- applies minimum-flow recirculation from the configured and chart-derived limits;
- evaluates the chart operating point, efficiency, speed and feasibility;
- applies legacy minimum-head operation and exposes required and operating discharge pressure; and
- returns power, operating rate, recirculation, required and operating pressure/head, speed and
  feasibility.

The pump also implements `LiquidStreamPropagator`. `propagate_stream()` is a thin adapter
over `evaluate()` that returns the delivered stream at the calculated operating pressure. This keeps
the pump compatible with a possible future sequential liquid process system without making that
orchestration infrastructure part of this PR.

The implementation includes `LiquidStream`, pump-specific validation and exceptions, and tests
covering normal operation, recirculation, minimum-head operation, infeasible points, invalid
pressure/efficiency, identity and legacy power parity.

## Why is the pump evaluated directly?

For the current duty-driven model, rate, suction pressure, required discharge pressure and density
are inputs. Required head follows directly from `dP / rho`, and the corresponding operating point,
speed and minimum-flow correction are obtained from the chart. No iterative control setting or
thermodynamic state must be converged, so the compressor root-finding solver is not required.

This conclusion is intentionally scoped to the current single-pump model. A future coupled liquid
network, system curve or optimization problem could introduce a separate solving requirement.

## Why return a `PumpResult` instead of only a stream?

Correct stream propagation requires the chart-based operating point because the pump may operate at
a higher discharge pressure than requested. `propagate_stream()` must therefore perform the pump
evaluation rather than only copy the requested pressure onto the outlet stream.

That evaluation also determines chart efficiency and pump power. `LiquidStream` contains
pressure, density and mass rate, but no enthalpy or efficiency, so chart-derived shaft work cannot be
reconstructed from the outlet stream alone. The operating-point calculation should happen once in
the pump domain and its result should be available to downstream consumers rather than recalculated
in a solver or energy layer.

A future energy domain may consume pump power or specific shaft work from the process result and add
motor/VSD losses, aggregation or other energy-domain concerns. That does not require moving the
chart evaluation out of the pump.

## Why use a dedicated liquid stream?

A pump user supplies a bulk liquid density without composition or temperature. The gas
`FluidStream` represents an equation-of-state-backed thermodynamic state whose density is calculated
from composition, temperature and pressure. It cannot faithfully represent the pump input without
inventing information the user did not provide.

`LiquidStream` therefore models only pressure, constant density and mass rate. It is deliberately a
separate liquid type rather than a reduced gas stream.

## Design boundary

This PR does not introduce:

- a liquid solver or root-finding strategy;
- a liquid pipeline or runner;
- liquid mixer/splitter units or a structural recirculation loop;
- generalisation of the gas `ProcessUnit` hierarchy to support multiple stream types; or
- YAML, result-presentation or backend/frontend integration.

Those are separate decisions. A liquid pipeline and structural recirculation loop may be useful if
liquid equipment becomes composable, persisted process flowsheets. The propagator interface keeps
that option available, but the current single-pump calculation does not require that infrastructure.

The pump is not polymorphic with gas process units and cannot be inserted into the existing gas
pipeline or solver.

### DIFF
--- a/src/libecalc/process/pump/exceptions.py
+++ b/src/libecalc/process/pump/exceptions.py
@@ -1,0 +1,32 @@
+"""Exceptions for invalid liquid streams and pump configuration."""
+
+from libecalc.common.errors.exceptions import EcalcError
+
+
+class InvalidLiquidStreamException(EcalcError):
+    def __init__(self, title: str | None = None, reason: str | None = None):
+        super().__init__(title=f"Invalid liquid stream: {title or ''}", message=reason or "")
+
+
+class NonPositiveDensityException(InvalidLiquidStreamException):
+    """Raised when a liquid density is not strictly positive."""
+
+    def __init__(self, density_kg_per_m3: float):
+        self.density_kg_per_m3 = density_kg_per_m3
+        super().__init__(reason="Liquid density must be positive.")
+
+
+class NegativeMassRateException(InvalidLiquidStreamException):
+    """Raised when a liquid stream mass rate is negative."""
+
+    def __init__(self, mass_rate_kg_per_h: float):
+        self.mass_rate_kg_per_h = mass_rate_kg_per_h
+        super().__init__(reason="Mass rate cannot be negative.")
+
+
+class NonPositivePressureException(InvalidLiquidStreamException):
+    """Raised when a liquid pressure is not strictly positive."""
+
+    def __init__(self, pressure_bara: float):
+        self.pressure_bara = pressure_bara
+        super().__init__(reason="Pressure must be positive.")
--- a/src/libecalc/process/pump/liquid_stream.py
+++ b/src/libecalc/process/pump/liquid_stream.py
@@ -1,0 +1,72 @@
+"""A constant-density liquid stream defined by pressure, density and mass flow rate.
+
+It carries no equation of state, composition, temperature or viscosity.
+"""
+
+from __future__ import annotations
+
+import dataclasses
+
+from libecalc.common.ddd import value_object
+from libecalc.common.units import UnitConstants
+from libecalc.process.pump.exceptions import (
+    NegativeMassRateException,
+    NonPositiveDensityException,
+    NonPositivePressureException,
+)
+
+
+@value_object
+class LiquidStream:
+    """A liquid stream modelled with constant density.
+
+    Attributes:
+        pressure_bara: Stream pressure [bara].
+        density_kg_per_m3: Liquid density [kg/m3].
+        mass_rate_kg_per_h: Mass flow rate [kg/h].
+    """
+
+    pressure_bara: float
+    density_kg_per_m3: float
+    mass_rate_kg_per_h: float
+
+    def __post_init__(self):
+        if self.density_kg_per_m3 <= 0:
+            raise NonPositiveDensityException(self.density_kg_per_m3)
+        if self.mass_rate_kg_per_h < 0:
+            raise NegativeMassRateException(self.mass_rate_kg_per_h)
+        if self.pressure_bara <= 0:
+            raise NonPositivePressureException(self.pressure_bara)
+
+    @property
+    def volumetric_rate_m3_per_hour(self) -> float:
+        """Volumetric flow rate [m3/h]"""
+        return self.mass_rate_kg_per_h / self.density_kg_per_m3
+
+    @property
+    def volumetric_rate_m3_per_day(self) -> float:
+        """Volumetric flow rate [m3/day]"""
+        return self.volumetric_rate_m3_per_hour * UnitConstants.HOURS_PER_DAY
+
+    def with_mass_rate(self, mass_rate_kg_per_h: float) -> LiquidStream:
+        """Return a new stream with the same state but a different mass rate."""
+        return dataclasses.replace(self, mass_rate_kg_per_h=mass_rate_kg_per_h)
+
+    def with_pressure(self, pressure_bara: float) -> LiquidStream:
+        """Return a new stream at a different pressure while keeping density unchanged."""
+        return dataclasses.replace(self, pressure_bara=pressure_bara)
+
+    @classmethod
+    def from_volumetric_rate(
+        cls,
+        volumetric_rate_m3_per_day: float,
+        pressure_bara: float,
+        density_kg_per_m3: float,
+    ) -> LiquidStream:
+        """Create a stream from an actual volumetric flow rate [m3/day]."""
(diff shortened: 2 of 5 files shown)


src/libecalc/process/pump/exceptions.py
```diff
@@ -0,0 +1,32 @@
+"""Exceptions for invalid liquid streams and pump configuration."""
+
+from libecalc.common.errors.exceptions import EcalcError
+
+
+class InvalidLiquidStreamException(EcalcError):
```
tj098895, line 6: We did call the equivalent for compressors for "process_exception", but that was with the intention of including pumps etc. Now we may possibly go in a direction of calling it sth related to "compressor"? ...

Well, in that case, is this domain "pump" or "liquidstream". I assume, as you indicate, it is separate and it makes sense to have this super class though :)

src/libecalc/process/pump/liquid_stream.py
```diff
@@ -0,0 +1,73 @@
...
+from libecalc.common.units import UnitConstants
+from libecalc.process.pump.exceptions import (
+    NegativeMassRateException,
+    NonPositiveDensityException,
+    NonPositivePressureException,
+)
+
+
+@value_object
+class SimplifiedLiquidStream:
```
tj098895: nice, easy to understand, reason about and test independently :)

tj098895: Do you think there will be a difference between a `SimplifiedLiquidStream` and `LiquidStream` in the future? What else would be in a `LiquidStream` - visqousity? Composition?

tj098895: could we rename it to `LiquidStream` now since we do not need to distinguish yet?

src/libecalc/process/pump/liquid_stream_propagator.py
```diff
@@ -0,0 +1,8 @@
+import abc
+
+from libecalc.process.pump.liquid_stream import SimplifiedLiquidStream
+
+
+class LiquidStreamPropagator(abc.ABC):
```
tj098895, line 6: `LiquidStreamPropagator` or `SimplifiedLiquidStreamPropagator`? ;)

src/libecalc/process/pump/pump.py
```diff
@@ -0,0 +1,46 @@
+"""A single-point pump for an incompressible liquid.
```
tj098895: you probably already said, but what is a single-point pump vs "multi-point" (or zero-point?)

tj098895: incompressible liquid - is that an assumption we have for simplified liquid stream - that it does not contains gas and is therefore not compressible? is that a property of pump or stream or both?

src/libecalc/process/pump/pump.py
```diff
@@ -0,0 +1,46 @@
...
+
+from libecalc.common.ddd.entity import Entity
+from libecalc.common.utils.ecalc_uuid import ecalc_id_generator
+from libecalc.process.process_pipeline.process_unit import ProcessUnitId
+from libecalc.process.pump.exceptions import NonPositivePressureException
+from libecalc.process.pump.liquid_stream import SimplifiedLiquidStream
+from libecalc.process.pump.liquid_stream_propagator import LiquidStreamPropagator
+
+
+class Pump(Entity[ProcessUnitId], LiquidStreamPropagator):
```
tj098895, line 96: we should probably have a separate ID system for pump as well - to avoid dependency. It is all UUIDs so it doesnt really matter - but makes it easier to change or add behaviour etc later if needed

src/libecalc/process/pump/pump.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def set_discharge_pressure(self, discharge_pressure_bara: float) -> None:
+        if discharge_pressure_bara <= 0:
+            raise NonPositivePressureException(discharge_pressure_bara)
+        self._discharge_pressure_bara = discharge_pressure_bara
+
+    def propagate_stream(self, inlet_stream: SimplifiedLiquidStream) -> SimplifiedLiquidStream:
+        if self._discharge_pressure_bara is None:
+            raise ValueError("Discharge pressure not set. Call set_discharge_pressure first.")
+        return inlet_stream.with_pressure(self._discharge_pressure_bara)
```
tj098895: I like this, makes it really clear what this actually means for a pump - it is "that easy/simple".

However, as we will need details on the actual work (later/soon), we need to add that functionality later - but for the "stream concern, we only need to make sure that we update necessary stream properties based on the effect of the pump", which is straight forward logic here.

Similarily, as for compressor and compressor charts, it makes sense to introduce that with "one line" here when we add that. To get some other component to calculate what a changed pressure means in head, enthalpy etc - currently the pump chart component - which can then easily be tested separately wo a stream - just the params it needs :)

tj098895 (approved):
Love it! :)

src/libecalc/process/pump/liquid_stream.py
kjbrak: I was just thinking that Liquid stream was not quite accurate since the FluidStream might also be Liquid. And "simplified stream" sounds like a its simplified version of our fluid stream, which is also not quite accurate since it can probably only be used for pumps. its sort of a simplified liquid pump stream, then its just to pick a combination of these words that is clear and gives the most sense.

Yes, I also thought maybe we could have a more generic liquid stream in the future, like if modelling pipe flow or the ESP pump with viscosity dependent  model as you mentioned earlier. Then we might have wanted tmep, viscosity and maybe something else on the liquid stream.

src/libecalc/process/pump/pump.py
kjbrak: outdated, comments on a work in progress

src/libecalc/process/pump/pump.py
kjbrak: No this is outdated, we need to make the outlet stream with the "operational" discharge pressure, not "required" discharge pressure, which will be different in the cases where there is "min head choking", or for single speed where required head must be met by choking operational P discharge down to required P discharge.

PR #1675, opened 2026-07-16
chore: pinpoint outlet pressure target failures to connection

tj098895:
Previously, an outlet pressure target failures was pinpointed to a pipeline. We have now implemented support for intermediate pressure targets, which means that there are more places than one place (outlet pressure + intermediate pressures) where we are not able to reach a target pressure. We therefore need to pinpoint that more specifically. It will now specify the section id of where it failed, which can unambigiously also pinpoint the last process unit and its outlet as where it actually failed.

Refs: equinor/ecalc-internal#1992

## Type of Work

### DIFF
--- a/src/libecalc/process/process_solver/solver.py
+++ b/src/libecalc/process/process_solver/solver.py
@@ -17,7 +17,7 @@ from libecalc.process.process_pipeline.process_error import (
     OfftakeExceedsInletError,
     ProcessError,
 )
-from libecalc.process.process_pipeline.process_pipeline import ProcessPipelineId
+from libecalc.process.process_pipeline.process_pipeline import ProcessPipelineId, ProcessPipelineSectionId
 from libecalc.process.process_pipeline.process_unit import ProcessUnitId
 from libecalc.process.process_solver.configuration import (
     Configuration,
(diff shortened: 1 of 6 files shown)


markusrf (approved):

src/libecalc/process/process_solver/solver.py
```diff
@@ -157,9 +157,16 @@ class TargetPressureUnreachableFailure(SolverFailure):
     achievable_pressure_bara: float
     target_pressure_bara: float
     direction: TargetDirection
-    source_id: ProcessPipelineId | None = None  # TODO: Replace with failing process unit!
+    source_id: ProcessPipelineSectionId | None = (
```
kjbrak, line 160: Could we consider keeping both the section ID and target connection ID? The section identifies which solver scope failed, while the connection identifies where the constraint was violated. If the section ID remains the only value, ecalc-engine should map it to the target connection before exposing it to the UI; it currently classifies this failure as `ProcessPipeline`.

kjbrak (approved):

src/libecalc/process/process_solver/solver.py
tj098895: Good observation, and I agree.

Currently we do not have access to connections/inlets/outlets in core, or very inconveniently, which, without a bigger refactoring, will lead to a lot of patchy code with arbitrary data structures just to get that data. I think it will be beneficiary to have that data in core, as a part of the domain objects, at least it is getting clearer to me that we should have a notion about outlets (and inlets?) in core. However, it needs a bigger refactoring as said.

A temporary golden mean is, as you suggest, to store section id here, and get hold of the connection where we have easily access to it, in ecalc-engine/worker - which is exactly what I do :)

PR #1676, opened 2026-07-17
chore: get operational points for a stream through a compressor

tj098895:
Refs: equinor/ecalc-internal#2041

## Type of Work

### DIFF
--- a/src/libecalc/process/process_units/compressor.py
+++ b/src/libecalc/process/process_units/compressor.py
@@ -1,5 +1,6 @@
 from typing import Final
 
+from libecalc.common.ddd import value_object
 from libecalc.domain.process.compressor.core.train.utils.common import (
     RECIRCULATION_BOUNDARY_TOLERANCE,
     calculate_outlet_pressure_and_stream,
@@ -16,6 +17,13 @@ from libecalc.process.process_pipeline.process_error import (
 )
 from libecalc.process.process_pipeline.process_unit import ProcessUnit, ProcessUnitId
 from libecalc.process.process_solver.boundary import Boundary
+
+
+@value_object
+class OperationalPoint:
+    actual_rate_am3_h: float
+    polytropic_efficiency: float
+    polytropic_head_joule_per_kg: float
 
 
 class Compressor(ProcessUnit):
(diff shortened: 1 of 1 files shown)


markusrf (approved):

src/libecalc/process/process_units/compressor.py
```diff
@@ -18,6 +19,13 @@
 from libecalc.process.process_solver.boundary import Boundary
 
 
+@value_object
+class OperationalPoint:
+    actual_rate_am3_h: float
```
kjbrak, line 24: Often just write m3/h when its actual rate. Not so important, up to you. Since the variable name already contains actual rate i think it might read better to have `actual_rate_m3_h` (or `actual_rate_m3_per_h`

(If variable name was flow_rate, then i guess flow_rate_am3_h would make more sense to distinguish it from rate at some reference conditions, e.g. Sm3/h)

src/libecalc/process/process_units/compressor.py
```diff
@@ -53,24 +61,43 @@ def propagate_stream(self, inlet_stream: FluidStream) -> FluidStream:
...
+        operational_point: OperationalPoint = self.get_operational_point(inlet_stream=inlet_stream)
+
+        return calculate_outlet_pressure_and_stream(
+            polytropic_efficiency=operational_point.polytropic_efficiency,
+            polytropic_head_joule_per_kg=operational_point.polytropic_head_joule_per_kg,
+            inlet_stream=inlet_stream,
+            fluid_service=self._fluid_service,
+        )
+
+    def get_operational_point(self, inlet_stream: FluidStream) -> OperationalPoint:
```
kjbrak, line 73: The new `get_operational_point` method will return clamped values for efficiency and head if rate is out of range of chart, so since the method is public the caller might get just clamped values and no indication of what has happened and if its valid.
Consider using the `calculate_polytropic_head_and_efficiency_single_point` already on the chart, its returns:
~~~
        self.polytropic_head = polytropic_head
        self.polytropic_efficiency = polytropic_efficiency
        self.chart_area_flag = chart_area_flag
        self.is_valid = is_valid
~~~

Or if only used within `propagate_stream`, could make it a private method. Guess it would be ok since propagate_stream validates rate values within ranges before calling it, but probably want it as public method or?

src/libecalc/process/process_units/compressor.py
```diff
@@ -53,24 +61,43 @@ def propagate_stream(self, inlet_stream: FluidStream) -> FluidStream:
...
                 polytropic_head,
                 polytropic_efficiency,
             ) = self.compressor_chart.calculate_head_and_efficiency_for_internal_point_between_given_speeds(
                 speed=self.speed,
-                rate=actual_rate,
+                rate=actual_rate_am3_h,
             )
-
-        return calculate_outlet_pressure_and_stream(
+        return OperationalPoint(
```
kjbrak, line 97: Should shaft speed be in the operational point? To make he operational point be able to reproduce the calculation?

src/libecalc/process/process_units/compressor.py
tj098895: Good point. I currently want it as a public, and it is currently of no interest whether is inside (on) or outside the chart - but it will matter.

I would in general want to outsource this to the chart as you suggest, and as I wrote in the comment, but wanted to get some feedback before I possibly did that.

Btw, what do you mean by "clamped"? That head and efficiency cannot be trusted? If it is for debugging I guess it is ok, but if a user "chooses to accept invalid points", then this can be quite misleading if I understand you correctly?

src/libecalc/process/process_units/compressor.py
tj098895: speed is stored in db, associated with the compressor entity, same with inlet stream. Currently understood to separate, as the shaft speed is set at an earlier stage (after solving), while this is for simulation ...

src/libecalc/process/process_units/compressor.py
kjbrak: I think the interp1d is clamping rate value outside the valid range to the endpoints (max/min rate) when finding the corresponding efficiency and head, so it will be silently wrong for a caller that calls this method outside or the propagate_stream context (where it validates rate values before calling), IF the caller gives in invalid rate.

and when not returning chart area flag or is_valid bool also, then no way of knowing for the user what happened or if valid or not, perhaps..?

src/libecalc/process/process_units/compressor.py
kjbrak: hmm ok. not quite sure i understand the context here but ok. 

Would maybe have thought speed was an operational point or solution, and that only the chart itself was connected to the compressor entity. But i would need to look more at how it is set up to know what to think about this.

tj098895: you are right, its not 100%. speed is related to the shaft, but currently stored as a property of the compressor ...

anyways, I will look at dealing with "operational points when things fail" next, so your feedback is very useful wrt. that :)

PR #1679, opened 2026-07-20
1965 shea stadium

tj098895:
Quite simply adding an initial support for events (name not decided, WIP), to:
- add necessary metadata for WHEN a change to an eCalc model occurs and WHY
- We only specify WHEN the change occurs. It will last until there is another event that either cancels it or builds on top of it. Therefore, the END date will be different for different parts of the system. A change in rate due to a tie-in, can for instance result in higher rate for a certain compressor, which requires it to be upgraded by rebundle/revamp or a parallell configuration to handle the increased rate. If there is another change in the future, such as lower rate again, and different conditions, we may change the parallell train back to a serial train, which has to be a new event. Other parts of the system may be affected by the initial event of a tie-in, but may not be affected by the second, when for instance we go into LPP (low production period, ie tail production) 
- as a global reason for a change may not make it clear whether it will affect process or energy, we also want to specify that - so a global event can have an event in either domain or both. instead of setting the same time, it must refer to the global event, to ensure that dates are not mixed up, but being consistent throughout the model
- every small physical change, such as rebundle in this case, must be related to a process event as well. Why do we do the rebundle, and when? 
- such an event, at pipeline level, will be related to a very specific change, such as changing a compressor chart. We decide whether a rebundle/revamp should result in changing the compressor (new ID), or same compressor, but with new capability (same ID). This may depend on user needs, and perhaps different users sees this differently? Therefore it is important to add semantics to changes, in order to make it possible to make good decisions on how this should affect physical and logical changes in an ecalc model.

So, this is currently solved by "strangling" this capabiltiy into the process simulation mapper, and I think that I will change this to be a part of a "temporal pipeline" instead, where it is the pipeline that "makes decision" on how an event affects it over time. In order to do that, we need to separate the classes we use for temporal and core domain, OR wrap domain classes with a temporal generic; if all temporal changes should work the same independently of what it wraps. I will focus on improving that on next part that requires temporal aspect, e.g. changign from a parallell to a serial configuration.

## Type of Work

### DIFF
--- a/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
+++ b/src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
@@ -43,7 +43,9 @@ from libecalc.presentation.yaml.yaml_types.models import YamlFluidModel
 from libecalc.presentation.yaml.yaml_types.models.yaml_compressor_stages import YamlControlMarginUnits
 from libecalc.presentation.yaml.yaml_types.models.yaml_fluid import YamlCompositionFluidModel, YamlPredefinedFluidModel
 from libecalc.presentation.yaml.yaml_types.process.yaml_process_pipeline import YamlProcessPipeline
-from libecalc.presentation.yaml.yaml_types.process.yaml_process_references import ProcessUnitReference
+from libecalc.presentation.yaml.yaml_types.process.yaml_process_references import (
+    ProcessUnitReference,
+)
 from libecalc.presentation.yaml.yaml_types.process.yaml_process_simulation import (
     YamlEcalcEvent,
     YamlProcessEvent,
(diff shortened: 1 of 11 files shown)


src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -116,11 +125,23 @@ def __init__(
...
+        process_events: list[YamlProcessEvent] | None = None,
     ):
         self._expression_evaluator = expression_evaluator.get_subset_for_period(process_simulation_period)
         self._fluid_service = fluid_service
         self._reference_service = reference_service
         self._resources = resources
+        self._ecalc_events_by_name: dict[str, YamlEcalcEvent] = {e.name: e for e in (ecalc_events or [])}
+        self._process_events_by_name: dict[str, YamlProcessEvent] = {e.name: e for e in (process_events or [])}
+
+        # Validate that all PROCESS_EVENT REFs point to known ECALC_EVENTs
```
tj098895, line 138: so every process event MUST refer to an ecalc/global event - which is a operational/energy efficiency measurement or external or internal decision of a change - a process event comes as a result of changes that affects the process domain

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -234,6 +255,57 @@ def _map_fluid_model(self, yaml_fluid_model: YamlFluidModel) -> FluidModel:
...
+                raise EcalcValidationException(
+                    f"Pipeline event CHANGE_TARGET '{yaml_event.change_target}' does not match any named item "
+                    f"in pipeline '{yaml_pipeline.name}'. Available items: {', '.join([str(item.name) for item in yaml_pipeline.items])}."
+                )
+            #
+            if (change_to_yaml_process_unit := self._reference_service.get_process_unit(yaml_event.change_to)) is None:
+                raise EcalcValidationException(
+                    f"Pipeline event CHANGE_TO '{yaml_event.change_to}' does not refer to a known process unit."
+                ) from None
+            match change_to_yaml_process_unit:
```
tj098895, line 278: currently only rebundle and compressor change event supported

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -234,6 +255,57 @@ def _map_fluid_model(self, yaml_fluid_model: YamlFluidModel) -> FluidModel:
...
+                case YamlCompressor():
+                    change_to_unit = self._get_compressor(change_to_yaml_process_unit)
+                case _:
+                    raise EcalcValidationException(
+                        f"Pipeline event CHANGE_TO '{change_to_yaml_process_unit}' does not match any compressor (chart) unit."
+                    )
+
+            # Validate optional REF points to a known process event
```
tj098895, line 286: a local change, such as a rebundle MUST refer to a process event - why do we need to rebundle? is it due to external factors, or is it just to upgrade?

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -234,6 +255,57 @@ def _map_fluid_model(self, yaml_fluid_model: YamlFluidModel) -> FluidModel:
...
+            if process_event_ref is None or process_event_ref not in self._process_events_by_name:
+                raise EcalcValidationException(
+                    f"Pipeline event REF '{process_event_ref}' does not match any PROCESS_EVENT. "
+                    f"Available: {', '.join(sorted(self._process_events_by_name.keys()))}."
+                )
+            process_event = self._process_events_by_name[process_event_ref]
+            # NOTE: We have verified that all process events have correct refs already
+            ecalc_event = self._ecalc_events_by_name[process_event.ref]
+
+            events.append(
```
tj098895, line 297: transforming the yaml event, to get necessary data to actually apply the event on the pipeline, to transform it to the new one - with a rebundled compressor

src/libecalc/presentation/yaml/mappers/process_simulation_mapper.py
```diff
@@ -368,11 +440,19 @@ def map_process_simulation(
...
 
+            # TODO: Simply events for now ...
+            pipeline_events = self._validate_and_map_pipeline_events(
+                yaml_pipeline=item,
+                unit_name_to_id=unit_name_to_id,
+                process_unit_map=process_unit_map,
+            )
             # TODO: We should move this class to this module/layer
             # in particular because it creates necessary connections, which means that they will get new IDs
             process_pipeline = ProcessPipeline(
```
tj098895, line 451: for now, we send events to the caller - so the caller needs to handle this themself. This should probably be handled in this temporal/ephemeral domain instead, and create a pipeline with changes over time

markusrf (approved):

PR #1681, opened 2026-07-21
feat(libecalc): add closed-form liquid pump process simulation

kjbrak:
## What is this PR all about?

Adds a closed-form liquid pump as its own process graph in the new (experimental) process domain, under a dedicated `PUMP_PROCESS_SIMULATIONS` YAML keyword, kept separate from the gas pipeline.

**What is added**

- `Pump` (`libecalc/process/pump/`) is fully closed-form. Its result `PumpEvaluationResult` carries efficiency, specific shaft work, shaft power, speed, feasibility, the operating rate/head and the recirculation rate. Minimum-flow recirculation and minimum-head choking are computed directly - no solver.
- `PumpProcessSimulation` builds a fixed liquid process graph and evaluates a sequence of data-defined inputs (the domain is time-agnostic - it carries no `Period`):

  ```text
  inlet -> recirc mixer -> pump -> recirc splitter -> choke -> outlet
                ^                        |
                +--------- recycle ------+
  ```

  It exposes typed `LiquidProcessUnit` nodes (entities with ids), serial `ProcessUnitConnection` edges (reusing the shared process primitive), and a `LiquidRecirculationLoop` (its own id, splitter id + mixer id). `evaluate(...)` returns, for each input in order, the flat pump result plus a `LiquidStream` on every connection, keyed by connection id.
- New YAML `PUMP_PROCESS_SIMULATIONS: - TYPE: PUMP` (with `PUMP_MODEL`, `INLET`, `REQUIRED_DISCHARGE_PRESSURE`), its mapper, and `YamlModel.get_pump_process_simulations()`.

**How a consumer uses it**

```python
from libecalc.process.pump.pump import Pump
from libecalc.process.pump.liquid_stream import LiquidStream
from libecalc.process.pump.pump_process_simulation import (
    PumpProcessSimulation,
    PumpOperatingInput,
)

simulation = PumpProcessSimulation(
    pump=Pump(pump_chart=chart_data, minimum_flow_rate_m3_per_hour=277.0),
    name="water_injection",
)

results = simulation.evaluate(
    [
        PumpOperatingInput(
            inlet_stream=LiquidStream.from_volumetric_rate(
                volumetric_rate_m3_per_day=24_000.0,
                pressure_bara=10.0,
                density_kg_per_m3=1010.0,
            ),
            required_discharge_pressure_bara=200.0,
        ),
    ]
)
for result in results:
    result.pump_result.shaft_power_mw   # operating point: power, efficiency, head, speed
    result.connection_streams           # {connection_id: LiquidStream} for a graph diagram
```

The graph structure (stable for persistence and visualisation) is read separately:

```python
simulation.get_process_units()             # typed nodes with ids (inlet, mixer, pump, splitter, choke, outlet)
simulation.get_process_unit_connections()  # edges: (id, from_unit_id, to_unit_id)
simulation.get_recirculation_loop()        # (id, splitter_id, mixer_id)
```

**Capabilities and patterns supported**

- Stable, libecalc-owned ids on every unit and connection.
- Typed units aligned with a generic process-unit taxonomy (inlet, direct mixer, pump, direct splitter, choke, outlet).
- Intermediate `LiquidStream` on every connection, keyed by connection id.
- A recirculation-loop descriptor (splitter + mixer) mirroring how the compressor models minimum-flow recycle.
- Time-agnostic evaluation behind a single stable entry point; the period/time mapping is kept by the caller (the YAML mapper), not the domain.
- Closed-form evaluation (no solver, no root-finding, no repeated stream re-runs).

**Intended integration in a consuming application**

A consuming application (for example a persistence-and-worker backend that renders process diagrams) can treat the pump like any other process graph, because libecalc owns the graph and its ids. The application only maps and schedules; it never re-implements pump physics or topology:

```python
# `repo` is the consuming application's own persistence layer; not part of libecalc.
# The mapper returns the simulation, the ordered data-defined inputs, and the matching periods.
simulation, operating_inputs, periods = yaml_model.get_pump_process_simulations()[0]

# 1. Persist the graph once (structure and ids come from libecalc)
for unit in simulation.get_process_units():
    repo.save_unit(id=unit.get_id(), unit_type=unit.unit_type, name=unit.name)
for connection in simulation.get_process_unit_connections():
    repo.save_connection(
        id=connection.get_id(),
        from_unit_id=connection.get_from_process_unit_id(),
        to_unit_id=connection.get_to_process_unit_id(),
    )
repo.save_recirculation_loop(simulation.get_recirculation_loop())

# 2. A worker evaluates the inputs and stores the results, re-attaching the period by index
results = simulation.evaluate(operating_inputs)
for period, result in zip(periods, results, strict=True):
    for connection_id, stream in result.connection_streams.items():
        repo.save_stream(connection_id=connection_id, period=period, stream=stream)
    repo.save_operating_point(period=period, pump_result=result.pump_result)  # shaft power, efficiency, head, speed
```

Note: unit and connection ids are generated when the simulation is built. A consumer that needs ids to be stable across runs persists them once and reuses them, exactly as the compressor process graph is handled.

**Why it is shaped this way, and how it differs from the compressor**

The pump deliberately reuses the compressor process graph *contract* (typed units, connections, per-connection streams, a recirculation loop, all with ids) so a consumer can persist and visualise pumps and compressors with one pattern. It deliberately does *not* reuse the compressor *runtime*:

- The compressor needs a solver (shaft speed, anti-surge and pressure control are root-finding problems) and obtains intermediate streams by re-running the pipeline to each unit. The pump is closed-form: required head follows from the pressure rise and density, recirculation clamps the rate to the minimum flow, and choking clamps the head to the curve - all in one pass. Importing the solver and runner machinery would add complexity with no benefit.
- The compressor pipeline is built around a compressible `FluidStream` with composition, EoS and temperature. A liquid is modelled as an incompressible `LiquidStream` (pressure, density, mass rate). It cannot flow through the gas pipeline types, and the persisted gas stream shape carries gas-only fields. Liquid is therefore kept as a separate track rather than forced through the gas hierarchy or changing working compressor code.

There is no planned support for generic, user-composable liquid process systems - only this pump process and, soon, a pump-system process (`TYPE: PUMP_SYSTEM`) with pumps in parallel. If a composable liquid system is ever needed, the internal stream projection can be replaced by a real per-unit liquid runner behind the same public `evaluate(...)` contract, without callers changing. The `LiquidStream` and `LiquidStreamPropagator` abstractions already exist to support that.

- Return only a flat pump result. Rejected: a flat result cannot populate a process graph, so it cannot support the diagram and per-connection streams a consumer wants.
- Fuse the pump into the gas `FluidStream` pipeline. Rejected: the pipeline is gas-typed and solver-driven; this would change working compressor infrastructure for no pump benefit.
- A full executable liquid process system with its own runner and configuration handlers. Deferred: unnecessary for a fixed, closed-form topology. The `LiquidStreamPropagator` interface keeps that door open.

## Between the lines?

- The new process domain (exposed through the `PUMP_PROCESS_SIMULATIONS` YAML keyword) is experimental and not yet wired to energy output or a consumer; this PR is a contract-first building block validated by tests (unit propagation, graph structure, streams-per-connection, order-preserving evaluation, and end-to-end YAML mapping).
- Zero-rate periods are treated as the pump being off: they still produce a full result (zero power, outlet pressure equal to the inlet pressure, zero-flow streams). Suction pressure and density are required to be positive in every period (they describe the physical inlet fluid); the required discharge pressure is only required when the pump runs, and is relaxed for off periods.
- `HEAD_MARGIN` is not supported in the new pump process domain - a point above the chart's maximum head is flagged as infeasible (`ABOVE_MAXIMUM_HEAD_AT_RATE`) rather than snapped to the maximum, so a facility pump chart that sets a non-zero `HEAD_MARGIN` is rejected.
- No behaviour change for existing pump (energy) consumers or for the compressor domain.
- Legacy installation pump YAML (`ENERGY_USAGE_MODEL: TYPE: PUMP`) is untouched.

## Refs.:
Refs: equinor/ecalc-internal#2059

### DIFF
--- a/src/libecalc/presentation/yaml/mappers/pump_process_simulation_mapper.py
+++ b/src/libecalc/presentation/yaml/mappers/pump_process_simulation_mapper.py
@@ -1,0 +1,164 @@
+from libecalc.common.errors.ecalc_validation_error import EcalcValidationException
+from libecalc.common.errors.exceptions import InvalidResourceException
+from libecalc.common.time_utils import Period
+from libecalc.common.variables import ExpressionEvaluator
+from libecalc.domain.process.value_objects.chart.chart import ChartData
+from libecalc.domain.regularity import Regularity
+from libecalc.domain.resource import Resources
+from libecalc.presentation.yaml.domain.expression_time_series_flow_rate import ExpressionTimeSeriesFlowRate
+from libecalc.presentation.yaml.domain.expression_time_series_fluid_density import ExpressionTimeSeriesFluidDensity
+from libecalc.presentation.yaml.domain.expression_time_series_pressure import ExpressionTimeSeriesPressure
+from libecalc.presentation.yaml.domain.reference_service import ReferenceService
+from libecalc.presentation.yaml.domain.time_series_expression import TimeSeriesExpression
+from libecalc.presentation.yaml.mappers.charts.user_defined_chart_data import UserDefinedChartData
+from libecalc.presentation.yaml.yaml_types.facility_model.yaml_facility_model import (
+    YamlPumpChartSingleSpeed,
+)
+from libecalc.presentation.yaml.yaml_types.process.yaml_process_simulation import YamlPumpProcessSimulation
+from libecalc.process.pump.liquid_stream import LiquidStream
+from libecalc.process.pump.pump import Pump
+from libecalc.process.pump.pump_process_simulation import (
+    PumpOperatingInput,
+    PumpProcessSimulation,
+)
+
+
+class PumpProcessSimulationMapper:
+    def __init__(
+        self,
+        expression_evaluator: ExpressionEvaluator,
+        reference_service: ReferenceService,
+        resources: Resources,
+        process_simulation_period: Period,
+    ):
+        self._expression_evaluator = expression_evaluator.get_subset_for_period(process_simulation_period)
+        self._reference_service = reference_service
+        self._resources = resources
+
+    def map(
+        self,
+        yaml_process_simulation: YamlPumpProcessSimulation,
+    ) -> tuple[PumpProcessSimulation, list[PumpOperatingInput], list[Period]]:
+        chart_data = self._get_chart_data(yaml_process_simulation.pump_model.chart)
+        pump = Pump(
+            pump_chart=chart_data,
+            minimum_flow_rate_m3_per_hour=yaml_process_simulation.pump_model.minimum_flow_rate,
+        )
+        simulation = PumpProcessSimulation(
+            pump=pump,
+            name=yaml_process_simulation.name,
+        )
+
+        regularity = Regularity(
+            expression_evaluator=self._expression_evaluator,
+            target_period=self._expression_evaluator.get_period(),
+        )
+        rate = ExpressionTimeSeriesFlowRate(
+            time_series_expression=TimeSeriesExpression(
+                expression_evaluator=self._expression_evaluator,
+                expression=yaml_process_simulation.inlet.rate,
+            ),
+            regularity=regularity,
+        )
+        rate_values = rate.get_stream_day_values()
+        suction_pressure = ExpressionTimeSeriesPressure(
+            time_series_expression=TimeSeriesExpression(
+                expression_evaluator=self._expression_evaluator,
+                expression=yaml_process_simulation.inlet.pressure,
+            ),
+        )
+        discharge_pressure = ExpressionTimeSeriesPressure(
+            time_series_expression=TimeSeriesExpression(
+                expression_evaluator=self._expression_evaluator,
+                expression=yaml_process_simulation.required_discharge_pressure,
+            ),
+        )
+        density = ExpressionTimeSeriesFluidDensity(
+            time_series_expression=TimeSeriesExpression(
+                expression_evaluator=self._expression_evaluator,
+                expression=yaml_process_simulation.inlet.density,
+            )
+        )
+
+        suction_values = suction_pressure.get_values()
+        discharge_values = discharge_pressure.get_values()
+        density_values = density.get_values()
+
+        # The domain is time-agnostic: build one physical input per period and keep the period
+        # vector here in the presentation layer, paired with the inputs (and the results) by index.
+        periods = list(rate.get_periods())
+        operating_inputs = [
+            self._to_operating_input(
+                rate_value=rate_value,
+                suction_pressure_bara=suction,
+                required_discharge_pressure_bara=discharge,
+                density_kg_per_m3=density_value,
+            )
+            for rate_value, suction, discharge, density_value in zip(
+                rate_values,
+                suction_values,
+                discharge_values,
+                density_values,
+                strict=True,
+            )
+        ]
+        if len(periods) != len(operating_inputs):
+            raise ValueError("Pump period vector and input vector length mismatch.")
+        return simulation, operating_inputs, periods
+
+    def _to_operating_input(
+        self,
+        rate_value: float,
+        suction_pressure_bara: float,
+        required_discharge_pressure_bara: float,
+        density_kg_per_m3: float,
+    ) -> PumpOperatingInput:
+        # Suction pressure and density describe the physical inlet fluid; they must be positive
+        # whether or not the pump runs.
+        self._require_positive(suction_pressure_bara, "suction pressure [bara]")
+        self._require_positive(density_kg_per_m3, "inlet density [kg/m3]")
+        inlet_stream = LiquidStream.from_volumetric_rate(
+            volumetric_rate_m3_per_day=rate_value,
+            pressure_bara=suction_pressure_bara,
+            density_kg_per_m3=density_kg_per_m3,
+        )
+
+        if rate_value > 0:
+            self._require_positive(required_discharge_pressure_bara, "required discharge pressure [bara]")
+            discharge = required_discharge_pressure_bara
+        else:
+            # Pump off (zero rate): the required discharge is a meaningless duty target. Keep the
+            # user's value when it is a valid absolute pressure, otherwise fall back to the inlet
+            # pressure (the pump delivers no head, so the outlet equals the inlet).
+            discharge = (
+                required_discharge_pressure_bara if required_discharge_pressure_bara > 0 else suction_pressure_bara
+            )
+
+        return PumpOperatingInput(
+            inlet_stream=inlet_stream,
+            required_discharge_pressure_bara=discharge,
+        )
+
+    @staticmethod
+    def _require_positive(value: float, subject: str) -> None:
+        if value <= 0:
+            raise EcalcValidationException(f"Pump {subject} must be greater than 0; got {value}.")
+
+    def _get_chart_data(self, reference: str) -> ChartData:
+        model = self._reference_service.get_pump_model(reference)
+        if model.head_margin != 0.0:
+            raise EcalcValidationException(
+                "HEAD_MARGIN is not supported by the new pump process domain "
+                "(points above the maximum head are flagged infeasible instead of snapped to it)."
+            )
+        resource = self._resources.get(model.file)
+        if resource is None:
+            raise EcalcValidationException(f"Pump chart resource '{model.file}' was not found.")
+        try:
+            return UserDefinedChartData.from_resource(
+                resource,
+                units=model.units,
+                is_single_speed=isinstance(model, YamlPumpChartSingleSpeed),
+            )
+        except InvalidResourceException as error:
+            raise EcalcValidationException(str(error)) from error
--- a/src/libecalc/presentation/yaml/model.py
+++ b/src/libecalc/presentation/yaml/model.py
@@ -45,6 +45,7 @@ from libecalc.presentation.yaml.domain.time_series_collections import TimeSeries
 from libecalc.presentation.yaml.domain.time_series_resource import TimeSeriesResource
 from libecalc.presentation.yaml.mappers.component_mapper import EcalcModelMapper
 from libecalc.presentation.yaml.mappers.process_simulation_mapper import ProcessSimulationMapper
+from libecalc.presentation.yaml.mappers.pump_process_simulation_mapper import PumpProcessSimulationMapper
 from libecalc.presentation.yaml.mappers.variables_mapper import map_yaml_to_variables
 from libecalc.presentation.yaml.mappers.variables_mapper.get_global_time_vector import (
     InvalidEndDate,
@@ -68,6 +69,10 @@ from libecalc.presentation.yaml.yaml_validation_context import (
     YamlModelValidationContextNames,
 )
 from libecalc.process.process_pipeline.process_pipeline import ProcessPipeline
+from libecalc.process.pump.pump_process_simulation import (
+    PumpOperatingInput,
+    PumpProcessSimulation,
+)
 
 DEFAULT_START_TIME = datetime(1900, 1, 1)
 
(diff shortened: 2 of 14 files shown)


src/libecalc/presentation/yaml/mappers/pump_process_simulation_mapper.py
```diff
@@ -0,0 +1,129 @@
...
+        self._reference_service = reference_service
+        self._resources = resources
+
+    def map(
+        self,
+        yaml_process_simulation: YamlPumpProcessSimulation,
+    ) -> tuple[PumpProcessSimulation, PumpProcessInput]:
+        chart_data = self._get_chart_data(yaml_process_simulation.pump_model.chart)
+        chart = Chart(chart_data)
+        minimum_flow_rate = (
```
tj098895: inst min flow a part of the pump or pump chart? ie like control margin ?

for compressors it is currently kept together witht he chart, in order to manipulate on the chart. it does probably not change at the same rate as the chart, but that would make it consistent for now at least? Also, to encapsulate the Chart (or PumpChart) within the pump perhaps?

src/libecalc/presentation/yaml/mappers/pump_process_simulation_mapper.py
```diff
@@ -0,0 +1,129 @@
...
+            yaml_process_simulation.pump_model.minimum_flow_rate
+            if yaml_process_simulation.pump_model.minimum_flow_rate is not None
+            else chart.minimum_rate
+        )
+        simulation = PumpProcessSimulation(
```
tj098895, line 47: like here, a PumpProcessSimulation takes a Pump as input perhaps?

src/libecalc/presentation/yaml/mappers/pump_process_simulation_mapper.py
```diff
@@ -0,0 +1,129 @@
...
+            validation_mask=pressure_validation_mask,
+        )
+        discharge_pressure = ExpressionTimeSeriesPressure(
+            time_series_expression=TimeSeriesExpression(
+                expression_evaluator=self._expression_evaluator,
+                expression=yaml_process_simulation.required_discharge_pressure,
+            ),
+            validation_mask=pressure_validation_mask,
+        )
+        density = ExpressionTimeSeriesFluidDensity(
```
tj098895, line 76: no validation mask here?

src/libecalc/presentation/yaml/mappers/pump_process_simulation_mapper.py
```diff
@@ -0,0 +1,129 @@
...
+            time_series_expression=TimeSeriesExpression(
+                expression_evaluator=self._expression_evaluator,
+                expression=yaml_process_simulation.inlet.density,
+            )
+        )
+
+        periods = list(rate.get_periods())
+        period_inputs = tuple(
```
tj098895: i dont see the tuple, it is just `PumpProcessPeriodInput` here right?

src/libecalc/presentation/yaml/mappers/pump_process_simulation_mapper.py
```diff
@@ -0,0 +1,129 @@
...
+            )
+        )
+        return (
+            simulation,
+            PumpProcessInput(period_inputs=period_inputs),
+        )
+
+    def _get_chart_data(self, reference: str) -> ChartData:
+        model = self._reference_service.get_pump_model(reference)
+        if model.head_margin != 0.0:
```
tj098895, line 149: temporary or permanently?

src/libecalc/presentation/yaml/model.py
```diff
@@ -155,6 +164,8 @@ def get_process_simulations(self) -> tuple[list[ProcessPipeline], list[ProcessSi
         )
         process_pipelines = []
         for yaml_process_simulation in self._configuration.process_simulations:
+            if not isinstance(yaml_process_simulation, YamlProcessSimulation):
```
tj098895: temporary? remove?

tj098895: ah, perhaps make a new root element for now, where you just call it `PUMP_PROCESS_SIMULATIONS` to keep them separate?

src/libecalc/process/pump/pump.py
```diff
@@ -32,28 +32,29 @@
...
-class PumpResult:
+class PumpEvaluationResult:
     """Result of a single pump evaluation.
 
     Attributes:
         inlet_stream: The liquid stream at suction conditions the pump was evaluated for
             (carries suction pressure, density and requested rate).
-        power_mw: Shaft power demand [MW].
+        shaft_power_mw: Shaft power [MW].
+        efficiency: Pump efficiency at the operating point, ``None`` when not running.
```
tj098895, line 49: so efficiency is None, but shaft_power_mw is 0 when not running ie rate == 0?

src/libecalc/process/pump/pump_process_simulation.py
```diff
@@ -0,0 +1,248 @@
...
+class LiquidProcessUnitType(StrEnum):
+    INLET = "INLET"
+    DIRECT_MIXER = "DIRECT_MIXER"
+    PUMP = "PUMP"
+    DIRECT_SPLITTER = "DIRECT_SPLITTER"
+    CHOKE = "CHOKE"
+    OUTLET = "OUTLET"
+
+
+@value_object
```
tj098895: it has an ID, hence it is an entity - it is what we care about wrt lifecycle, change of state etc

src/libecalc/process/pump/pump_process_simulation.py
```diff
@@ -0,0 +1,248 @@
...
+    id: ProcessUnitConnectionId
+    splitter_id: ProcessUnitId
+    mixer_id: ProcessUnitId
+
+
+@value_object
+class PumpProcessPeriodInput:
+    """The temporal inputs for one period of a pump process simulation."""
+
+    period: Period
```
tj098895: period should not be a part of this domain, it is handled and remvoed before sent to the simulation

src/libecalc/process/pump/pump_process_simulation.py
```diff
@@ -0,0 +1,248 @@
...
+        pump_id: Identity of the pump unit; generated when not provided.
+    """
+
+    def __init__(
+        self,
+        chart: ChartData,
+        minimum_flow_rate_m3_per_hour: float,
+        name: str = "pump",
+        process_simulation_id: PumpProcessSimulationId | None = None,
+        pump_id: ProcessUnitId | None = None,
```
tj098895: here we should have an instance of a Pump as parameter (with encapsulated chart, min flow and id) :)

src/libecalc/process/pump/pump_process_simulation.py
```diff
@@ -0,0 +1,248 @@
...
+        delivered_stream_after_choke = requested_stream.with_pressure(delivered_pressure_bara)
+
+        streams = (
+            requested_stream,
+            operating_stream_at_suction,
+            operating_stream_after_pump,
+            delivered_stream_before_choke,
+            delivered_stream_after_choke,
+        )
+        return MappingProxyType(
```
tj098895, line 253: why MappingProxyType and not dict? Is it sth related to serialization?

tests/libecalc/presentation/yaml/mappers/test_pump_process_simulation_mapper.py
```diff
@@ -0,0 +1,49 @@
...
+from libecalc.presentation.yaml.model import YamlModel
+from libecalc.presentation.yaml.yaml_entities import ResourceStream
+
+
+def test_pump_process_simulation_maps_and_evaluates(simple_yaml, configuration_service_factory):
+    yaml_text = (
+        simple_yaml.main_file.read()
+        + """
+
+PROCESS_SIMULATIONS:
```
tj098895: nice. can you add an example yaml in web in order for the rest of the team to have access to and use?

tj098895 (approved):
nice, please see my comments.

There are many simplifictions I want to make, both here and in existing code, but worth to fix soon instead.

Also, I see that it is hard to know what is the boundary of what is going on in core vs backend, and I want to clarify. We can talk later again :)

src/libecalc/presentation/yaml/mappers/pump_process_simulation_mapper.py
kjbrak: tuple comprehension

src/libecalc/presentation/yaml/mappers/pump_process_simulation_mapper.py
kjbrak: removing validation mask, and validating per period directly soon

src/libecalc/process/pump/pump.py
kjbrak: Yes, efficiency not defined when pump is not running, and power 0, and yes pump not running when rate is 0.

src/libecalc/process/pump/pump_process_simulation.py
kjbrak: `MappingProxyType` returns a read-only "view" so a caller cannot mutate the
streams after construction, preserving the object's immutability (not only "shallow" field level immutability of the value_object/frozen dataclass. A `dict` would satisfy the `Mapping` type but leave the contents mutable.

Could consider frozendict, but its not a dependency so maybe `MappingProxyType` is the best choice here. For serialization it needs dict(...) at the boundary.

src/libecalc/presentation/yaml/mappers/pump_process_simulation_mapper.py
kjbrak: I dont think it should be a part of the chart, because it can be viewed as an operational setting set by the operator of the pump deterining the recirc flow floor. I am thinking the chart is a spec on the pump but the the min flow is an operational contraint on top.

src/libecalc/presentation/yaml/mappers/pump_process_simulation_mapper.py
kjbrak: Permanently I think. If we allow head margin spec then we either have one of these two evils:
- Return valid calc, but with power corresponding to (clamped) max head and operational discharge pressure below required discharge pressure. This would give an outlet stream with pressure lower than what the user specified, but would still have valid calc..
- Or (as legacy did): valid calc and power at max head, but with operational discharge pressure equal to required discharge pressure which decouples the power from the actual pressure delivered, which is a violation of the energy balance (i.e. some pressure has magically been created without the pump power accounting for it).

I think its better to not support it, and rather just give the power at the required discharge pressure even if its above max head, but then have `failure_status = ABOVE_MAXIMUM_HEAD_AT_RATE`  ( is_valid = False ).
- They would then get the full power for that case but can choose to use it or not (it would be invalid)
- If they want the max head power instead they would need to reduce their delta pressure somewhat.
- And: if the user wants a valid point for this case, they would have to extend the pump chart max head values.

They cant get both power not matching dP AND a pressure not connected to that power, doesnt make sense and I think legacy is wrong here.

src/libecalc/presentation/yaml/model.py
kjbrak: changed to `PUMP_PROCESS_SIMULATIONS`

src/libecalc/presentation/yaml/mappers/pump_process_simulation_mapper.py
tj098895: I think that it makes sense to question this, as you suggest. And to postpone. I agree that we should strive for correctness and validity, and see how far that will get us. From experience and knowledge, it is easy to ignore those facts - and we shouldnt.

src/libecalc/process/pump/pump.py
tj098895: I think this shows that it is important to group data correctly, otherwise we end up with a lot of data that needs to be set to "something" and support many different states - that doesnt make sense to support. I think we will get back to this wrt. conditions and states - a user NEEDS to define what "rate == 0" means, and what that should lead to; e.g. bypass or shutdown etc.

PR #1708, opened 2026-08-07
feat: extract expressions functionality

jsolaas:


src/libecalc/expression/extract_expressions.py
```diff
@@ -0,0 +1,174 @@
...
+from pydantic import BaseModel
+
+from libecalc.expression.expression import Expression, ExpressionType
+from libecalc.presentation.yaml.yaml_types.components.yaml_expression_type import _ExpressionMarker
+
+
+def _is_expression_type(annotation: Any) -> bool:
+    """Return True if *annotation* is ``Annotated[..., EXPRESSION_MARKER, ...]``."""
+    if typing.get_origin(annotation) is typing.Annotated:
+        return any(isinstance(m, _ExpressionMarker) for m in typing.get_args(annotation)[1:])
```
tj098895: skipping first because that is the "real type hint"? what if there are 0 or 1, can it happen? Would be ok to search in args directly instead of specifying `[1:]`?

tj098895: use `EXPRESSION_MARKER` instead?

src/libecalc/expression/extract_expressions.py
```diff
@@ -0,0 +1,174 @@
...
+        return True
+
+    origin = typing.get_origin(annotation)
+
+    # Union / Optional
+    if origin is typing.Union:
+        return any(_contains_expression_type(arg) for arg in typing.get_args(annotation))
+
+    # types.UnionType (X | Y) – Python 3.10+
+    if isinstance(annotation, types.UnionType):
```
tj098895: move up, since it is "the same as Union"? Just a different way of writing in python 3.10+ ...

src/libecalc/expression/extract_expressions.py
```diff
@@ -0,0 +1,174 @@
...
+    # list[T]
+    if origin is list:
+        args = typing.get_args(annotation)
+        return bool(args) and _contains_expression_type(args[0])
+
+    # Annotated[T, ...] where T itself might contain expressions (but the
+    # metadata didn't match validate_expression – e.g. Field wrapping)
+    if origin is typing.Annotated:
+        return _contains_expression_type(typing.get_args(annotation)[0])
+
```
tj098895: what if dict? or none above makes it, log type, so that we notice if there are some cases we dont match?

src/libecalc/expression/extract_expressions.py
```diff
@@ -0,0 +1,174 @@
...
+        for arg in args:
+            result.extend(_unwrap_to_model_types(arg))
+        return result
+
+    return []
+
+
+def _has_expression_marker(metadata: list[Any]) -> bool:
+    """Return True if *metadata* contains an ``_ExpressionMarker``."""
+    return any(isinstance(m, _ExpressionMarker) for m in metadata)
```
tj098895: `EXPRESSION_MARKER`?

src/libecalc/expression/extract_expressions.py
jsolaas: We are specifically trying to check for `YamlExpressionType` so we could probably check a specific index. I'm assuming the order in Annotation is kept.

Searching all args would also work

src/libecalc/expression/extract_expressions.py
tj098895: or would it be `type(EXPRESSION_MARKER)`

tj098895 (approved):
Seems like a smart way to label and detect expressions in the model without "searching for matching strings etc", but does it cover all cases? Log in cases where we expect a match, but we dont, instead of just falling back to False and similar?

In general, I think this looks contained and controlled, which imo is the right way to use AI ... (as I see it now)

src/libecalc/expression/extract_expressions.py
jsolaas: Yes, checking for that specific instance instead

PR #1715, opened 2026-08-12
fix: prevent mass loss for supercritical fluids in liquid remover

olelod:
NeqSim's flash mislabels supercritical CO₂ as liquid (`vapor_fraction=0`). The LiquidRemover then discards 100% of mass.

This happens e.g. in CO₂ injection compressor trains when intercooled stage pressures exceed a temperature-dependent threshold (e.g. ~74 bar at 31°C, ~290 bar at 35°C, ~560 bar at 80°C).

**Fix:** When `vapor_fraction ≤ 0.0001`, compute the EoS critical point (`criticalPointFlash`, cached per composition). If the stream is above it — supercritical, pass through unchanged. If below — genuinely liquid, raise `NoGasPhaseError` to avoid infinite anti-surge recirculation.

Closes equinor/ecalc-internal#2107

**Discussion:** equinor/ecalc-internal#2102

### DIFF
--- a/src/ecalc_neqsim_wrapper/fluid_service.py
+++ b/src/ecalc_neqsim_wrapper/fluid_service.py
@@ -511,6 +511,39 @@ class NeqSimFluidService(FluidService):
         standard_density = self._get_standard_density(fluid_model)
         return mass_rate_kg_per_h * 24.0 / standard_density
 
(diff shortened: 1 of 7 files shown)


jsolaas (approved):

jsolaas (approved):
Link to issue/discussion?

PR #1716, opened 2026-08-13
fix: clarify difference between instance and definition references

jsolaas:
This changes instance references to be required, while definition
references are optional. It also changes the instance reference from
process simulation to pipeline by removing the TARGET key.


tj098895:
Does this lead to change in YAML? Before/after example?

tj098895:
So, this will make it possible to use "any reference" whenever a reference is needed, so that when we parse and validate - we will always need to check that the type is correct and supported? E.g. to not reference to a process unit when a pipeline is expected etc? Or was this already the case, but never checked?

tj098895 (approved):
nice, I love when stuff I had thought of fixing just gets fixed :)

jsolaas:
> So, this will make it possible to use "any reference" whenever a reference is needed, so that when we parse and validate - we will always need to check that the type is correct and supported? E.g. to not reference to a process unit when a pipeline is expected etc? Or was this already the case, but never checked?

Always the case, not sure about the check though. We should be able to check that.

PR #1724, opened 2026-08-19
docs: add changelogs for v13.4 through v13.11

olelod:
Adds changelogs for versions v13.4–v13.11, which were released without documentation updates.


frodehk (approved):

tj098895 (approved):
very good

PR #1728, opened 2026-08-21
feat: energy domain ABCs

olelod:
ABCs for the new energy domain:

- **demand.py** — Demand base class, MechanicalPower, ElectricalPower, FuelGasRate, DieselRate
- **consumer.py** — Consumer[TDemand] ABC
- **provider.py** — Provider[TProvides], Source[TProvides], Converter[TRequires, TProvides] ABCs

Components say what they need to deliver what is asked, and what their capacity is.

### DIFF
--- a/src/libecalc/energy/demand.py
+++ b/src/libecalc/energy/demand.py
@@ -1,0 +1,68 @@
+from __future__ import annotations
+
+import abc
+from dataclasses import dataclass
+from typing import Self
+
+from libecalc.common.units import Unit
+
+
+@dataclass(frozen=True)
+class Demand(abc.ABC):
+    """Base for all energy demand types.
+
+    Additional carriers (e.g. ThermalPower/Steam) can be added as subclasses
+    when thermal energy modeling is needed.
+    """
+
+    value: float
+
+    def __post_init__(self) -> None:
+        if self.value < 0:
+            from libecalc.energy.errors import NegativeDemandError
+
+            raise NegativeDemandError(self.value, type(self).__name__)
+
+    @property
+    @abc.abstractmethod
+    def unit(self) -> Unit: ...
+
+    def __add__(self, other: Self) -> Self:
+        if type(self) is not type(other):
+            msg = f"Cannot add {type(self).__name__} and {type(other).__name__}"
+            raise TypeError(msg)
+        return type(self)(value=self.value + other.value)
+
+    def __sub__(self, other: Self) -> Self:
+        if type(self) is not type(other):
+            msg = f"Cannot subtract {type(other).__name__} from {type(self).__name__}"
+            raise TypeError(msg)
+        return type(self)(value=self.value - other.value)
+
+
+@dataclass(frozen=True)
+class MechanicalPower(Demand):
+    @property
+    def unit(self) -> Unit:
+        return Unit.MEGA_WATT
+
+
+@dataclass(frozen=True)
+class ElectricalPower(Demand):
+    @property
+    def unit(self) -> Unit:
+        return Unit.MEGA_WATT
+
+
+@dataclass(frozen=True)
+class FuelGasRate(Demand):
+    @property
+    def unit(self) -> Unit:
+        return Unit.STANDARD_CUBIC_METER_PER_DAY
+
+
+@dataclass(frozen=True)
+class DieselRate(Demand):
+    @property
+    def unit(self) -> Unit:
+        return Unit.LITRES_PER_DAY
--- a/src/libecalc/energy/provider.py
+++ b/src/libecalc/energy/provider.py
@@ -1,0 +1,45 @@
+from __future__ import annotations
+
+import abc
+from uuid import UUID
+
+from libecalc.energy.demand import Demand
+
+
+class Provider[TProvides: Demand](abc.ABC):
+    """Anything that supplies energy — from outside the system,
+    by converting one energy type to another, or by distributing across providers."""
+
+    @abc.abstractmethod
+    def get_id(self) -> UUID: ...
+
+    @abc.abstractmethod
+    def capacity(self) -> TProvides | None:
+        """Maximum this provider can deliver. None = unlimited."""
+        ...
+
+
+class Source[TProvides: Demand](Provider[TProvides]):
+    """Energy enters the system from an external source.
+
+    Examples: power from shore (ElectricalPower), fuel gas supply (FuelGasRate).
+    """
+
+    ...
+
+
+class Converter[TRequires: Demand, TProvides: Demand](Provider[TProvides]):
+    """Converts one energy type to another.
+
+    TRequires is what this converter needs as input, TProvides is what it delivers.
+
+    Examples:
+        Converter[ElectricalPower, MechanicalPower] — electric motor
+        Converter[FuelGasRate, ElectricalPower] — generator set
+        Converter[FuelGasRate, MechanicalPower] — gas turbine
+    """
+
+    @abc.abstractmethod
+    def get_input_demand(self, output_demand: TProvides) -> TRequires:
+        """Given output needed, what input is required?"""
+        ...
--- a/tests/libecalc/energy/test_energy_domain.py
+++ b/tests/libecalc/energy/test_energy_domain.py
@@ -1,0 +1,362 @@
+"""Test energy domain ABCs using a simplified version of the ltp_export example.
+
+INST_A topology (energy flow):
+
+    fuel_gas_sac (Source[FuelGasRate]) — unlimited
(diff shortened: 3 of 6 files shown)


src/libecalc/energy/demand.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+FUEL_UNIT: dict[FuelType, Unit] = {
+    FuelType.FUEL_GAS: Unit.STANDARD_CUBIC_METER_PER_DAY,
+    FuelType.DIESEL: Unit.LITRES_PER_DAY,
+}
+
+
+@dataclass(frozen=True)
+class Power:
```
frodehk: Could separate mechanical- and electrical power types be an option here? I.e. MechanicalPower and ElectricalPower (and not only PowerType in a general Power).

And:
`Demand = ElectricalPower | MechanicalPower | FuelRate`

src/libecalc/energy/provider.py
```diff
@@ -0,0 +1,53 @@
...
+class Source(Provider):
+    """Energy enters the system from an external source.
+
+    Examples: power from shore, offshore wind, fuel gas supply.
+    """
+
+    ...
+
+
+class Converter(Provider):
```
frodehk: Do we need a Converter class, if we say that normal behavior of the Provider.supply() is to convert (Generator, Turbine, Motor)? And that Source is the special case with no conversion?

src/libecalc/energy/provider.py
```diff
@@ -0,0 +1,53 @@
...
+
+class Provider(abc.ABC):
+    """Anything that supplies energy — from outside the system,
+    by converting one energy type to another, or by distributing across providers."""
+
+    @abc.abstractmethod
+    def get_id(self) -> str: ...
+
+    @abc.abstractmethod
+    def supply(self, demand: Demand) -> Supply: ...
```
frodehk: `demand_to_supply`: Demand -> Supply?

Edit: No, `demand` is better.

src/libecalc/energy/supply.py
```diff
@@ -0,0 +1,12 @@
...
+
+from dataclasses import dataclass
+
+from libecalc.energy.demand import Demand
+
+
+@dataclass(frozen=True)
+class Supply:
+    delivered: Demand
+    upstream_demand: Demand | None = None
```
frodehk: `input_demand` or `input_requirement`? Not sure if it is better. The challenge with `upstream_demand` is that it can be misunderstood as a demand from the upstream side, while in reality it is the demand/requirement from the provider (e.g. ElectricalMotor) to the upstream (e.g. the Generator) - in order to deliver the requested supply.

src/libecalc/energy/supply.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from libecalc.energy.demand import Demand
+
+
+@dataclass(frozen=True)
+class Supply:
```
frodehk: `requested` as well, or is it redundant?

src/libecalc/energy/demand.py
olelod: Done. Split into  MechanicalPower ,  ElectricalPower ,  FuelGasRate ,  DieselRate .

src/libecalc/energy/provider.py
olelod: Added generics, Converter[FuelGasRate, MechanicalPower]  makes the energy flow visible in the type signature — you can see what goes in and what comes out without reading any implementation.

src/libecalc/energy/supply.py
olelod: input_demand

src/libecalc/energy/supply.py
olelod: added requested, and some derived properties

src/libecalc/energy/supply.py
```diff
@@ -0,0 +1,23 @@
...
+    requested: Demand
```
frodehk: `Converter` and `Provider` are generic, should `Supply` follow same pattern?

src/libecalc/energy/supply.py
```diff
@@ -0,0 +1,23 @@
...
+from dataclasses import dataclass
+
+from libecalc.energy.demand import Demand
+
+
+@dataclass(frozen=True)
+class Supply[TDemand: Demand]:
+    requested: TDemand
+    delivered: TDemand
+    input_demand: Demand | None = None
```
frodehk: `Converter` already has the typed `input_demand(TProvides)`. Do we need it here as well? If we should have typed here, I guess it would be `TRequires | None `- is it sufficient to have `Converter.input_demand()`?

olelod: Removed  Supply  — providers just say what they need to deliver what is asked, and what their capacity is. The rest will be the solver's job.

src/libecalc/energy/provider.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import annotations
+
+import abc
+
+from libecalc.energy.demand import Demand
+
+
+class Provider[TDemand: Demand](abc.ABC):
```
frodehk: Suggestion: `Provider[TProvides: Demand](abc.ABC)`

TDemand says that the type is Demand - TProvides is more informative imo.

src/libecalc/energy/provider.py
```diff
@@ -0,0 +1,44 @@
...
+    @abc.abstractmethod
+    def get_id(self) -> str: ...
+
+    @abc.abstractmethod
+    def capacity(self) -> TDemand | None:
+        """Maximum this provider can deliver. None = unlimited."""
+        ...
+
+
+class Source[TDemand: Demand](Provider[TDemand]):
```
frodehk: Suggestion: `Source[TProvides: Demand](Provider[TProvides])`

frodehk:
Could we add one small test implementation? E.g. an ElectricMotor as `Converter[ElectricalPower, MechanicalPower]`? Mainly to test the generic types and illustrate use of `capacity() `and `input_demand()`.

tj098895:
> Could we add one small test implementation? E.g. an ElectricMotor as `Converter[ElectricalPower, MechanicalPower]`? Mainly to test the generic types and illustrate use of `capacity() `and `input_demand()`.

Why not just take the "ltp good" example that our current test yaml is based on, but where the energy stuff was removed. A real life example. Start simple. A real life example shows how easy it is to set up etc as well.

Do we have an overview of all "use cases" somewhere?

src/libecalc/energy/energy_units/bus.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
+from libecalc.energy import Consumer, ElectricalPower, Provider, Source
+
+
+@dataclass
+class ElectricalBus(Provider[ElectricalPower]):
```
frodehk: Just a thought: An alternative to include domain implementations in energy_units now, can be to have one or two local test impementations in the tests. And then take the discussions on different impelementations of energy units in the next step. Not sure what is best.

olelod: Did that first, then changed my mind... moved back now.

tests/libecalc/energy/test_energy_domain.py
```diff
@@ -0,0 +1,368 @@
...
+            BrokenConverter()  # type: ignore[abstract]
+
+    def test_converters_are_substitutable(self):
+        """Any Converter[FuelGasRate, T] can be used polymorphically.
+
+        This is the value proposition: code operating on the abstract interface
+        works across all concrete implementations without knowing the type.
+        """
+
+        def fuel_required(converter: Converter[FuelGasRate, Demand], demand: Demand) -> FuelGasRate:
```
frodehk: Suggestion:
```
fuel_required[TProvides: Demand](
   converter: Converter[FuelGasRate, TProvides], demand: TProvides)
) -> FuelGasRate
```

frodehk (approved):

src/libecalc/energy/provider.py
```diff
@@ -0,0 +1,44 @@
...
+    TRequires is what this converter needs as input, TProvides is what it delivers.
+
+    Examples:
+        Converter[ElectricalPower, MechanicalPower] — electric motor
+        Converter[FuelGasRate, ElectricalPower] — generator set
+        Converter[FuelGasRate, MechanicalPower] — gas turbine
+    """
+
+    @abc.abstractmethod
+    def input_demand(self, demand: TProvides) -> TRequires:
```
jsolaas:
```suggestion
    def get_input_demand(self, output_demand: TProvides) -> TRequires:
```

jsolaas (approved):

PR #1733, opened 2026-08-25
Process definition/instance error messages

jsolaas:
- **chore: improve error messages for invalid definitions**
- **chore: improve invalid instance reference error messages**


src/libecalc/presentation/yaml/yaml_reference_service.py
olelod: Not from this PR, but saw it now. This one can produce `invalid reference reference`? A little awkward, maybe?

olelod (approved):

PR #1734, opened 2026-08-25
feat(libecalc): add energy unit implementations

olelod:
Concrete implementations of the energy domain ABCs from #1728.

### DIFF
--- a/src/libecalc/energy/energy_units/consumers.py
+++ b/src/libecalc/energy/energy_units/consumers.py
@@ -1,0 +1,156 @@
+from typing import Final
+
+from libecalc.energy.consumer import Consumer
+from libecalc.energy.demand import DieselRate, ElectricalPower, FuelGasRate, MechanicalPower
+from libecalc.energy.energy_unit import EnergyUnitId
+
+
+class BaseLoad(Consumer[ElectricalPower]):
+    """Fixed electrical load (e.g. heating, steam generator, lighting).
+
+    Some base loads (e.g. steam generator) may become converters if thermal
+    energy modeling is introduced.
+    """
+
+    def __init__(self, name: str, load: float, energy_unit_id: EnergyUnitId | None = None) -> None:
+        self._name = name
+        self._load = load
+        self._id: Final[EnergyUnitId] = energy_unit_id or BaseLoad._create_id()
+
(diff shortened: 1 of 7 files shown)


src/libecalc/energy/energy_units/node.py
```diff
@@ -0,0 +1,32 @@
+from dataclasses import dataclass
+
+from libecalc.energy.demand import Demand, ElectricalPower, FuelGasRate
+
+
+@dataclass
+class EnergyNode[T: Demand]:
```
frodehk: `Junction` instead of `EnergyNode`? EnergyNode sounds like the common type for every components in the network (consumers, providers, sources)

src/libecalc/energy/energy_units/consumers.py
```diff
@@ -0,0 +1,106 @@
...
+    """Pump requiring mechanical power (e.g. water injection pump)."""
+
+    name: str
+    power_mw: float
+
+    def get_id(self) -> str:
+        return self.name
+
+    def get_demand(self) -> MechanicalPower:
+        return MechanicalPower(self.power_mw)
```
jsolaas:
```suggestion

class Pump(Consumer[MechanicalPower]):
    """Pump requiring mechanical power (e.g. water injection pump)."""
    def __init__(name: str, consumer_id: UUID | None, power):
        self._id = consumer_id or generate_id()
        self._power = power

    def get_id(self) -> UUID:
        return self._id
    def get_name(self) -> str:
        return self.name
    def get_demand(self) -> MechanicalPower:
        return MechanicalPower(self.power_mw)
```

jsolaas (approved):

src/libecalc/energy/energy_units/consumers.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def __init__(self, name: str, load: float, energy_unit_id: EnergyUnitId | None = None) -> None:
+        self._name = name
+        self._load = load
+        self._id: Final[EnergyUnitId] = energy_unit_id or BaseLoad._create_id()
+
+    def get_id(self) -> EnergyUnitId:
+        return self._id
+
+    def get_name(self) -> str:
```
olelod, line 23: Should this one also be abstract? EnergyUnits always have name?

PR #1739, opened 2026-08-26
feat(libecalc): add EnergyUnitId with Entity pattern

olelod:
Introduce `EnergyUnit` base class extending `Entity[EnergyUnitId]`, following the `ProcessUnit`/`ProcessUnitId` pattern.

### DIFF
--- a/tests/libecalc/energy/test_energy_domain.py
+++ b/tests/libecalc/energy/test_energy_domain.py
@@ -30,8 +30,7 @@ INST_A topology (energy flow):
 
 from __future__ import annotations
 
-from dataclasses import dataclass, field
-from uuid import UUID, uuid4
+from typing import Final
 
 import pytest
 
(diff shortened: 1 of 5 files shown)


tests/libecalc/energy/test_energy_domain.py
```diff
@@ -144,100 +158,121 @@ def get_input_demand(self, output_demand: MechanicalPower) -> ElectricalPower:
...
 
     def get_demand(self) -> MechanicalPower:
         return MechanicalPower(self.power_mw)
 
 
-@dataclass
 class Pump(Consumer[MechanicalPower]):
-    name: str
-    power_mw: float
+    def __init__(self, name: str, power_mw: float, *, id: EnergyUnitId | None = None) -> None:
```
tj098895: Shall we use * in args? Should we be consistent?

tj098895 (approved):

tests/libecalc/energy/test_energy_domain.py
olelod: removed

PR #1743, opened 2026-08-27
refactor: remove generics from energy domain, rename Demand to Energy

olelod:


src/libecalc/energy/consumer.py
```diff
@@ -2,16 +2,21 @@
...
     """Demands energy. End of the energy chain.
 
     Examples: compressor (mechanical), pump (mechanical),
               base load (electrical), flare (fuel).
     """
 
+    input_energy_type: type[Energy]
+
+    def get_input_energy_type(self) -> type[Energy]:
+        return self.input_energy_type
```
jsolaas:
```suggestion
    @classmethod
    @abc.abstractmethod
    def get_input_energy_type(cls) -> type[Energy]: ...
```

I would suggest this instead, then just implement that directly in the subclasses

jsolaas (approved):

src/libecalc/energy/energy_units/junction.py
```diff
@@ -1,17 +1,28 @@
...
     Connections to providers and consumers are managed by the Network,
     not by internal lists. The solver resolves capacity and demand
     from the network graph.
     """
 
+    @classmethod
+    @abc.abstractmethod
+    def get_energy_type(cls) -> type[Energy]: ...
+
+    def get_input_energy_type(self) -> type[Energy]:
```
frodehk: classmethod, to be consistent with similar methods in consumers/converters/sources?

frodehk (approved):

PR #1753, opened 2026-09-01
feat(libecalc): add energy network YAML types with example and tests

olelod:


src/libecalc/examples/energy/energy_network.yaml
```diff
@@ -0,0 +1,123 @@
...
+      TYPE: ELECTRICAL_CABLE
+      INPUT: power_from_shore
+      EFFICIENCY: 0.97
+
+    # --- Junctions ---
+    - NAME: electrical_bus
+      TYPE: ELECTRICAL_BUS
+      INPUT: [subsea_cable, wind_turbine]
+      DISPATCH_STRATEGY: PRIORITY
+
```
olelod, line 71: When you have two inputs (fan in) the bus/manifold needs to be defined in the yaml. And also a strategy from where to "draw" energy first.

src/libecalc/examples/energy/energy_network.yaml
```diff
@@ -0,0 +1,123 @@
...
+    - NAME: fuel_manifold
+      TYPE: FUEL_GAS_MANIFOLD
+      INPUT: [fuel_gas]
+
+    # --- Consumers ---
+    - NAME: base_load
+      TYPE: ELECTRICAL
+      INPUT: genset_a
+      LOAD: 5  # MW
+
```
olelod, line 81: Several components "drawing" energy from the same component (fan out) does not specifically require a bus/manifold to be defined, but it will be added to the network in the mapping.

src/libecalc/examples/energy/energy_network.yaml
```diff
@@ -0,0 +1,121 @@
...
+#   flare_gas ──► flare (1200 Sm3/d)
+#
+#   diesel ──► diesel_consumers (500 l/d)
+#
+#   power_from_shore ──► subsea_cable ──┐
+#                                       ├──► electrical_bus ──► heating (10 MW)
+#   wind_turbine ───────────────────────┘
+
+ENERGY_NETWORK:
+  SOURCES:
```
olelod, line 23: SOURCES are the root nodes of the network. The emission calculations will later look for aggregated numbers for the sources that produce emission.

src/libecalc/examples/energy/energy_network.yaml
```diff
@@ -0,0 +1,121 @@
...
+    - NAME: genset_a
+      TYPE: GENERATOR_SET
+      INPUT: fuel_gas
+      MODEL: genset_17mw
+      CAPACITY: 17  # MW
+
+    - NAME: export_turbine
+      TYPE: GAS_TURBINE
+      INPUT: fuel_gas
+      MODEL: turbine_30mw
```
olelod, line 52: May want to move it somewhere else than what we have in legacy today - but it should point to a model linking turbine load and energy usage.

src/libecalc/examples/energy/energy_network.yaml
```diff
@@ -0,0 +1,121 @@
...
+    - NAME: wind_turbine
+      TYPE: ELECTRICAL
+      CAPACITY: 4.4  # MW
+
+  UNITS:
+    # --- Converters ---
+    - NAME: genset_a
+      TYPE: GENERATOR_SET
+      INPUT: fuel_gas
+      MODEL: genset_17mw
```
olelod, line 46: May want to move it somewhere else than what we have in legacy today - but it should point to a model linking produced electric power and energy usage.

src/libecalc/examples/energy/energy_network.yaml
```diff
@@ -0,0 +1,121 @@
...
+    - NAME: diesel_consumers
+      TYPE: DIESEL
+      INPUT: diesel
+      RATE: 500  # l/d
+
+    # Process simulation driven
+    - NAME: export_train
+      TYPE: MECHANICAL
+      INPUT: export_turbine
+      PROCESS_SIMULATION: export_compressor_sim
```
olelod, line 116: This here should point to the PROCESS_SIMULATION part of the YAML somehow.

src/libecalc/examples/energy/energy_network.yaml
```diff
@@ -0,0 +1,121 @@
...
+#
+#   flare_gas ──► flare (1200 Sm3/d)
+#
+#   diesel ──► diesel_consumers (500 l/d)
+#
+#   power_from_shore ──► subsea_cable ──┐
+#                                       ├──► electrical_bus ──► heating (10 MW)
+#   wind_turbine ───────────────────────┘
+
+ENERGY_NETWORK:
```
olelod, line 22: Topic for discussion: Everything in PROCESS_SIMULATION uses stream day rates, so regularity is not a thing there. How should it enter here? Energy network wide? Potentially one for every source/component. Inherit from the component where the input is received from? The regularity of a compressor can not exceed the regularity of the motor driving it, or the genset driving the motor.

jsolaas (approved):

src/libecalc/presentation/yaml/yaml_types/energy/yaml_energy_network.py
```diff
@@ -0,0 +1,442 @@
...
+    def check_input_references(self):
+        valid = self._valid_input_names()
+        for c in self.units:
+            for ref in _get_input_names(c):
+                if ref not in valid:
+                    raise ValueError(f"'{c.name}' has INPUT '{ref}' which is not a known source or provider.")
+        return self
+
+    @model_validator(mode="after")
+    def check_no_cycles(self):
```
frodehk, line 412: EnergyNetwork.create_topological_order() already does this - should Yaml-validation have this responsibility, or is it the network´s responsibility?

src/libecalc/presentation/yaml/yaml_types/energy/yaml_energy_network.py
```diff
@@ -0,0 +1,442 @@
...
+
+        graph: dict[str, set[str]] = {c.name: set(_get_input_names(c)) for c in self.units}
+        try:
+            tuple(TopologicalSorter(graph).static_order())
+        except CycleError as e:
+            raise ValueError(f"Cycle detected in energy network: {e}") from None
+        return self
+
+    @model_validator(mode="after")
+    def check_energy_type_compatibility(self):
```
frodehk, line 423: EnergyNetwork.validate_connection() already does this: same question, Yaml- or network responsibility?

src/libecalc/presentation/yaml/yaml_types/energy/yaml_energy_network.py
olelod: Was thinking it should be addressed in the mapping - the sooner it is discovered the better, so probably let it be the responsibility of the yaml validation.

PR #1768, opened 2026-09-07
fix(libecalc): give each energy role check a single definition

kjbrak:
EnergyNetwork checked node roles with hand-written isinstance tuples across six call sites, and the lists had drifted apart. Transporters were missing from two of them.

A network containing a cable could not be computed, because get_input_energy raised "Unsupported energy unit type" for a node that get_output_energy handled fine. A cable with nothing feeding it was accepted, and then reported the energy it needed as if that energy existed.

Four named type unions now defines the required capabilities in one place: ProvidesEnergy, RequiresEnergy, DerivesInputFromOutput and HasCapacity. Junctions answer get_input_energy themselves by returning what they deliver, keeping their lossless behaviour in the junction rather than the network. 

The allocation error reports the successor's own predecessor count instead of the total node count.

### DIFF
--- a/src/libecalc/energy/network.py
+++ b/src/libecalc/energy/network.py
@@ -15,6 +15,13 @@ from libecalc.energy.source import Source
 
 type EnergyNetworkNode = Consumer | Source | Converter | Transporter | Junction
 
+# Role groups the network asks about. Each question has one definition, so a new role
+# cannot be handled in one place and forgotten in another.
+ProvidesEnergy = Source | Converter | Transporter | Junction
+RequiresEnergy = Consumer | Converter | Transporter | Junction
+DerivesInputFromOutput = Junction | Converter | Transporter
+HasCapacity = Source | Converter | Transporter
+
 
 @value_object
 class EnergyConnection:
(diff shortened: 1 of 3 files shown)


kjbrak:
Since earlier the base classes that perhaps would have been used here were reverted; Do you prefer just being specific on the isinstance checks here and type out the tuple each place?

src/libecalc/energy/network.py
```diff
@@ -15,6 +15,13 @@
 
 type EnergyNetworkNode = Consumer | Source | Converter | Transporter | Junction
 
+# Role groups the network asks about. Each question has one definition, so a new role
+# cannot be handled in one place and forgotten in another.
+ProvidesEnergy = Source | Converter | Transporter | Junction
```
tj098895, line 20: nicely identified roles/categories and suggestions for simplifications, but can we make it easier by only having to update one place to make sure that we suddenly dont break or forget about this for implementations by e.g. using interfaces with certain methods (which would then act as a protocol/ducktype), or enforce (all) nodes to answer the questions: "do you consume energy" and "do you provide energy" and add fitting boolean?

tj098895 (approved):

jsolaas (approved):

frodehk (approved):
Nice!

src/libecalc/energy/network.py
kjbrak: Yes, considered runtime_checkable protocol, but the way it is now. then Consumer has get_input_energy() (no argument and the others have argument. That is appearantly ok for pyright but not for runtime check isinstance since signatures dont count there, only method name.

So renaming those with singatures to eg. `derive_input_energy(output)` would work.

yes, a `bool` method is a nice idea, but it would not narrow the type and after the if bool check one would still need an isinstance or cast for type narrowing? And also, might be fragile since the bool is providing the capability it only asserts one and will thus be decoupled from the actual capability right?

kjbrak: And also, feel like type unions are more explicit and lower cognitive load than protocols.

PR #1771, opened 2026-09-08
refactor: move energy type contract to EnergyUnit

frodehk:
Moves the input and output energy type contract to `EnergyUnit`.
 
Next step in upcoming PR: Move calculations, capacities, demands, and conversion models into `EnergyNetworkEvaluation`. This will make the remaining role definitions unnecessary and leave `EnergyNetwork` responsible only for topology.

Refs: equinor/ecalc-internal#2203

---

## Type of Work

### DIFF
--- a/src/libecalc/energy/consumer.py
+++ b/src/libecalc/energy/consumer.py
@@ -13,6 +13,10 @@ class Consumer(EnergyUnit, abc.ABC):
     @abc.abstractmethod
     def get_input_energy_type(cls) -> type[Energy]: ...
 
+    @classmethod
+    def get_output_energy_type(cls) -> None:
+        return None
+
     @abc.abstractmethod
     def get_input_energy(self) -> Energy:
         """Return the input energy demanded by this consumer."""
--- a/src/libecalc/energy/network.py
+++ b/src/libecalc/energy/network.py
@@ -16,7 +16,6 @@ from libecalc.energy.source import Source
 # Role groups the network asks about. Each question has one definition, so a new role
 # cannot be handled in one place and forgotten in another.
 ProvidesEnergy = Source | Converter | Transporter | Junction
-RequiresEnergy = Consumer | Converter | Transporter | Junction
 DerivesInputFromOutput = Junction | Converter | Transporter
 HasCapacity = Source | Converter | Transporter
 
--- a/src/libecalc/energy/network.py
+++ b/src/libecalc/energy/network.py
@@ -22,7 +22,7 @@ HasCapacity = Source | Converter | Transporter
 
 @value_object
 class EnergyConnection:
-    """A directed connection between two energy units."""
+    """A directed connection between two energy nodes."""
 
     source_id: EnergyUnitId
     target_id: EnergyUnitId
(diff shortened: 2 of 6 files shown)


src/libecalc/energy/consumer.py
```diff
@@ -13,6 +13,10 @@ class Consumer(EnergyUnit, abc.ABC):
     @abc.abstractmethod
     def get_input_energy_type(cls) -> type[Energy]: ...
 
+    @classmethod
+    def get_output_energy_type(cls) -> None:
```
tj098895, line 17: I always prefer, when possible, to also make a type out of `None`, to avoid `NULLs` and problems related to that, to have some more type safety and avoid `NullPointerException`s (or similar None does not have and to avoid the extra None check) ie more fluent code etc. Can wait though, but its a preference of mine for clarity, fewer bugs etc :D

src/libecalc/energy/network.py
```diff
@@ -39,13 +37,13 @@ class EnergyNetwork:
 
     def __init__(
         self,
-        nodes: Iterable[EnergyNetworkNode],
+        nodes: Iterable[EnergyUnit],
```
tj098895, line 39: rename to `units` as well for consistency, or do we want to differentiate between a node and a unit? ...

tj098895: I do think that there may be a difference, but not yet. A node is "just a physical network component", while an "energy unit" has business logic etc, but it depends on how we implement it. After all it is probably the same thing, just 2 different aspects of it ...

tj098895 (approved):
nice :)

src/libecalc/energy/network.py
frodehk: I was renaming the other way - from units to nodes - but of course ended up with the inconsistency with EnergyUnit in the end:-) Agree, either node or unit. (I felt node was more consistent with a network, but no problem with units)

tj098895: didnt check code, but imo a unit has different capabilities than a node, most likely. possibly an implementation of a node. a node only has "network/graph related capabilities" possiby; ie 2 different interfaces. I will have to think a bit about this :)

src/libecalc/energy/consumer.py
frodehk: I agree in avoiding NULLs when possible. Here `None` is not another energy type, but absence of energy input. I guess a `NoEnergy` type still requires some checking/handling?

tj098895: it depends on how it is being used, but not necessarily. If Option had been a (more integrated) part of Python, it would be better to use None. I do however understand the choice, and it is I guess more Pythonic, but if I see we benefit from having a `NoEnergy` type, I might change it :) can stay as is though, just wanted to share 🗡️

src/libecalc/energy/network.py
frodehk: Was thinking more like energy units (domain) are inputs for the nodes (network), if that makes sense. But we may not need to make that distinction?

tj098895: agree - just need to figure out "the network is fed with e.g. that info" :D

frodehk: Suggestion: we keep nodes as it is in EnergyNetwork for now. Next PR will move quite a few things out from the EnergyNetwork. Maybe it is clearer for us then, what´s the best choice?

PR #1774, opened 2026-09-09
feat(libecalc): add sampled compressor to energy domain

olelod:
## Summary

Closes equinor/ecalc-internal#2205

Adds `SAMPLED_COMPRESSOR`: a compressor modelled directly from a tabulated FILE (RATE/SUCTION_PRESSURE/DISCHARGE_PRESSURE vs. FUEL and/or POWER), matching legacy's `COMPRESSOR_TABULAR` convention, with no separate `MODELS` entry or manual turbine/compressor wiring required.

Depending on FILE's columns and INPUT's energy type, the unit is expanded before mapping into one of four flows:
- **FUEL only** → fuel gas consumed directly
- **POWER only** → POWER is interpreted according to INPUT's own energy unit: electrical power consumed directly if INPUT is electrical (e.g. a generator set), or mechanical power consumed directly if INPUT is a separately-declared `GAS_TURBINE`/`ELECTRICAL_MOTOR` unit
- **Both FUEL and POWER** → an implicit turbine (inverted from FILE's own samples) feeding the compressor

## Reimplementing convex_hull and sampled_compressor

Legacy's `CompressorModelSampled(1D/2D/3D)` and its convex-hull extrapolation (`domain.process.compressor.core.sampled`) are `libecalc.domain`-coupled and array-batch oriented. The new energy domain reimplements both dependency-free and scalar (one operating point in, one value out): `sampled_compressor.py` collapses the 1D/2D/3D class family into a single `SampledCompressor` entry point that picks the right interpolator based on which of rate/ps/pd actually vary in the sample data, and `convex_hull.py` trims the shared boundary-extrapolation geometry from ~495 to ~220 lines by dropping unused batch/array machinery.


olelod:
Superseded by stacked PRs for easier review: #1788 → #1789 → #1790 → #1791 → #1792.

PR #1775, opened 2026-09-09
feat: add energy network evaluation

frodehk:
Adds `EnergyNetworkEvaluation`, which accepts consumer demands by node ID and propagates them through the energy network.

Consumer demand is supplied separately from the network structure.

**Next**
Remove demand values from consumers and the duplicated calculation methods from `EnergyNetwork`. Capacity and feasibility will be handled separately.

Refs: equinor/ecalc-internal#2203

---

## Type of Work

### DIFF
--- a/src/libecalc/energy/energy_network_evaluation.py
+++ b/src/libecalc/energy/energy_network_evaluation.py
@@ -1,0 +1,141 @@
+from collections.abc import Mapping
+
+from libecalc.energy import Consumer, Converter, Energy, EnergyUnit, EnergyUnitId
+from libecalc.energy.energy_units import Junction, Transporter
+from libecalc.energy.errors import (
+    EnergyAllocationRequiredError,
+    InvalidEnergyNetworkError,
+    InvalidEnergyNetworkEvaluationInputError,
+)
+from libecalc.energy.network import EnergyNetwork
+
+
+class EnergyNetworkEvaluation:
+    def __init__(
+        self,
+        energy_network: EnergyNetwork,
+        consumer_demands: Mapping[EnergyUnitId, Energy],
+    ) -> None:
+        self._energy_network = energy_network
+        self._consumer_demands = dict(consumer_demands)
+        self._validate_inputs()
+
+    # Per-node energy
+    def get_input_energy(
+        self,
+        node_id: EnergyUnitId,
+    ) -> Energy | None:
+        node = self._energy_network.get_node(node_id)
+
+        if isinstance(node, Consumer):
+            return self._consumer_demands[node_id]
+
+        if node.get_input_energy_type() is None:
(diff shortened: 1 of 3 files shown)


src/libecalc/energy/energy_network_evaluation.py
```diff
@@ -0,0 +1,141 @@
...
+    InvalidEnergyNetworkEvaluationInputError,
+)
+from libecalc.energy.network import EnergyNetwork
+
+
+class EnergyNetworkEvaluation:
+    def __init__(
+        self,
+        energy_network: EnergyNetwork,
+        consumer_demands: Mapping[EnergyUnitId, Energy],
```
tj098895: in general, I like that you use `Mapping`, but did we use that in the past? Do we currently use other Mapping implementations than dict? I would want to postpone instead of having different ways to do it. I do like the approach tho, but then we should have a general overhaul to change to this, if that is possible

src/libecalc/energy/energy_network_evaluation.py
```diff
@@ -0,0 +1,141 @@
...
+from libecalc.energy.energy_units import Junction, Transporter
+from libecalc.energy.errors import (
+    EnergyAllocationRequiredError,
+    InvalidEnergyNetworkError,
+    InvalidEnergyNetworkEvaluationInputError,
+)
+from libecalc.energy.network import EnergyNetwork
+
+
+class EnergyNetworkEvaluation:
```
tj098895, line 11: I would probably want to simulate the whole network first (or is it evaluate?) - to calculate the "numbers" for every node in the network. Next I would query by traversing the network (again), and querying directly the number for each node, or the node I want the number for, if needed. Then I would separate between calculation and querying. Also, normally we would want to calculate the whole network initially, and sometimes query individuals. Could also be possible to calculate a subgraph, by e.g. adding a node id as parameter, and tell it to calculate only the subgraph/network.

dont need to do anything about it now though, just 5 cents :)

still not sure if we would want to get one by one, or the whole network queried at once tho. lets see what makes sense .)

tj098895 (approved):

src/libecalc/energy/energy_network_evaluation.py
frodehk: That is true - we mainly have used dictionaries. Used Mapping here to accept a dict-like interface, read-only. However, I agree - we dont need to add it here now.

src/libecalc/energy/energy_network_evaluation.py
frodehk: Good point. And I have thought about it myself - calculating the whole network and store results. However, as a first step I introduced on-demand calculation per node. But we can discuss and change this later.

PR #1777, opened 2026-09-09
2201 yaml version

tj098895:
Very initial and simple way of introducing versions in YAML without being intrusive.

- YAML v1 (legacy) should not need to worry about it. If not specified, it fallbacks to v1.
- YAML > v2 (eCalc NG) MUST worry about it - to use new functionality

I have plans to extend this in future, and to make the doc informative using when it is being introduced as a part of annotations or extra data in yaml objects, when it is deprecated or removed etc for additional information for users in both doc and runtime (validation).

Being more explicit of when we add new stuff and when it is breaking wrt YAML is very informative for developers and users, and we can use this for automation as well. It makes ourselves also more attentive to consequences of making these changes, and what we need to take into consideration.

But, for now;
- required version for new eCalc YAML
- new YAML keywords are only allowed for new eCalc YAML

## Type of Work

### DIFF
--- a/src/libecalc/presentation/yaml/yaml_types/components/yaml_asset.py
+++ b/src/libecalc/presentation/yaml/yaml_types/components/yaml_asset.py
@@ -2,6 +2,7 @@ from pydantic import ConfigDict, Field, field_validator, model_validator
 from pydantic_core.core_schema import ValidationInfo
 
 from libecalc.common.string.string_utils import get_duplicates
+from libecalc.common.version import Version
 from libecalc.presentation.yaml.yaml_types import YamlBase
 from libecalc.presentation.yaml.yaml_types.components.yaml_installation import YamlInstallation
 from libecalc.presentation.yaml.yaml_types.energy.yaml_energy_network import YamlEnergyNetwork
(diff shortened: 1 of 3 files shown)


src/libecalc/presentation/yaml/yaml_types/components/yaml_asset.py
```diff
@@ -234,3 +242,59 @@ def validate_unique_references(self):
                 f" Duplicated references are: {', '.join(duplicated_references)}"
             )
         return self
+
+    @model_validator(mode="before")
+    @classmethod
+    def validate_v2_fields_require_v2(cls, data):
```
jsolaas, line 248: We should consider dynamically creating the class instead, that would give a better json-schema

tj098895: not sure what that means. lets discuss.

(this was currently considered as a very simple initial start, but it can easily grow big and very helpful if we use it wisely imo, such as adding version info on each (root) element", and dynamically provide schema, doc etc, but I wanted to keep this simple as a start - just to get the idea of yaml versioning in place etc

jsolaas: Basically creating separate classes for each version. Discriminated union would be an alternative.

tj098895:
closing. the initiative was to make a stripped down minimal introduction of version, but ended up being something more. Will instead of going for the stripped down version - or make a proper way of dealing with versions that will lead to many good things.

PR #1779, opened 2026-09-10
Refactor energy network, evaluation and connection

jsolaas:
- Focus on energy per connection
- Force separation between energy units and network by only keeping unit ids in network, connections keep energy type information


src/libecalc/energy/network.py
```diff
@@ -5,40 +5,39 @@
...
 
+EnergyNetworkId = NewType("EnergyNetworkId", UUID)
+EnergyConnectionId = NewType("EnergyConnectionId", UUID)
+
 
 @value_object
 class EnergyConnection:
     """A directed connection between two energy nodes."""
 
+    id: EnergyConnectionId
```
tj098895, line 20: if it has an ID and we need to track it, its an entity (ID is created in domain), otherwise the ID only exists at database/infrastructure layer. I think it is an entity, and it seems to make it easier rather than tracking it by from-to ID association? And since the temporal energy is associated with it, and not the unit. yes :)

src/libecalc/energy/energy_network_evaluation.py
tj098895, line 14: Kan/skal vi splitte dette i 2/3 moduler, slik at network ikke kjenner til energyunit, unit kjenner ikke til network, men simulering kjenner til begge?

- energy
   - energy_network (topology, strategies?)
   - energy_model (impl of models, units, etc)?
   - energy_simulation (ways to traverse/walk network, get data from energy models etc, apply into network)

 half.-thought

src/libecalc/energy/network.py
```diff
@@ -99,6 +148,8 @@ def _add_connections(
 
             self._successors[connection.source_id].add(connection.target_id)
             self._predecessors[connection.target_id].add(connection.source_id)
+            self._connections[connection.source_id, connection.target_id] = connection
+            self._connections_by_id[connection.id] = connection
 
     def _validate_connection(
```
frodehk, line 161: Should validate_connection also reject duplicate connection IDs? And duplicate (source_id, target_id) pairs?

frodehk (approved):
Nice! I like the approach — clearer separation between topology and energy units. Giving connections responsibility for the energy type makes sense.

PR #1785, opened 2026-09-15
feat: report energy capacity failures

frodehk:
Capacity was previously exposed through separate boolean methods called after energy propagation. Callers had to provide capacities again, making it possible to evaluate different inputs than those used for dispatch. The boolean result also provided no context about the required energy and configured capacity.

Energy propagation now returns capacity failures together with connection energy. This ensures that the same capacities are used for both dispatch and capacity evaluation. Each failure includes the required energy and configured capacity, allowing callers to understand the failure rather than only receiving `true` or `false`.

Missing capacity still means unlimited, while invalid capacity input remains a validation error.

Refs: equinor/ecalc-internal#2229

---

## Type of Work


src/libecalc/energy/capacity.py
```diff
@@ -0,0 +1,19 @@
+from enum import StrEnum
+
+from libecalc.energy.energy_types import Energy
+
+
+class CapacityFailureStatus(StrEnum):
```
tj098895: I think it may be more relevant to have `EnergyFailureStatus`, and one of them would be `Capacity_Exceeded` among many others, then we could easily match on that to pick relevant info from the failure status object - we want to have many different failure status objects. Sometimes exceptions, sometimes first-class citizen object like this.

src/libecalc/energy/energy_network.py
```diff
@@ -41,8 +42,11 @@ def propagate_energy(
         self,
         connection_demands: dict[EnergyConnectionId, Energy],
         capacities: dict[EnergyUnitId, Energy] | None = None,
-    ) -> dict[EnergyConnectionId, Energy]:
-        """Calculate connection energy from demands on consumer-targeting connections.
+    ) -> tuple[
+        dict[EnergyConnectionId, Energy],
+        dict[EnergyUnitId, CapacityFailure],
```
tj098895, line 47: connection on success, unit on failure? or do we want to be consistent and have e.g. connection for both, and it will either return success or failure for each pair of require/supply connection? If there is a failure, detailed info on whether it was the total supply or an individual contributor that did not fulfill the requirement is informed about? If there is a failure, currently it seems that we continue - here we may make a choice, to abort early or try to continue as much as possible - keeping in mind, how to deal with "følgefeil"? Should parents skips calculation, or should that continue? Imo. currently we should skip, and possibly add a custom "error" that not_calculated due to underlying error etc

src/libecalc/energy/capacity.py
frodehk: Good point!

src/libecalc/energy/energy_network.py
frodehk: In my view the energy belongs to connections (flows from a to b) - and should be keyed by connection id. Capacity is about to the sum of one units outgoing energy (what it can deliver) - one unit can have several outgoing connections. Hence, imo capacity error should be by unit id. Exceeding capacity is an infeasible result, but not a calculation error - hence I was thinking we should continue propagation to get total required energy and all capacity failures.  Imo NOT_CALCULATED etc. should be resrved for actual calculation errors. But there may be reasons that we want to abort early anyway?

src/libecalc/energy/capacity.py
frodehk: @tj098895 : Updated with `EnergyFailureStatus`, and a `EnergyFailure` base class.

src/libecalc/energy/energy_network.py
tj098895: Maybe ....?

But in that case, are further calculations up towaards the source labelled as failures or not; ie "følgefeil", is that considered a failure or not? We may see this differently, but right now I think that følgefeil should be labelled as well - easy to track origin of failures e.g. visually, and label it as "følgefeil" ?

We agree on where the "energy is tied to", connections, the supply/demand pair :)

But shouldnt the error be noted the on the connetion, that the demand is NOT able to be fulfilled? But, then the SOURCE of the error is considered to be "somewhere" on the supply side? Which imo is to be provided in the failure object :) Or? (It is a question of definition, but for consistency my gut feeling is that it would be more consistent like that?)

tj098895 (approved):

PR #1788, opened 2026-09-15
feat: add SampledCompressor energy model (1/5)

olelod:
1 of 5, splitting #1774 into stacked, independently reviewable PRs.

`SampledCompressor` energy model: interpolation over sampled rate/power/fuel data, with a reimplementation of the existing convex-hull extrapolation logic (ported from the legacy domain) for extrapolating outside the sampled envelope.

Stack: **1/5** → #1789 → #1790 → #1791 → #1792
Supersedes #1774.


src/libecalc/energy/models/sampled_compressor.py
```diff
@@ -0,0 +1,810 @@
...
+        )
+
+    def evaluate(
+        self,
+        rate: float | None = None,
+        suction_pressure: float | None = None,
+        discharge_pressure: float | None = None,
+    ) -> float:
+        variable = {
+            RATE_NAME: rate,
```
tj098895: can this be either required to use fixed names and map to an object earlier, or use different classes depending on data, instead of having to map here?

src/libecalc/energy/models/sampled_compressor.py
```diff
@@ -0,0 +1,810 @@
...
+        self.support_max_rate = variable == RATE_NAME
+        self._interpolator = interp1d(
+            x=sorted_points,
+            y=sorted_values,
+            fill_value=fill_value,
+            bounds_error=False,
+            assume_sorted=True,
+        )
+
+    def evaluate(
```
tj098895, line 281: overload to signal allowed combinations? If that makes sense?

tests/libecalc/energy/models/test_sampled_compressor.py
```diff
@@ -0,0 +1,714 @@
...
+                "rate": [1.0, 1.0, 1.0, 1.0],
+                "suction_pressure": [1.0, 2.0, 4.0, 5.0],
+                "discharge_pressure": [1.0, 1.0, 1.0, 1.0],
+            },
+            (PS_NAME,),
+        ),
+    ],
+)
+def test_dispatcher_matches_legacy(data, evaluation, active_variables):
+    common = SampledCompressor(**data)
```
tj098895: can we instead use a factory that returns a `SampledCompressor` interface that all of the implementations use, and have same signatures? If possible to determine implementation at object creation time?

```
sampled_compressor = SampledCompressorFactory.create(**data)
sampled_compressor.evaluate(*)
```
?

tj098895 (approved):
nice separation :)

src/libecalc/energy/models/convex_hull.py
```diff
@@ -0,0 +1,220 @@
...
+import numpy as np
+from numpy.typing import NDArray
+from scipy.interpolate import LinearNDInterpolator
+from scipy.spatial import ConvexHull
+
+FloatArray = NDArray[np.float64]
+IntArray = NDArray[np.int64]
+
+
+@dataclass(frozen=True)
```
tj098895, line 40: can we make sure that this is completely separate, and does not import any code form libecalc/ecalc (see architecture tests)?

src/libecalc/energy/models/convex_hull.py
```diff
@@ -0,0 +1,220 @@
...
+class HalfConvexHull:
```
tj098895, line 41: Is this same impl as before? Can this or should this be used other places? It is a "general convex hull implementation"? Pumps? 

Will this most likely be as it is and never change, or?

src/libecalc/energy/models/sampled_compressor.py
olelod: You mean making a `Rate/Ps/Pd-Compressor1D(SampledCompressorBase)` kind of thing?

olelod: In that case i guess the same should apply to 2D. Different classes.

src/libecalc/energy/models/convex_hull.py
olelod: libecalc.energy only imports stdlib/numpy/scipy plus  libecalc.common. I'll add to the architecture test suite.

src/libecalc/energy/models/convex_hull.py
olelod: Only sampled compressors use convex hull. Nothing else. Pumps may use the simpler tabular models. This will stay like this or (hopefully) be removed.

tests/libecalc/energy/models/test_sampled_compressor.py
olelod: The constructor of `SampledCompressor` does this today. It picks 1D/2D/3D internally in `_build_model`. Should it be separated into a specific `create` method?

src/libecalc/energy/models/sampled_compressor.py
tj098895: if that makes the code easier and can be isolated like that it would be great. Does it make sense?

tests/libecalc/energy/models/test_sampled_compressor.py
tj098895: To minimize the amount of code and responsibility of a given class, I prefer to have this in a separate class, called a Factory or Creator (https://refactoring.guru/design-patterns/factory-method/python/example). Also, it should be static.

tj098895: SampledCompressor can and should be an interface - and all different implementations should follow same interface only. Makes sense? (because we have so many sampled compressors implementations)

tj098895: Or even a Compressor interface in Energy Context, if all different types of compressor implementations can follow same interface (same data arguments etc), but I guess that is not true?

src/libecalc/energy/models/convex_hull.py
tj098895: :D

tj098895: it is a nice way to say that we want to to be like that :)

tests/libecalc/energy/models/test_sampled_compressor.py
olelod: •  SampledCompressor  is now an ABC with a shared interface ( evaluate() ,  get_max_standard_rate() ), implemented identically by all 7 classes (1D×3, 2D×3, 3D).
• Built via  SampledCompressorFactory.create(**data) .
• 1D and 2D split into one class per variable combination, each with its own  _active_variables .
• Raw lists from FILE are mapped to fixed names early, in  SampledCompressorData .

src/libecalc/energy/models/sampled_compressor.py
```diff
@@ -0,0 +1,1141 @@
...
+
+    @abstractmethod
+    def get_max_standard_rate(
+        self,
+        suction_pressure: float | None = None,
+        discharge_pressure: float | None = None,
+    ) -> float: ...
+
+    @abstractmethod
+    def get_fuel_power_samples(self) -> tuple[FloatArray, FloatArray] | None:
```
tj098895: used internally? can we change to python lists or similar if this is for "external" use only?

olelod: internal use only

src/libecalc/energy/models/convex_hull.py
kjbrak, line 40: Should be eq=False ? Since it has numpy objects on it which cant be hashed or compared?

PR #1789, opened 2026-09-15
feat: add sampled compressor energy units (2/5)

olelod:
2 of 5, stacked on #1788.

Wraps the model as an `EnergyUnit` (`Converter`/`Consumer`) so it can sit in a network graph.

Stack: #1788 → **2/5** → #1790 → #1791 → #1792
Supersedes #1774.


src/libecalc/energy/energy_units/consumers.py
```diff
@@ -32,3 +37,71 @@ class DieselConsumer(Consumer):
...
+
+class SampledCompressorElectricalConsumer(SampledCompressorConsumerBase):
+    """A SampledCompressor driven by an electrical motor."""
+
+    @classmethod
+    def get_input_energy_type(cls) -> type[ElectricalPower]:
+        return ElectricalPower
+
+
+class SampledCompressorMechanicalConsumer(SampledCompressorConsumerBase):
```
tj098895, line 102: should we separate between having a fixed small set of energy_consumers (and providers and sources), that has the only responsibility to be a wrapper for a unique combination of input and output energy type? With that limited and simple vocabulary, we will be able to have a very small and understandable module that focuses on (energy network) topology only.

we should be ablet o call a sampledcompressor directly, and currently no relation to networks etc - just the input a sampledcompressor would need; rate, pressure and power/fuel ...

tj098895: but we have to somehow still figure out what the sampled compressor model represents, and map correctly the results from the sampled compressor.

olelod:
Superseded by a fresh redesign of this stack: #1799 and #1800 (a final wiring PR is still to follow). Closing in favor of that stack.

PR #1800, opened 2026-09-22
feat: expand a COMPRESSOR_SAMPLED unit into consumer and turbine node specs

olelod:
## Type of Work

- [x] Minor: adds new YAML input, backwards-compatible

- [x] I have added tests
- [x] I have used conventional commits syntax

## What is this PR all about?

Adds a new `COMPRESSOR_SAMPLED` unit type: fuel/power use comes from a lookup table instead of a process model. This PR only adds the YAML type and the logic that expands one into node specs — it isn't wired into the network yet, so nothing existing is affected.

If the table has both FUEL and POWER, the unit expands into two nodes: a generated turbine feeding the compressor. Otherwise it's a single node.


src/libecalc/presentation/yaml/mappers/energy/compressor_sampled_expansion.py
```diff
@@ -0,0 +1,218 @@
...
+    energy_type: type[Energy]
+    read_power: bool
+
+
+@dataclass(frozen=True)
+class TurbineSpec:
+    key: NodeKey
+    name: str
+    inputs: tuple[NodeKey, ...]
+    model: SampledCompressor
```
frodehk: Do we need to send in the full model to `TurbineSpec`? It seems like the turbine only needs the power to fuel relation. Can we represent this as a `FuelPowerCurve` class instead? E.g. where we then have `SampledCompressor.fuel_power_curve -> FuelPowerCurve`, and the `FuelPowerCurve` has methods for `power_for_fuel` and `fuel_for_power`? `TurbineSpec` could then depend on the curve, or only on curve.fuel_for_power

src/libecalc/presentation/yaml/mappers/energy/compressor_sampled_expansion.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+@dataclass(frozen=True)
+class ConsumerSpec:
+    key: NodeKey
+    name: str
+    inputs: tuple[NodeKey, ...]
+    model: SampledCompressor
+    energy_type: type[Energy]
+    read_power: bool
```
frodehk: `read_power` is a bit difficult to interpret. If I understand correctly it controls which  `SampledCompressorResult` field to use for demand? Could this be an explicit `demand_source` enum (ENERGY_USAGE/POWER) instead?

And: Is `input_energy_type` more precise than `energy_type`?

src/libecalc/presentation/yaml/mappers/energy/compressor_sampled_expansion.py
```diff
@@ -0,0 +1,218 @@
...
+    consumer: ConsumerSpec
+
+
+@dataclass(frozen=True)
+class TurbineDriven:
+    turbine: TurbineSpec
+    consumer: ConsumerSpec
+
+
+Expansion = Single | TurbineDriven
```
frodehk: Since `NodeKey` and inputs describe an intermediate graph, would a single `Expansion`(nodes, connections) make the mapping more explicit - more similar to the termns used in `EnergyNetworkTopology` - and avoid the `Single` and `TurbineDriven` wrappers? 

Or are the wrappers needed in the planned mapping to the EnergyNetworkTopology?

